PR #55, opened 2015-10-20
address model, also base files required for Rails to start

patmbolger:
Initial draft of Address model and associated migration and spec files.

Also, added back/altered specific files that were required to have Rails server start.

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -1,0 +1,121 @@
+<<<<<<< Local Changes
+/*
+ * This is a manifest file that'll be compiled into application.css, which will include all the files
+ * listed below.
+ *
+ * Any CSS and SCSS file within this directory, lib/assets/stylesheets, vendor/assets/stylesheets,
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/application.css
```diff
@@ -0,0 +1,121 @@
+<<<<<<< Local Changes
```
diraulo: @patmbolger you have merge conflicts marks here you might want to get rid of them ;)

app/assets/stylesheets/application.css
```diff
@@ -0,0 +1,121 @@
...
+    display: block;
+    color: white;
+}
+.navbar-default .navbar-text > a {
+    color: green;
+}
+
+
+
+=======
```
diraulo: these last 2 lines too :)

spec/models/address_spec.rb
```diff
@@ -0,0 +1,41 @@
+require 'rails_helper'
+
+RSpec.describe Address, type: :model do
+  describe 'Fixtures' do
+    it 'should have a valid factory' do
```
diraulo: No spec implementation here

PR #60, opened 2015-10-23
Peoplemodel

vijayakarumudi:
Pull request for  issue #64 create UserModel

### DIFF
--- a/app/models/user.rb~
+++ b/app/models/user.rb~
@@ -1,0 +1,12 @@
+class User < ActiveRecord::Base
+   has_one :pets_account
+   actable
+   validates_presence_of :email
+   validates_presence_of :first_name
+   validates_presence_of :last_name
+   validates_format_of   :phone, 
+    #with: /\A\d{3}-\d{3}-\d{4}|\A\d[0-9]\z/,
+    with: /\A\d{3}-\d{3}-\d{4}|\A\d{11}\z/,
+      message: 'should be in form of 123-345-7989',
+    allow_blank: true
+end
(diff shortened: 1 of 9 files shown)


app/models/user.rb~
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
```
joaopapereira: This file should not be here because is temporary

joaopapereira:
All the files that end with ~ should not be in version control

app/models/user.rb~
```diff
@@ -0,0 +1,12 @@
...
+   has_one :pets_account
```
joaopapereira: We removed this association, pets_account will no longer exist

app/models/user.rb~
```diff
@@ -0,0 +1,12 @@
...
+   actable
+   validates_presence_of :email
+   validates_presence_of :first_name
+   validates_presence_of :last_name
+   validates_format_of   :phone,
```
joaopapereira: Change this to start using the validators that we already have:

app/models/user.rb~
```diff
@@ -0,0 +1,12 @@
+class User < ActiveRecord::Base
+   has_one :pets_account
+   actable
+   validates_presence_of :email
```
joaopapereira: The validation of this field should look something like:
validates :email, :email => true,
                    :uniqueness => true,
                    :presence => true

joaopapereira: And also the UT for these validation

joaopapereira:
After these changes I think we are good to go

PR #59, opened 2015-10-23
Peoplemodel

vijayakarumudi:
pull request for issue #64 create UserModel

PR #64, opened 2015-10-25
job model, migration, spec tests, closes #59

patmbolger:


app/models/job.rb
```diff
@@ -0,0 +1,12 @@
+class Job < ActiveRecord::Base
+  belongs_to :company
+  belongs_to :company_person
+  belongs_to :address, polymorphic: true
```
joaopapereira: shouldn't this line be:
has_one :address, as: location

app/models/job.rb
```diff
@@ -0,0 +1,12 @@
...
+  belongs_to :job_category
```
joaopapereira: We are missing the connection between Job and Skills

app/models/job.rb
```diff
@@ -0,0 +1,12 @@
...
+  
+  validates_presence_of :title
```
joaopapereira: Don't we need restrictions of size?

app/models/job.rb
```diff
@@ -0,0 +1,12 @@
...
+  validates_presence_of :description
```
joaopapereira: Don't we need restrictions of size?

joaopapereira:
Looks good

PR #66, opened 2015-10-25
Feature/issue #55

amba178:
Commented out has_one :resume for now in the JS model.

### DIFF
--- a/app/models/job_seeker.rb
+++ b/app/models/job_seeker.rb
@@ -1,5 +1,5 @@
 class JobSeeker < ActiveRecord::Base
-  acts_as :person 
+  acts_as :user
   belongs_to :job_seeker_status
   #has_one    :resume 
   has_one	   :address, as: :location
--- a/app/models/job_seeker.rb
+++ b/app/models/job_seeker.rb
@@ -1,4 +1,5 @@
 class JobSeeker < ActiveRecord::Base
+  acts_as :person 
   belongs_to :job_seeker_status
   #has_one    :resume 
   has_one	   :address, as: :location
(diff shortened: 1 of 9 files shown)


app/models/job_seeker.rb
```diff
@@ -0,0 +1,23 @@
+class JobSeeker < ActiveRecord::Base
+  belongs_to :job_seeker_status
+  #has_one    :resume
```
joaopapereira: We are missing the act_as :user.
I just added the Pull Request that enables this. Please merge the changes and add the act_as

app/models/job_seeker.rb
```diff
@@ -0,0 +1,23 @@
+class JobSeeker < ActiveRecord::Base
+  belongs_to :job_seeker_status
+  #has_one    :resume 
+  has_one	   :address, as: :location
+  has_and_belongs_to_many :agency_people, :join_table => "agencies_seekers"
```
joaopapereira: Is this the relationship of JS with JD and CM?

app/models/job_seeker.rb
```diff
@@ -0,0 +1,23 @@
...
+  #has_one    :resume 
+  has_one	   :address, as: :location
+  has_and_belongs_to_many :agency_people, :join_table => "agencies_seekers" 
+  validates_presence_of :year_of_birth, :resume
+  YEARS = 100 
+  validate :validate_year_of_birth  
+
+
+  private 
+  	def validate_year_of_birth
```
joaopapereira: Do you think that we will need this validation elsewhere? or just in this class?

joaopapereira:
Do we have the job_seeker_status PR?

app/models/job_seeker.rb
amba178: Good point, just need here.  I was trying to remind myself that this is not an instance method. I might be wrong, Joao.

amba178:
will create new one.

PR #70, opened 2015-10-26
Feature/issue #55

amba178:
Edited User, took out email field will be provided by Devise. Updated spec file for user.

### DIFF
--- a/app/models/job_seeker.rb
+++ b/app/models/job_seeker.rb
@@ -1,5 +1,5 @@
 class JobSeeker < ActiveRecord::Base
-  acts_as :person 
+  acts_as :user
   belongs_to :job_seeker_status
   #has_one    :resume 
   has_one	   :address, as: :location
--- a/app/models/job_seeker.rb
+++ b/app/models/job_seeker.rb
@@ -5,21 +5,8 @@ class JobSeeker < ActiveRecord::Base
   has_one	   :address, as: :location
   has_and_belongs_to_many :agency_people, :join_table => "agencies_seekers" 
   validates_presence_of :year_of_birth, :resume
-  YEARS = 100 
-  validate :validate_year_of_birth  
+  validate :year_of_birth  
 
 
-  private 
-  	def validate_year_of_birth
-
-  	 self.year_of_birth =~/\A\d{4}\z/ ? year=self.year_of_birth :
-  	         errors.add(:year_of_birth, "cannot be more than or less 4 digits")
-  	         
-  	  if year 
-  	  	 (year.to_i<=Date.today.year)&&( year.to_i>=(Date.today.year - YEARS)) ? 
-  	  	         self.year_of_birth = year.to_s : errors.add(:year_of_birth, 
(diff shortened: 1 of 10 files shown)


app/models/user.rb
```diff
@@ -1,10 +1,7 @@
 class User < ActiveRecord::Base
    actable
-   validates:email,:email => true,
```
joaopapereira: This should be taken care on Issue 60

app/models/job_seeker.rb
```diff
@@ -0,0 +1,25 @@
...
+  	def validate_year_of_birth
+
+  	 self.year_of_birth =~/\A\d{4}\z/ ? year=self.year_of_birth :
+  	         errors.add(:year_of_birth, "cannot be more than or less 4 digits")
+  	         
+  	  if year 
+  	  	 (year.to_i<=Date.today.year)&&( year.to_i>=(Date.today.year - YEARS)) ? 
+  	  	         self.year_of_birth = year.to_s : errors.add(:year_of_birth, 
+  	  	         	 "should be between #{Date.today.year}-#{Date.today.year-YEARS}")
+  	  end
```
patmbolger: Look at using rails extensions for more readable code.  For instance you could say things like:

year_of_birth.between?(18.years.ago, 100.years.ago)

etc....

Also, I would recommend setting the minimum age at 18 (as in the code snippet above).  Actually, I would not bother with checking for max year (100).  You could just check on age 18 or older, and, if not, have the message say 'must be 18 years or older'.

app/models/user.rb
```diff
@@ -1,10 +1,7 @@
...
-            :uniqueness => true,
-            :presence => true
-   validates_presence_of :first_name
-   validates_presence_of :last_name
-   validates   :phone, :phone => true
+   validates_presence_of :first_name, :last_name, :phone 
+   validates_format_of :phone, with:  /\A\(?([0-9]{3})\)?[-|" ".*]?([0-9]{3})[-|" ".*]?([0-9]{4})\z/,
+          :message => "phone number should be 619-234-7890 format"
```
patmbolger: Check if you could use the validator that already exists at app/validators/phone_validator.rb

db/migrate/20151022163836_create_users.rb
```diff
@@ -1,12 +1,12 @@
 class CreateUsers < ActiveRecord::Migration
   def change
     create_table :users do |t|
-      t.string :email
```
joaopapereira: This change is not in this Feature

spec/models/user_spec.rb
```diff
@@ -6,7 +6,6 @@
    
    describe 'Database schema' do
     it { is_expected.to have_db_column :id }
-    it { is_expected.to have_db_column :email}
```
joaopapereira: Same as before

spec/models/user_spec.rb
```diff
@@ -16,16 +15,6 @@
 
    describe 'check model restrictions' do 
 
-     describe 'Email check' do
```
joaopapereira: Same as before

joaopapereira:
Please take a look at these comments and push again

PR #71, opened 2015-10-27
Feature/issue #70

amba178:
renamed table name to seeker_agency_people,
Still can not allow me to create the table if I wanted to index the columns so commented for now the index part. This is what says:
Index name 'index_seekers_agency_people_on_agency_person_id_and_job_seeker_id' on table 'seekers_agency_people' is too long; the limit is 62 characters

### DIFF
--- a/db/migrate/20151026060621_agency_people_job_seekers_join_table.rb
+++ b/db/migrate/20151026060621_agency_people_job_seekers_join_table.rb
@@ -1,9 +1,9 @@
 class AgencyPeopleJobSeekersJoinTable < ActiveRecord::Migration
   	def change
-    	create_table :agencies_seekers, id: false do |t|
+    	create_table :seekers_agency_people, id: false do |t|
     		t.integer :agency_person_id 
     		t.integer :job_seeker_id 
     end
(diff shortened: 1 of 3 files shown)


db/migrate/20151026060621_agency_people_job_seekers_join_table.rb
```diff
@@ -0,0 +1,9 @@
+class AgencyPeopleJobSeekersJoinTable < ActiveRecord::Migration
+  	def change
+    	create_table :seekers_agency_people, id: false do |t|
+    		t.integer :agency_person_id 
+    		t.integer :job_seeker_id 
+    end
+   	 # add_index :seekers_agency_people, [:agency_person_id, :job_seeker_id]
+  end
```
patmbolger: add_index has a ':name' option - this should allow you to get around the problem with the index name being too large.

http://apidock.com/rails/ActiveRecord/ConnectionAdapters/SchemaStatements/add_index

joaopapereira:
We are still missing the other table that is present in
AgileVentures/MetPlus_tracker#70

PR #74, opened 2015-10-27
Companymodel

vijayakarumudi:
Pull request for Job categories model: AgileVentures/MetPlus_PETS/Issue#67


app/models/job_category.rb
```diff
@@ -0,0 +1,8 @@
+class JobCategory < ActiveRecord::Base
+ has_many   :jobs
+ has_and_belongs_to_many:skills
+
+ validates_presence_of :name
+ validates_presence_of :description
+
+end
```
patmbolger: This is missing associations:

has_many :job_specialities
has_many :agency_people, through: :job_specialities

(these can be added in the absence of the models on the "other side" of the association - you just have to make sure that your spec validations of these associations are set to 'pending' until those other models are delivered)

db/migrate/20151027162612_create_job_categories.rb
```diff
@@ -0,0 +1,15 @@
...
+    create_table :job_categories do |t|
+      t.string :name
+      t.string :description
+      t.references :skills, index: true, foreign_key: true 
+      t.references :jobs, index: true, foreign_key: true
+
+      t.timestamps null: false
+    end
+  end
+end
```
patmbolger: Same comments as above - the foreign key for the job_specialities table should be added to this table.

spec/models/job_category_spec.rb
```diff
@@ -0,0 +1,43 @@
+require 'rails_helper'
+
+RSpec.describe JobCategory, type: :model do
+
+   
+   describe 'Associations' do
+
+     
+     xit { is_expected.to have_many :jobs }
```
patmbolger: This can be removed from "pending" status since the other model (Job) already exists.

spec/models/job_category_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+     xit { is_expected.to has_and_belongs_to_many(:skills)}
```
patmbolger: This can be removed from "pending" status since the other model (Skill) already exists.

spec/models/job_category_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
```
patmbolger: Please add tests for the missing two associations noted in the first comment above (these should be marked as pending until the other two models are delivered).

joaopapereira:
Agree with the previews comments

vijayakarumudi:
I am closing

PR #76, opened 2015-10-28
agency model and agency<>company join table

patmbolger:
Closes AgileVentures/MetPlus_tracker-#79

### DIFF
--- a/db/migrate/20151028135231_create_join_table_agency_company.rb
+++ b/db/migrate/20151028135231_create_join_table_agency_company.rb
@@ -1,0 +1,8 @@
+class CreateJoinTableAgencyCompany < ActiveRecord::Migration
+  def change
+    create_join_table :agencies, :companies do |t|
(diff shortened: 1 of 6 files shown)


db/migrate/20151028135231_create_join_table_agency_company.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJoinTableAgencyCompany < ActiveRecord::Migration
+  def change
+    create_join_table :agencies, :companies do |t|
+      # t.index [:agency_id, :company_id]
```
joaopapereira: Is there any special reason for these to be commented out?

patmbolger: No .... the generator commented those out upon creation.  I left them commented since I'm not sure we need indices on that table (or these particular indices, for that matter).

PR #78, opened 2015-10-28
AgencyPerson model

patmbolger:
closes AgileVentures/MetPlus_tracker#76


patmbolger:
CI check failed due to the Agency model not being in existence yet.  My prior PR creates the Agency model, after which the CI tests will pass.

amba178:
@patmbolger, sorry  for collision on AP model , I should not have done that.  if you follow the comment from @joaopapereira under issue-#70, asked me for AgencyPerson model since i was doing the join table AP<>JS and I forgot to check the waffle. should i close the PR?

PR #79, opened 2015-10-28
Feature/issue #55

amba178:
@joaopapereira, I looked up validator folder how you implemented the validate_each for specific field and I applied to year_of_birth. it is cool trick.


joaopapereira:
Please do not remove the email tests of fields form the database

PR #81, opened 2015-10-28
Feature/issue #70

amba178:


app/models/agency_person.rb
```diff
@@ -0,0 +1,4 @@
+class AgencyPerson < ActiveRecord::Base
+  belongs_to :address
+  belongs_to :agency
```
joaopapereira: We are missing the connection to the JS table through agency_people_job_seekers

joaopapereira:
Please correct my previous comment and we can merge this issue

PR #82, opened 2015-10-29
SkillLevel model

patmbolger:
Closes AgileVentures/MetPlus_tracker#84

### DIFF
--- a/app/models/skill_level.rb
+++ b/app/models/skill_level.rb
@@ -1,0 +1,6 @@
+class SkillLevel < ActiveRecord::Base
+  validates_presence_of :name
+  validates_length_of :name, maximum: 20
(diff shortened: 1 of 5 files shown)


joaopapereira:
Looks good, but i think we should add on the creation file some default values like:
Beginner
Intermediate
Expert

patmbolger:
By "creation file" do you mean the migration file?

joaopapereira:
Yes I do

patmbolger:
Not sure what you mean .... we can only specify one default value on each column ... or is there some other mechanism you are referring to?

patmbolger:
Added records create (beginner, etc....) to seeds.rb

joaopapereira:
Looks good

PR #87, opened 2015-10-30
closes AgileVentures/MetPlus_tracker#83

patmbolger:
1) Created join table
2) Added association to Skill model (with JobCategory using new join table)
3) Added spec test of above association (pending until JobCategory model is delivered)


joaopapereira:
Looks good

PR #88, opened 2015-10-30
Closes AgileVentures/MetPlus_tracker#81

patmbolger:
Created AgencyRole model
Migration file for agency_roles table
Rspec tests for AgencyRole

Migration file for join table: AgencyPerson and AgencyRole
Migration file for join table: AgencyPerson and JobCategory

Moved certain tests from 'pending' to active state for AgencyPerson and Job models

### DIFF
--- a/app/models/agency_person.rb
+++ b/app/models/agency_person.rb
@@ -4,8 +4,8 @@ class AgencyPerson < ActiveRecord::Base
   belongs_to :agency
   belongs_to :address
   has_and_belongs_to_many :agency_roles
-  has_many :job_specialities
-  has_many :job_categories, through: :job_specialities
+  # has_many :job_specialities
+  has_many :job_categories, join_table: :job_specialities
   
   validates_presence_of :agency_id
 end
--- a/app/models/agency_role.rb
+++ b/app/models/agency_role.rb
@@ -1,2 +1,6 @@
 class AgencyRole < ActiveRecord::Base
+  has_and_belongs_to_many: agency_people
+  validates_presence_of :role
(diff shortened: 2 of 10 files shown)


joaopapereira:
Conflicts. Please review code

patmbolger:
The conflict is due to the fact that the 'job_specialities' table in the ERD is create in two conflicting ways:
1) Associations between models that pass through that table use 'has many, through', however,
2) The table itself did not have an id and thus was modeled as a simple join table.

So, here I went ahead and changed existing associations to reflect this.  

In particular, AgencyPerson was:

    has_many :job_specialities
    has_many :job_categories, through: :job_specialities

and I changed this to:

    has_many :job_categories, join_table: :job_specialities

So, the question is whether we need 'job_specialities' will be a model in its own right, and contain additional fields beyond the foreign keys, or can it be a simple join table?  

The idea between AgencyPerson<>JobCategory association is to allow, for example, a Job Developer (AgencyPerson) to specialize in placing people into certain categories of jobs.  (e.g. 'Bob Smith' places people in 'Shipping Clerk', 'Truck Driver', 'Inventory Control' and 'Accounts Receivable Clerk' jobs.

With that, I cannot see the use of making 'job_specialities' its own model, and if we don't need that why clutter up our design?

Thoughts?

joaopapereira:
I agree with you, I think we can live without that model. In the future if we see it is needed we can add it

joaopapereira:
Add the 3 job roles to the seeds file
Job Developer
Case Manager
Agency Manager

joaopapereira:
Looks ok now

PR #89, opened 2015-11-01
Feature/issue #77

amba178:
CompanyPerson model

### DIFF
--- a/spec/factories/company_people.rb
+++ b/spec/factories/company_people.rb
@@ -1,0 +1,8 @@
+FactoryGirl.define do
+  factory :company_person do
+    company nil
+    address nil
+    company_roles ""
+  end
+
+end
--- a/spec/models/company_person_spec.rb
+++ b/spec/models/company_person_spec.rb
@@ -1,0 +1,46 @@
+require 'rails_helper'
+
+describe CompanyPerson, type: :model do
+   describe 'Database schema' do
+    it { is_expected.to have_db_column :id }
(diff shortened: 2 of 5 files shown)


spec/factories/company_people.rb
```diff
@@ -0,0 +1,8 @@
+FactoryGirl.define do
+  factory :company_person do
+    company nil
+    address nil
```
patmbolger: Recommend changing "address nil" to just "address".  This will reference the :address factory.

(We should do the same for "company" after that model is delivered.)

spec/models/company_person_spec.rb
```diff
@@ -0,0 +1,49 @@
...
+       it { is_expected.to validate_presence_of :last_name }
+       
+     end
+
+     describe 'Phone check' do
+       it { should_not allow_value('asd', '123456', '123 123 12345', '123 1231 1234', '1123 123 1234', ' 123 123 1234').for(:phone)}
+
+       it { should allow_value('+1 123 123 1234', '123 123 1234', '(123) 123 1234', '1231231234', '+1 (123) 1231234').for(:phone)}
+
+     end
```
patmbolger: This model does not have first_name, last_name or phone.  Was it your intent to add those attributes?

amba178: magic of acts_as :user. all attributes of User class are available to company_person class.
but we can not assume as a part of company_person columns fields.

amba178:
Needed to do PR the updated version.

PR #91, opened 2015-11-02
Closes AgileVentures/MetPlus_tracker#77

amba178:
fixed factories  and updated rspec file.


joaopapereira:
The change looks good. Next time try not to commit changes unrelated to this PR like the change in users model

PR #92, opened 2015-11-02
Closes AgileVentures/MetPlus_tracker#78

amba178:


joaopapereira:
Add defined roles in seeds
Human Resources
Company manager

patmbolger:
Salem - this task also includes the join table creation, as noted in waffle:

"This includes the join table 'company_people_company_roles' (between 'company_people' and 'company_roles')"

amba178:
- Need to seed some data to it.

PR #94, opened 2015-11-02
Closes AgileVentures/MetPlus_tracker#70

amba178:
- fixed merge conflict


spec/factories/agency_people.rb
```diff
@@ -1,7 +1,7 @@
 FactoryGirl.define do
   factory :agency_person do
-    agency
-    address
-  end
+    agency_id  "2"
+    address_id "3"
+end
```
patmbolger: Salem - you don't need to deliver the AgencyPerson model material (model, factor) - all that is already in upstream development.  If you rebase against that branch you'd have it in your feature branch before you created your PR.

However, if it is easier to keep that stuff here, then that is fine.  If so, however, please undo your changes to this factory file.  The references to 'agency' and 'address' are actually references to those factories as defined in their respective factory files.  With that, we get a true mock object for each of those models, complete with attributes as defined in their specific factories.  Hope this makes sense - if not we can discuss later today.

amba178: I dont know what i was doing there...i will fix it.

PR #104, opened 2015-11-04
closes AgileVentures/MetPlus_tracker#96

patmbolger:
Configured ActionMailer for use by the 'devise' gem

Added 'figaro' gem to manage env-specific, and sensitive, config data (see config/application.yml.example)

Added Main controller skeleton (just sufficient to start testing Devise account confirmation via email)


joaopapereira:
You didnt need to do any change in the test environment file?

patmbolger:
No tests yet, unfortunately.  I was hurrying to get this ready so that we have a chance of showing the client registration/login/logout features on Friday.

I intend to deliver tests (will add an issue for that).

patmbolger:
On second thought, 'm not sure what I would test here.  I am using Devise code (even their Mailer class) and would be testing Devise functionality that the gem developers have already tested.

When we add our own Mailer we will definitely need tests (spec and cucumber).

joaopapereira:
I believe we should have 2 different PRs for this. But going to merge due to the urgency

PR #106, opened 2015-11-05
Closes AgileVentures/MetPlus_tracker#92

sidhene:
- added phone number field to registration page
- changed navbar logged in username from green to yellow for better visibility


joaopapereira:
The name of the PR should be different. Something like:
Closes AgileVentures/MetPlus_tracker#92

joaopapereira:
We are missing tests for the controller

PR #107, opened 2015-11-05
changed erb 2 haml and fixed url: login, logout, users/new

amba178:
customized url, seeded some data to display and changed erb to haml. 
Note: Emy is adding some fields, and using erb syntax. 
erb syntax error in haml template.


joaopapereira:
Do we have some tests for this?

joaopapereira:
Is this PR related with any issue?

amba178:
On my side, it is issue #93 and  partly combined Emy's issues #91 and #92. 
I shouldn't have done that maybe,  during PP session I took the task to customize the url, seeding, and erb2haml part. I hope it does not create  any conflict.

joaopapereira:
please assign the task the correct name. and about the tests?

amba178:
I will do that.

config/routes.rb
```diff
@@ -1,15 +1,21 @@
...
+  devise_for :users, :path_names => {:sign_up => "new", :sign_out => 'logout', 
+                                     :sign_in => 'login' }
+  
+
+
+   devise_scope :user do
+     #match  "/new"     => 'devise/registrations#new',   via: 'get'
+     match  '/login'   =>  'devise/sessions#new',       via: 'get'
+     match  '/logout'  => 'devise/sessions#destroy',    via: 'get'
+  end
```
patmbolger: Salem - since you are altering (or adding to?) the default routes for Devise perhaps you could add some tests that verify those, fir example:

    context 'Routing' do
        it 'routes /user/login to devise/sessions#new' do
            expect(get: '/user/login').to route_to(controller: 'devise/sessions', action: 'new')
        end
        it 'routes /user/new to devise/registrations#new' do
            expect(get: '/user/new').to route_to(controller: 'devise/registrations', action: 'new')
        end

        etc ....
    end

I would do this for all of the new routes (including GET, POST, etc.) introduced by the added routes code.

And, please note that the details in my example above might be wrong, but the rspec methods are correct (e.g. "route_to").

PR #109, opened 2015-11-06
Feature/issue #93

amba178:
I didn't understand when you testing... Do you want me to write test for views.There are from devise.


app/views/devise/registrations/new.html.haml
```diff
@@ -0,0 +1,29 @@
...
+= form_for(resource, as: resource_name, url: registration_path(resource_name)) do |f|
+  = devise_error_messages!
+  .field
+    = f.label :email
+    %br/
+    = f.email_field :email, autofocus: true
+  .field
+    = f.label :first_name
+    %br/
+    = f.email_field :first_name
```
joaopapereira: This is incorrect

app/views/devise/registrations/new.html.haml
```diff
@@ -0,0 +1,29 @@
...
+  .field
+    = f.label :last_name
+    %br/
+    = f.email_field :last_name
```
joaopapereira: This is incorrect, is pointing to a email field

app/views/devise/registrations/new.html.haml
amba178: Yes, that is original error. initial was there. I hopefully Emy took care of it.

app/views/devise/registrations/new.html.haml
amba178: The same thing here. Emy version should fix that. I did not want to create conflict by changing the fields error you talking about.

amba178:
Will create new PR.

PR #111, opened 2015-11-08
Closes AgileVentures/MetPlus_tracker#93

amba178:
- all devise views should be in haml format
- /login instead of users/sign_in
-/logout instead of users/sign_out 
-/users/new instead of users/sign_up 
these customized route have their own test spec file.
seeded some data 

let me know if I have typo or erros pls.


joaopapereira:
Looks good

PR #119, opened 2015-11-12
Cancancan integration

vijayakarumudi:
PR for Closes AgileVentures/MetPlus_tracker#90


app/models/user.rb
```diff
@@ -8,4 +8,33 @@ class User < ActiveRecord::Base
...
+  
+  def is_employer?
+      self.role == 'employer'
+  end
+  
+  def is_admin?
+      self.role == 'admin'
+  end
+
+
```
patmbolger: These methods are not correct.

The User model has a "belong_to" relationship with JobSeeker, AgencyPerson and CompanyPerson via the "acts-as" gem.  The users table has these columns as a a result:

    actable_id
    actable_type

These have to be used to find the associated model for the user record (one of the three listed above).  

For instance, if user.actable_type = 'JobSeeker', then we can return true for #is_job_seeker?.

For AgencyPerson and CompanyPerson, we have to go further and interrogate the agency_roles, and company_roles table, respectively.

db/migrate/20151022163836_create_users.rb
```diff
@@ -5,6 +5,7 @@ def change
       t.string :first_name
       t.string :last_name
       t.string :phone
+      t.string :role
```
patmbolger: We don't want this field unless we decide we are going to duplicate information maintained elsewhere in the DB (see my prior comment).  I think the complexity of duplication is not worth the benefit of an easy "role" reference in this table.

app/models/user.rb
joaopapereira: I was thinking maybe something like this:
def is_job_seeker?
  self.is_a?(JobSeeker)
end


Or maybe even we could give this a new twist and implement the function

is_job_seeker in JobSeeker class

and then in the ability class instead of:
user.is_job_seeker?
we would have
user.respondes_to? is_job_seeker?

I leave it open ended in this point the one you like better

db/migrate/20151022163836_create_users.rb
joaopapereira: I agree with Patrick.
Nevertheless we will need a field in this table called "admin" or "super_user" boolean to check if the current user should have super powers or not

app/models/user.rb
patmbolger: Will "is_a?" work in this instance?  (that is, does acts-as cause the user instance to respond as the class on the other side of the belongs_to association?)

vijayakarumudi:
I closed this one

PR #120, opened 2015-11-12
Connects to AgileVentures/MetPlus_tracker#112 navbar nov11

sidhene:
Minor updates/repairs to navbar that I noticed were needed while doing the homepage mockups. For example, the "Logout" button had gotten lost and replaced by text. See ticket #112 'Home page layout' for details.


joaopapereira:
Took some tries but got name right.
The connect you made in you comment is linking the wrong repository.
Next time please use the naming that we defined

PR #124, opened 2015-11-13

amba178:
Do not have a test case for strong parameters, for updating user. Think the only way we can test if we can emutate the form for updating, maybe cucumber. I have tested it manually and it works.
Let me know if you want me to do through cucumber.


joaopapereira:
Yes I would like to start having cucumber test cases for integration in Login/registration and such. If we already have the ability to reach the pages I think we can start testing them.
So the question is can we login with a user and update the user information?

amba178:
I am very sure we can login and update with user info. Moreover it is time to start cucumber.

PR #125, opened 2015-11-16

amba178:
- You can leave me any feedback. Initially, I started out by commenting the strong parameters for update.


features/support/shared_db_connection.rb
```diff
@@ -0,0 +1,12 @@
...
+  @@shared_connection = nil
+ 
+  def self.connection
+    @@shared_connection || retrieve_connection
+  end
+end
+ 
+# Forces all threads to share the same connection. This works on
+# Capybara because it starts the web server in a thread.
+ActiveRecord::Base.shared_connection = ActiveRecord::Base.connection
```
patmbolger: I am leery of this approach.  The Capybara author says: "This may have thread safety implications and could cause strange failures, so use caution with this approach." 

https://github.com/jnicklas/capybara#transactions-and-database-setup

spec/factories/users.rb
```diff
@@ -1,10 +1,12 @@
 FactoryGirl.define do
   factory :user do
```
patmbolger: Recommend adding right above this line:

    sequence :email do | n |
        "unique#{n}@gmail.com"
    end

spec/factories/users.rb
```diff
@@ -1,10 +1,12 @@
...
 	password 'MyString1233'
+	password_confirmation 'MyString1233'
 	first_name 'John'
 	last_name "Doe"
 	email 'johndoe@place.com'
```
patmbolger: If follow suggestion above (adding sequence), this line is simply:

    email

app/controllers/application_controller.rb
```diff
@@ -8,8 +8,15 @@ class ApplicationController < ActionController::Base
...
   def configure_permitted_parameters
+
+       
     devise_parameter_sanitizer.for(:sign_up) << :first_name
     devise_parameter_sanitizer.for(:sign_up) << :last_name    
     devise_parameter_sanitizer.for(:sign_up) << :phone
+      
+    ["first_name","last_name", "phone"].each{ |field|
+      devise_parameter_sanitizer.for(:account_update)<<field.to_sym 
+    }
```
patmbolger: Is there a reason why the array does not have symbols?  (we can avoid the "to_sym" method call if using symbols directly)

amba178: @patmbolger, No reason. I guess I am very fan of String. well, considering cost, performance maybe symbol way to go. Good comments, Patrick. On the rest comments, I am doing reading some reading on it. Thanks, Patrick.

features/support/shared_db_connection.rb
amba178: @patmbolger: how does  DatabaseCleaner.strategy = :truncation relate to shared_db_connection.rb file. This is what my problem was, let say  a user login right than if i want to go to edit page/user profiles, devise says I have to login before continue to edit page.From my understanding once user login, the thread is done and can not be shared with any new thread, in this case editing the page. It seems to me in the background there is some javascript going on and some how we need to make thread connection.  But I have tried to set DatabaseCleaner.strategy = :truncation and got long delay/freezes up at the end the test timed me out. Basically it fails. I think there is a fix for it if we still want to use :truncation. Let me know your finding. thanks.

amba178: Very good comment though.

features/manage_users.feature
```diff
@@ -0,0 +1,33 @@
+Feature: Manage Users
```
joaopapereira: The indentation in this file looks a bit odd

features/step_definitions/users_steps.rb
```diff
@@ -5,15 +5,50 @@
   end
 end
 
-When(/^I fill in the fields$/) do |table|
+When(/^I fill in the fields:$/) do |table|
```
joaopapereira: This step should be inside web_steps.rb

features/step_definitions/users_steps.rb
```diff
@@ -5,15 +5,50 @@
...
 
 Given(/^I activate user "([^"]*)"$/) do |email|
   user = User.find_by_email(email)
   expect(user.activate(user.activation_token)).to be true
-end
\ No newline at end of file
+end
+
+
+Given(/^the following (.+) records:$/) do |factory, table|
```
joaopapereira: This should be in generic_steps.rb

features/step_definitions/users_steps.rb
```diff
@@ -5,15 +5,50 @@
...
+     visit '/login' 
+     fill_in "Email", :with => email
+     fill_in "Password", :with => password
+     check "user_remember_me"
+     click_button "Log in"
+    end
+   expect(page).to have_content "Signed in successfully."
+end
+
+When(/^I visit profile for "(.*?)"$/) do |first_name|
```
joaopapereira: This should be inside navigation_steps.rb

joaopapereira:
Please do not create a new PR for this topic for now, there are other more important tasks that we need to do.

When you pick up again this part, please review my previous comments

amba178:
I am not going to creat new PR. I have read the your comment, and if you want to put aside that fine with me.

features/step_definitions/users_steps.rb
amba178: This is not my code, just wanted to reuse previous code and was there before in the user_steps.rb

PR #127, opened 2015-11-18
Connects to AgileVentures/MetPlus_tracker#11

patmbolger:
- Added agency_admin controller and functional tests for that controller
- Added "Admin" button to main menu if signed_in user is Agency Admin
- Implemented initial "main admin" page
- Added methods to determine 1) agency, 2) agency admin and 3) agency manager from sign in user (if user is an AgencyPerson)
- Added a constant (ROLE) to AgencyRole and CompanyRole.  This specifies the exact strings to be used for the various roles.
- Changed code to reflect use of these constants (factories, unit tests, seeds.db, etc.)
- Added 'fax' and 'description' attributes to Agency
- Added initial cucumber scenarios for agency admin features (log in, go to main admin page)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -64,6 +64,8 @@ group :development, :test do
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
+  
+  gem 'launchy'
 end
 
 group :test do
@@ -73,6 +75,7 @@ group :test do
   gem 'email_spec'
   gem 'poltergeist'
   gem "codeclimate-test-reporter"
+  gem 'selenium-webdriver'
 end
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -29,10 +29,11 @@ end
 # Use ActiveModel has_secure_password
 gem 'bcrypt', '~> 3.1.7'
 
-
 gem 'devise'
 
 gem 'figaro'
+
+gem 'cancancan', '~> 1.10'
 
 
 # Use Unicorn as the app server
--- a/Gemfile
+++ b/Gemfile
@@ -75,7 +75,6 @@ group :test do
   gem 'email_spec'
   gem 'poltergeist'
   gem "codeclimate-test-reporter"
-  gem 'selenium-webdriver'
 end
 
 group :production do
--- a/Gemfile
+++ b/Gemfile
@@ -65,7 +65,6 @@ group :development, :test do
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
   
-  gem 'launchy'
 end
 
 group :test do
@@ -74,7 +73,9 @@ group :test do
   gem 'shoulda-matchers', git: 'https://github.com/thoughtbot/shoulda-matchers', require: false
   gem 'email_spec'
   gem 'poltergeist'
-  gem "codeclimate-test-reporter"
+  gem 'codeclimate-test-reporter'
+  gem 'launchy'
+  gem 'selenium-webdriver' # Enables running cuke tests with browser (see env.rb)
 end
 
 group :production do
--- a/app/helpers/agency_admin_helper.rb
+++ b/app/helpers/agency_admin_helper.rb
@@ -1,10 +1,7 @@
 module AgencyAdminHelper
-  def current_user_agency_admin
+  def current_user_agency_admin?
     return false unless user_signed_in?
     return false unless ((person = current_user.actable).is_a? AgencyPerson)
-    person.agency_roles.each do |ar|
-      return true if ar.role == AgencyRole::ROLE[:AA]
-    end
-    false
+    person.agency_roles.pluck(:role).include? AgencyRole::ROLE[:AA]
   end
 end
--- a/app/helpers/agency_admin_helper.rb
+++ b/app/helpers/agency_admin_helper.rb
@@ -1,0 +1,10 @@
+module AgencyAdminHelper
+  def current_user_agency_admin
+    return false unless user_signed_in?
+    return false unless ((person = current_user.actable).is_a? AgencyPerson)
+    person.agency_roles.each do |ar|
+      return true if ar.role == AgencyRole::ROLE[:AA]
+    end
+    false
+  end
+end
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -21,6 +21,9 @@ class Agency < ActiveRecord::Base
   end
   
   def self.this_agency(logged_in_user)
+    raise RunTimeError, 'Logged in user in not an agency person' unless
+            logged_in_user.actable.is_a? AgencyPerson
+            
     @@this_agency = @@this_agency || 
           Agency.find(logged_in_user.actable.agency_id)
   end
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -12,25 +12,25 @@ class Agency < ActiveRecord::Base
   validates :fax, :phone => true
   
   @@agency_admin = nil
-  @@agency_manager = nil
   @@this_agency = nil
-    
+  
+  # For the following methods, 'logged_in_user' can be either an
+  # AgencyPerson object, or the User object associated with an AgencyPerson
+  # In either case, this object needs to represent:
+  #   an AgencyPerson object (directly or via user.actable), and,
+  #   that person must be logged in
+  
   def self.agency_admin(logged_in_user)
     @@agency_admin = @@agency_admin || 
           find_user_with_role(logged_in_user, :AA)
   end
   
   def self.this_agency(logged_in_user)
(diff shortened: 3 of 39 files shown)


app/models/agency_role.rb
```diff
@@ -1,6 +1,14 @@
 class AgencyRole < ActiveRecord::Base
-  has_and_belongs_to_many :agency_people
+  has_and_belongs_to_many :agency_people,
+                          join_table: 'agency_people_roles'
   validates_presence_of :role
   validates_length_of   :role, maximum: 40
+  
+  ROLE = {JD: 'Job Developer',
+          CM: 'Case Manager',
+          AM: 'Agency Manager',
```
joaopapereira: What is the difference between agency manager and admin?

app/models/agency.rb
```diff
@@ -9,4 +9,49 @@ class Agency < ActiveRecord::Base
   validates :phone, :phone => true
   validates :email, :email => true
   validates :website, :website => true
+  validates :fax, :phone => true
+  
+  @@agency_admin = nil
```
joaopapereira: After reviewing this piece of code it looks like a huge compromise in terms of architecture just to allow us to display the administrator name in the admin area.
Do we have a requirement to display the manager name?
Create a Singleton with a 1 to 1 relation between agencies and admins solve our problem. Like a hash agency as the key and user as the value? This way we will not compromise the architecture.
If we have 2 managers we do not display any name in the page? Is this a requirement?
The difference between JD,CM and the manager role is the ability to add users, change images on the website. Does it make sense to have more then 1?
These are some of the questions the I have on this post of the implementation

app/helpers/agency_admin_helper.rb
```diff
@@ -0,0 +1,10 @@
+module AgencyAdminHelper
+  def current_user_agency_admin
```
joaopapereira: I would rename this function to current_use_agency_admin? To be more explicit that it is a Boolean return that we expect

config/routes.rb
```diff
@@ -20,6 +22,8 @@
 
   # You can have the root of your site routed with "root"
   root 'main#index'
+  
+  get 'agency_admin/home'
```
joaopapereira: This line is duplicated

config/routes.rb
```diff
@@ -1,6 +1,8 @@
 Rails.application.routes.draw do
 
 
+  get 'agency_admin/home'
```
joaopapereira: Maybe we can organize the admin part as stated in this routing guide http://guides.rubyonrails.org/routing.html#controller-namespaces-and-routing

features/agency_admin.feature
```diff
@@ -0,0 +1,43 @@
...
+  | MetPlus | Detroit | 456 Sullivan Street | 48204   |
+  | MetPlus | Detroit | 3 Auto Drive        | 48206   |
+
+Scenario: login as agency admin
+  Given I am on the home page
+  And I login as "aa@metplus.org" with password "qwerty123"
+  Then I should see "Signed in successfully."
+  And I should see "Admin"
+  
+Scenario: go to main admin page
```
joaopapereira: Maybe we should have negative scenarios as well to make sure unauthorized people do not see the admin button

spec/controllers/agency_admin_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      sign_in @agency_admin
+      expect(Agency.agency_admin(@agency_admin)).to eq @agency_admin
+    end
+    it 'agency admin - from job developer' do
+      sign_in @job_developer
+      expect(Agency.agency_admin(@job_developer)).to eq @agency_admin
+    end
+  end
+
+end
```
joaopapereira: Missing tests for case manager?

spec/controllers/agency_admin_controller_spec.rb
```diff
@@ -0,0 +1,76 @@
...
+      @agency_admin.save!
+      
+      @job_developer = FactoryGirl.build(:agency_person, agency: @agency)
+      @job_developer.agency_roles << @job_developer_role
+      @job_developer.save!
+    end
+    
+    it 'this agency - from agency manager' do
+      sign_in @agency_manager
+      expect(Agency.this_agency(@agency_manager)).to eq @agency
```
joaopapereira: Shouldn't these tests use the function current_user instead of @.....?
Because the test is just checking that the function this_agency is correctly implemented and not testing anything in the controller itself.

Gemfile
```diff
@@ -64,6 +64,8 @@ group :development, :test do
 
   # Spring speeds up development by keeping your application running in the background. Read more: https://github.com/rails/spring
   gem 'spring'
+  
+  gem 'launchy'
```
joaopapereira: Will we need this gem in the app?

joaopapereira:
Nice job tackling this huge requirement I see a lot of good stuff on it.
Please review my comments. We can talk tomorrow if you have any doubt

app/models/agency_role.rb
patmbolger: The AM is the person in charge of the agency.  The AA manages the PETS system using capabilities available only to that role.  Of course, these roles could be assigned to the same person.

As an example scenario, the AM makes a business decision to hire a JD.  The AM directs the AA to register the new JD in PETS.  At that point, the systems sends an email to the JD notifying him/her of the registration.  (at this point, the JD may have to click a link in the email to confirm the registration).

app/helpers/agency_admin_helper.rb
patmbolger: agreed

Gemfile
patmbolger: We don't need it in production.  In fact, we only need it in test - I will move it there.

app/models/agency_role.rb
joaopapereira: My question at this moment is. Do we really want to add one more level of complexity in this?
Because have 2 different types of users accessing the administration will make the administration area more complex.
Maybe for now we can go forward with this, but if we see that having these 2 roles will make our cancancan configuration much more complicated and the administration page itself, I would advise against the usage of these 2 roles in the first release. We can always implement the AA in a second take

spec/controllers/agency_admin_controller_spec.rb
patmbolger: That might be good to do, but right now 'current_user' is not within the scope of the spec tests.  Will have to come back to this later.

app/models/agency_role.rb
patmbolger: OK, I like simpler until and if we need to get more complex.  I will consolidate admin and manager functions into just an AM role.

spec/controllers/agency_admin_controller_spec.rb
joaopapereira: If current user is not in scope of these tests what part of the controller is being tested here?

features/support/env.rb
```diff
@@ -79,7 +79,7 @@
     #debug:       true
     )
   end
-  Capybara.default_driver    = :poltergeist
+  # Capybara.default_driver    = :poltergeist
   Capybara.javascript_driver = :poltergeist
 end
```
patmbolger: I was getting a failure on one of my scenarios when running under Poltergeist.  That scenario (and all others) worked fine in live app as well as in a browser test (driven by selenium).  

João - I think you should move us to a different web driver, as you had talked about yesterday.  This is an area where a little experience will make that transition much more efficient and less risky.

spec/controllers/agency_admin_controller_spec.rb
patmbolger: My point is that 'current_user' does not seem to be available in the spec test, but I think it *should* be.  I may need some help to get that to work.

joaopapereira:
This commit is quite big, we need to try to resolve smaller tickets in the future, because it is hard to review all the code.
Looks good, just need to rebase and I will merge it

app/models/user.rb
```diff
@@ -15,50 +15,27 @@ def self.is_job_seeker?(user)
...
       return false unless user.actable_type == "CompanyPerson"
-      person = CompanyPerson.find user.actable_id 
-      person.company_roles.each do |ca|
-       return true if ca.role == "Company Admin" 
-       end
+      user.actable.company_roles.pluck(:role).include? CompanyRole::ROLE[:EA]
     end
     
-    def self.is_employee?(user)
+    def self.is_company_contact?(user)
```
joaopapereira: Company contact?

patmbolger: see the description of employer roles here: http://www.agileventures.org/projects/metplus/documents/system-behavior ("A word about employer roles").  

I am proposing two roles.  I am calling them company roles internally (consistent with Company model naming), but externally (with client) referring to them as employer roles.

In the future we will probably add additional roles for Company people, I imagine.

PR #131, opened 2015-11-23

amba178:
In order to update user profiles we need to add these :first_name, :last_name, and  :phone parameters to devise_parameter_sanitizer, account_upate.


joaopapereira:
Looks good

PR #132, opened 2015-11-26
connects AgileVentures/MetPlus_tracker#131

amba178:
This is implementation for login page, ticket#113.


joaopapereira:
The change looks good

PR #133, opened 2015-11-26
Closes AgileVentures/MetPlus_tracker#127

patmbolger:
Added agencies_controller with two actions: edit and update
Added spec controller tests
Added cucumber integration tests


app/helpers/agency_admin_helper.rb
```diff
@@ -4,4 +4,8 @@ def current_user_agency_admin?
     return false unless ((person = current_user.actable).is_a? AgencyPerson)
     person.agency_roles.pluck(:role).include? AgencyRole::ROLE[:AA]
   end
+  
+  def person_full_name(person)
```
joaopapereira: Shouldn't this be a model method?

app/models/agency.rb
```diff
@@ -9,10 +9,7 @@ class Agency < ActiveRecord::Base
   validates :phone, :phone => true
   validates :email, :email => true
   validates :website, :website => true
-  validates :fax, :phone => true
```
joaopapereira: Why do we stop validating the fax?

config/routes.rb
```diff
@@ -1,24 +1,22 @@
...
-
-#  post '/login' => 'session#create'
-
-#  get '/logout' => 'session#destroy'
+  
+  resources :agencies, only: [:edit, :update]
+  
+  root 'main#index'
+  
+  get 'agency_admin/home'
```
joaopapereira: Maybe this URL should look like /admin/agency/
Like this is more rest like

joaopapereira:
Looks good. Just clarify my comments

patmbolger:
I moved the 'full_name' method to User as this will be needed by both AgencyPerson and CompanyPerson.

I moved the agencies controller under the 'admin' namespace.

app/models/user.rb
```diff
@@ -37,5 +37,9 @@ def self.is_company_contact?(user)
       return false unless user.actable_type == "CompanyPerson"
       user.actable.company_roles.pluck(:role).include? CompanyRole::ROLE[:EC]
     end
+    
+    def self.full_name(person)
```
joaopapereira: This should be a instance method not a class method

config/routes.rb
```diff
@@ -1,24 +1,22 @@
...
-
-#  post '/login' => 'session#create'
-
-#  get '/logout' => 'session#destroy'
+  
+  resources :agencies, path: '/admin/agencies', only: [:edit, :update]
+  
+  root 'main#index'
+  
+  get 'agency_admin/home'
```
joaopapereira: Shouldn't this rout be under the resources part above?

app/models/agency.rb
```diff
@@ -9,10 +9,7 @@ class Agency < ActiveRecord::Base
...
-  
-  @@agency_admin = nil
```
joaopapereira: I believe we have one Issue for these changes, One to refactor this part?
Or are you thinking in the refactor ticket to refector the find_user_with_role function?

app/models/user.rb
patmbolger: what was I thinking?  thanks for catching that ....

app/models/agency.rb
patmbolger: Yes, I have refactored that in the next (pending ) PR.

config/routes.rb
patmbolger: I moved it under admin.

joaopapereira:
Looks good.

PR #135, opened 2015-11-30
Closes AgileVentures/MetPlus_tracker#137

patmbolger:
added agency branch model

### DIFF
--- a/db/migrate/20151127164027_create_branches.rb
+++ b/db/migrate/20151127164027_create_branches.rb
@@ -1,0 +1,11 @@
+class CreateBranches < ActiveRecord::Migration
+  def change
+    create_table :branches do |t|
+      t.references :agency, index: true, foreign_key: true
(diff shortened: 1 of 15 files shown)


db/migrate/20151127164027_create_branches.rb
```diff
@@ -0,0 +1,10 @@
+class CreateBranches < ActiveRecord::Migration
+  def change
+    create_table :branches do |t|
+      t.references :agency, index: true, foreign_key: true
+      t.string :code, limit: 8
+
```
joaopapereira: arent ẃe missing the address_id?

patmbolger: The Branch model has_one address, so the foreign key is in the addresses table (using the polymorphic ID and type).

joaopapereira: Ok

joaopapereira:
Look good

PR #137, opened 2015-12-03
closes AgileVentures/MetPlus_tracker#140

patmbolger:
Added cucumber hook @selenium - use that for tests that require response to a modal dialog (such as "Delete Agency Branch?").  This will use Selenium as the javascript driver.  (The default javascript driver is Poltergeist, but that driver has limitations with respect to modal dialog interaction).

Note that Selenium is *not* headless - it will cause the test to be run with a "head" in Firefox.


joaopapereira:
Looks good

PR #138, opened 2015-12-03
connects AgileVentures/MetPlus_tracker#139

amba178:
Did not know where input and label tags apply in the page in this context, but I have attached id to each tags as placeholder, '#putidhere' for now. I think it is good practice to put comment where these css stuff apply in the page whenever we create them.


joaopapereira:
Looks good

PR #140, opened 2015-12-10
Connects to AgileVentures/MetPlus_tracker#129

patmbolger:
Completed agency_people show, edit and update actions.

### DIFF
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -1,0 +1,40 @@
+class AgencyPeopleController < ApplicationController
+  def create
+  end
+
+  def new
+  end
+
+  def show
+    @agency_person = AgencyPerson.find(params[:id])
+  end
+
+  def edit
+    @agency_person = AgencyPerson.find(params[:id])
+  end
+
+  def update
+    @agency_person = AgencyPerson.find(params[:id])
+    @agency_person.assign_attributes(agency_person_params)
+    
+    if @agency_person.valid?
+      @agency_person.save
+      flash[:notice] = "Agency person was successfully updated."
+      redirect_to agency_person_path(@agency_person)
+    else
+      @model_errors = @agency_person.errors
+      render :edit
+    end
+  end
+
+  def destroy
+  end
+  
+  private
+  
+  def agency_person_params
+    params.require(:agency_person).permit(:first_name, :last_name, :branch_id,
+                          agency_role_ids: [], job_category_ids: [])                    
+  end
+  
+end
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -1,8 +1,25 @@
 class AgencyPeopleController < ApplicationController
-  def create
-  end
 
   def new
+    # THIS CODE IS NOT READY FOR TESTING
+    @agency = Agency.this_agency(current_user)
+    @agency_person = AgencyPerson.new
+  end
+  
+  def create
+    # THIS CODE IS NOT READY FOR TESTING
+    @agency = Agency.find(params[:agency_id])
+    @agency_person = AgencyPerson.new
+    @agency_person.assign_attributes(agency_person_params)
+    @agency.agency_people << @agency_person
+    if @agency_person.valid?
+      @agency_person.save
+      flash[:notice] = "Person was successfully created."
+      redirect_to agency_admin_home_path
+    else
+      @model_errors = @agency_person.errors
+      render :new
+    end
   end
 
   def show
@@ -22,6 +39,16 @@ class AgencyPeopleController < ApplicationController
       flash[:notice] = "Agency person was successfully updated."
       redirect_to agency_person_path(@agency_person)
     else
+      if @agency_person.errors[:agency_admin]
+        
+        # If the :agency_error was set by the model this means that the agency person
+        # being edited is the sole agency admin (AA), and that role was unchecked in the
+        # edit view. Removing the sole AA is not allowed.
+        # In this case, reset the AA role and add a flash message.
+        
+        @agency_person.agency_roles << AgencyRole.find_by_role(AgencyRole::ROLE[:AA])
+        flash.now[:warning] = 'Admin agency role reset - see error message'
+      end
       @model_errors = @agency_person.errors
       render :edit
     end
@@ -34,7 +61,8 @@ class AgencyPeopleController < ApplicationController
   
   def agency_person_params
     params.require(:agency_person).permit(:first_name, :last_name, :branch_id,
-                          agency_role_ids: [], job_category_ids: [])                    
+                          agency_role_ids: [], job_category_ids: [],
+                          job_seeker_ids: [])                    
   end
   
 end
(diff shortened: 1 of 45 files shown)


app/controllers/agency_people_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    else
+      if @agency_person.errors[:agency_admin]
+        
+        # If the :agency_error was set by the model this means that the agency person
+        # being edited is the sole agency admin (AA), and that role was unchecked in the
+        # edit view. Removing the sole AA is not allowed.
+        # In this case, reset the AA role and add a flash message.
+        
+        @agency_person.agency_roles << AgencyRole.find_by_role(AgencyRole::ROLE[:AA])
+        flash.now[:warning] = 'Admin agency role reset - see error message'
```
joaopapereira: This message do not give much information. If the error message is enough there is no need for this notification.

app/models/agency_person.rb
```diff
@@ -8,4 +8,49 @@ class AgencyPerson < ActiveRecord::Base
...
+  def sole_agency_admin?
+    # Is this person even an admin?
+    return false unless agency_roles.pluck(:role).include? AgencyRole::ROLE[:AA]
+    
+    return false if other_agency_admin?
+    
+    true
+  end
+  
+  def other_agency_admin?
```
joaopapereira: Can't this function use the one in Agency model to retrieve all the users with a specific role?

app/views/agency_people/_form.html.haml
```diff
@@ -0,0 +1,69 @@
...
+  .form-group
+    = person_form.label :agency_roles, 'Roles', class: 'control-label col-sm-2'
+    .col-sm-8
+      = person_form.collection_check_boxes(:agency_role_ids,        |
+                             AgencyRole.all, :id, :role) do |box|   |
+        = box.label(class: 'checkbox disabled col-sm-8 normal-label') do
+          - if disable_agency_admin?(agency_person, box.text)
+            = box.check_box(disabled: true)
+            = box.text
+            <input type='hidden' name='agency_person[agency_role_ids][]' |
```
joaopapereira: You can use the hidden_field helper instead of using the HTML

joaopapereira:
Please review my comment above

app/controllers/agency_people_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    @agency_person = AgencyPerson.new
+  end
+  
+  def create
+    # THIS CODE IS NOT READY FOR TESTING
+    @agency = Agency.find(params[:agency_id])
+    @agency_person = AgencyPerson.new
+    @agency_person.assign_attributes(agency_person_params)
+    @agency.agency_people << @agency_person
+    if @agency_person.valid?
```
amba178: you do not need to check if valid?. You can just say if @agency_person.save.

app/controllers/agency_people_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+    @agency_person = AgencyPerson.find(params[:id])
+  end
+
+  def edit
+    @agency_person = AgencyPerson.find(params[:id])
+  end
+
+  def update
+    @agency_person = AgencyPerson.find(params[:id])
+    @agency_person.assign_attributes(agency_person_params)
```
amba178: same here, you can just say if @gency_person.update_attributes(agency_person_params).
you can get rid of valid? and @agency_person.save lines.

app/views/agency_people/_form.html.haml
```diff
@@ -0,0 +1,69 @@
...
+          = box.text
+          
+  .form-group
+    .col-sm-2.col-sm-offset-2
+      = submit_tag submit, method: :patch, class: 'btn btn-primary'
+    .col-sm-2.col-sm-offset-1
+      = link_to 'Cancel', cancel_url, class: 'btn btn-danger'
+      
+
+
```
amba178: You can do this unless if you think it is  long line code.
= submit_tag submit, method: :patch, class: 'btn btn-primary col-sm-2.col-sm-offset-2'

app/controllers/agency_people_controller.rb
```diff
@@ -0,0 +1,68 @@
...
+        # edit view. Removing the sole AA is not allowed.
+        # In this case, reset the AA role and add a flash message.
+        
+        @agency_person.agency_roles << AgencyRole.find_by_role(AgencyRole::ROLE[:AA])
+        flash.now[:warning] = 'Admin agency role reset - see error message'
+      end
+      @model_errors = @agency_person.errors
+      render :edit
+    end
+  end
```
amba178: in update action there some redundancy, you can cut some line of codes.

app/controllers/agency_people_controller.rb
patmbolger: The reason I added the flash message is that I not only show a model validation error (standard practice), but I also undo an action that the user took on the form - that is, I reset the 'Agency Admin' role checkbox to 'checked' status.  This is not standard practice, and I want the user to be aware of this action taken by the system.

app/models/agency_person.rb
patmbolger: I could but I'm not sure it buys me much.  Here, I start with a known person and proceed to traverse all people in the same agency until I find the first other person that is also an admin.

Agency.agency_admins first has to figure out what agency it's dealing with, and then gather all admins.  This (figuring out the agency) is a needless step for the AgencyPerson method, since I am starting with a known person and hence a known agency.

app/views/agency_people/_form.html.haml
patmbolger: Thanks for the tip.  That helper is not very well documented, but I got it to work.

app/controllers/agency_people_controller.rb
patmbolger: That code is just cut-and-paste and not in use yet - so, will clean up when I develop the action in the next PR.

app/controllers/agency_people_controller.rb
patmbolger: Yes, cleaned it up.

app/controllers/agency_people_controller.rb
patmbolger: Not sure what you're referring to ...

app/models/agency_person.rb
joaopapereira: I am talking about the method other_agency_admin?
I believe that might be the redundant method

patmbolger: Yes, that's what I thought.  My comment still applies.  AgencyPerson#other_agency_admin? just tries to find *one other* admin.  Agency.find_users_with_role (or, Agency.agency_admins) 1) confirms that the person is an AgencyPerson, and 2) returns an array containing *all* agency admins - both of which are overkill for this limited need.

I agree there is some overlap in the respective logic, but not sure it is worth the extra overhead of using the Agency method to avoid that.

joaopapereira: I believe that when we need to repeat some piece of code from one place to another something smell. I do not think that the amount of agency people that we have per agency will make a function that return all agency people with a specific role inefficient. 
Maybe we also have a problem in the Agency.find_user_with_role, this function should not do 2 jobs: 
1 - find the agency of a user      <- this part should be inside the AgencyPerson, where a user know what is his agency
2 - find the users with the role, but given an agency instance

patmbolger: refactored for both comments

PR #142, opened 2015-12-22
Closes AgileVentures/MetPlus_tracker#129

patmbolger:
1. Added devise_invitable
2. New agency person process
  - Agency admin invites new person to confirm account in PETS
  - New person clicks "Accept invitation" in invite email
  - New person enters password (and confirmation), is signed in, and directed to Agency landing page
3. Above process will also accommodate adding a new company person (not yet built)
4. Delete agency person
5. Added 'status' field to AgencyPerson ('Invited', 'Active', 'Inactive) (these will need to be added to CompanyPerson as well)
6. Unit, functional and integration tests


app/controllers/people_invitations_controller.rb
```diff
@@ -0,0 +1,58 @@
+class PeopleInvitationsController < Devise::InvitationsController
```
joaopapereira: The indentation on this file look a bit off

app/models/agency_person.rb
```diff
@@ -3,14 +3,22 @@ class AgencyPerson < ActiveRecord::Base
...
   has_and_belongs_to_many :job_categories, join_table: 'job_specialities'
   has_and_belongs_to_many :job_seekers, join_table: 'seekers_agency_people'
 
   validates_presence_of :agency_id
   
+  STATUS = { IVT:   'Invited',
+             ACT:   'Active',
+             INACT: 'Inactive' }
+          
+  # validates :status, inclusion: STATUS.values
```
joaopapereira: Is this commented for any particular reason?

config/locales/devise_invitable.en.yml
```diff
@@ -0,0 +1,24 @@
+en:
```
joaopapereira: Maybe it would be nice to be able to store this information in the database, because if we need to change it, we need to do a code change.

app/controllers/people_invitations_controller.rb
patmbolger: Fixed.

app/models/agency_person.rb
patmbolger: Uncommented and add validation confirmation to unit test.

config/locales/devise_invitable.en.yml
patmbolger: Perhaps, but wouldn't we lose the ability to easily support other languages?  (Client at one point asked about supporting languages other than English).

If we want to put in DB, I'd prefer to do that as a separate story.

PR #144, opened 2015-12-24
Js registration implementation

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#130

1. Jobseeker Registration process[create,update,list,show] with views for all the functions, Rspec and Cucumber tests.


joaopapereira:
Please rebase your code

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+   
+   
+  end
+ 
+  def update
+    @jobseeker = JobSeeker.find(params[:id])
+  
+    if @jobseeker.update_attributes(jobseeker_params)
+      current_user = @jobseeker.user
+      sign_in :user, current_user
```
joaopapereira: Do we need to sign in the user again?

config/routes.rb
```diff
@@ -18,6 +20,19 @@
   
   get 'agency_admin/home', path: '/admin/agency_admin/home'
 
+ 
+
+  resources :job_seekers #, path: '/job_seekers/'
```
joaopapereira: Please remove the commented part if we will not need them

features/step_definitions/jobseeker_steps.rb
```diff
@@ -0,0 +1,26 @@
...
+    hash['confirmed_at'] = Time.now
+        
+    jobseeker = JobSeeker.new(hash)
+    jobseeker.save
+  end
+end
+
+
+Given(/^I am on the Jobseeker Registration page$/) do
+  #pending # express the regexp above with the code you wish you had
```
joaopapereira: This step definition should be in navigation steps, because it moves the user to a place

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+  
+  def new
+    @jobseeker = JobSeeker.new
+  end
+  
+  def create
+    @jobseeker = JobSeeker.new(jobseeker_params)
+    if @jobseeker.save
+       flash[:notice] = "A message with a confirmation and link has been sent to your email address. Please follow the link to activate your account."
+    redirect_to root_path
```
joaopapereira: incorrect indentation

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -0,0 +1,65 @@
+require 'rails_helper'
+
+RSpec.describe JobSeekersController, type: :controller do
```
joaopapereira: Missing post tests

joaopapereira:
Please review my previous comments

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    else
+     render 'new'
+    end
```
patmbolger: Need to set up the model errors for display - as in the update action:

    @model_errors = @jobseeker.errors

(put before 'render :new')

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+
+  def show
+    @jobseeker = JobSeeker.find(params[:id])
+     
+   
+  end
+
+  def destroy
+    @jobseeker = Jobseeker.find(params[:id])
+    @jobseeker.destrory
```
patmbolger: misspelled 'destroy' method.

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    redirect_to root_path
```
patmbolger: recommend adding a flash[:notice] message before redirect (.e.g. 'Job Seeker Deleted.')

app/views/job_seekers/show.html.haml
```diff
@@ -0,0 +1,24 @@
...
+		= @jobseeker.first_name
+		%br
+		= @jobseeker.last_name
+		%br
+		= @jobseeker.email
+		%br
+		= @jobseeker.phone
+		%br
+		= @jobseeker.year_of_birth
+
```
patmbolger: We will need buttons for 'Edit' and 'Delete' actions.

(or, we can add as a separate story if you don't want to add those now)

app/views/layouts/_menu.html.haml
```diff
@@ -28,7 +28,12 @@
           %li
             %div{:class => 'navbar-text'}
               Hello,
-              = link_to current_user.first_name, edit_user_registration_path(current_user)
+              -if current_user.actable_type == 'JobSeeker' 
+                = link_to current_user.first_name, edit_job_seeker_path(current_user.actable_id)
+              -else current_user
```
patmbolger: Don't need 'current user' in *else* clause.

config/environments/development.rb
```diff
@@ -47,5 +47,7 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
+  config.action_mailer.delivery_method = :sendmail
```
patmbolger: You don't need this line.

We have been using a gmail account for development environment.  Please continue to use that.

The ActionMailer configuration is defined in config/environment.rb.

The parameters used in that file are defined in config/application.yml.  That file defines ENV (environmental) variables used in the application.  Since these vars tend to be private (like email account name and password), the file is not included in git. 

You need to do this at the command line (assuming you are in root directory of the app):

cp config/application.yml.example  config/application.yml

Then, replace the placeholder values with actual values.  I emailed you those values this morning (to your yahoo account).

If you have any trouble with this let me know and we can PP on Monday.

config/environments/test.rb
```diff
@@ -39,6 +39,7 @@
 
   # Raises error for missing translations
   # config.action_view.raise_on_missing_translations = true
-
+  config.action_mailer.delivery_method = :sendmail
```
patmbolger: See comments above for development.rb - those apply here too.

Also, we do not want to send emails in the test environment, as there could be many emails per rspec/cucumber run, and the tests could be run many times a day.

You don't need to send actual emails for testing, in any case.  In fact, we have been testing without emails being sent since the start (see line 32 in this same file).

For cucumber tests, you can use the methods provided by the email_spec gem.

features/job_seeker.feature
```diff
@@ -0,0 +1,52 @@
...
+  Given I am on the Jobseeker Registration page
+  When I fill in "First Name" with "test"
+  And I fill in "Last Name" with "js80"
+  And I fill in "Email" with "testjobseeker80@gmail.com"
+  And I fill in "Phone" with "345-890-7890"
+  And I fill in "Password" with "password"
+  And I fill in "Password Confirmation" with "password"
+  And I fill in "Year Of Birth" with "1990"
+  Then I click the "Create Job seeker" button 
+  Then I wait for 30 seconds
```
patmbolger: After changes to your email configuration, you should not need to wait as emails will not actually be sent out - so "sending" an email will be close to instantaneous.

features/step_definitions/jobseeker_steps.rb
```diff
@@ -0,0 +1,26 @@
...
+    
+    jobseeker = hash.delete 'jobseeker'
+    hash['actable_type'] = 'JobSeeker'
+    hash['password_confirmation'] = hash['password']
+    hash['confirmed_at'] = Time.now
+        
+    jobseeker = JobSeeker.new(hash)
+    jobseeker.save
+  end
+end
```
patmbolger: Strongly recommend moving this method to fixtures_steps.rb - that file is meant to have all such data-loading steps, and we should have all of those steps in one place.

spec/controllers/job_seekers_controller_spec.rb
patmbolger: Also, missing PATCH (#update) tests.

app/controllers/job_seekers_controller.rb
vijayakarumudi: we do not need to sign in  the user again.I removed the line.

config/routes.rb
vijayakarumudi: I removed all the commented parts.

app/controllers/job_seekers_controller.rb
vijayakarumudi: I fixed the indentation.

app/controllers/job_seekers_controller.rb
vijayakarumudi: I fixed this line of code.

app/controllers/job_seekers_controller.rb
vijayakarumudi: I fixed this one.

app/views/job_seekers/show.html.haml
vijayakarumudi: Create a separate story for Edit and Delete actions.

features/step_definitions/jobseeker_steps.rb
vijayakarumudi: I added this step definition to the navigation steps.

app/controllers/job_seekers_controller.rb
vijayakarumudi: I added the flash message before redirect.

app/views/layouts/_menu.html.haml
vijayakarumudi: I removed.

features/step_definitions/jobseeker_steps.rb
vijayakarumudi: I moved above code to  fixtures_steps.rb

app/controllers/job_seekers_controller.rb
amba178: I think  we do need to sign_in @jobseeker.user in order to @jobseeker stay as still signed in user otherwise the Devise will authenticate when we submit as update and at the authentication current_user id and submitted id will be different thus devise will log out  the user.  Since we know @jobseeker, it act as user but not really user. It will make since if we demonstrate it, i guess.

joaopapereira: Do we need to sign in when the user is updated? Dont understand why

patmbolger: Please replace the above two lines with:

    sign_in :user, @jobseeker.user, bypass: true

Note that 'current_user' is a Devise method that returns the current, logged-in User object.  This name should not be used for a local variable.

Also, 'bypass: true' is the recommended option for this situation - see "Solution 3" here: https://github.com/plataformatec/devise/wiki/How-To:-Allow-users-to-edit-their-password

(more info here: http://www.rubydoc.info/github/plataformatec/devise/master/Devise/Controllers/SignInOut#sign_in-instance_method)

app/controllers/job_seekers_controller.rb
```diff
@@ -0,0 +1,60 @@
...
+    if @jobseeker.update_attributes(jobseeker_params)
+      current_user = @jobseeker.user
+      sign_in :user, current_user
+      flash[:notice] = "Jobseeker was updated successfully."
+      redirect_to root_path
+    else
+      @model_errors = @jobseeker.errors
+      render 'edit'
+    end
+  end
```
patmbolger: Please note that the way this is currently configured, the user has to either reenter their password every time they make a change in the 'edit' form, even if they never change their password.  

We need to make password entry a requirement *only if* the use wants to change their password.

For example, the standard Devise user registration form looks like this:

screen shot 2015-12-30 at 2 54 43 pm

In  your current form, I suggest you:

- strip out the password and password_confirmation values from the jobseeker_params hash, *if* the password field is blank (which would indicate that the user does not want to change their password.  That will prevent 'password can't be blank' model errors and allow the rest of the changes to be saved.  You can do this by something like:

````    
    person_params = jobseeker_params
    if not person_params.value? 'password'
        person_params.delete('password')
        person_params.delete('password_confirmation')
    end
````

(I did not test the code above).

- add a message for the Password field like the one in the standard form, telling the user not to enter a password unless they want to change their password.

File: app/controllers/job_seekers_controller.rb (hunk 1 above)
Comment (patmbolger) [inline app/controllers/job_seekers_controller.rb]: @João - please the link above from the Devise wiki.  We are using 'solution 3' here.

Comment (amba178) [inline app/controllers/job_seekers_controller.rb]: @patmbolger: That is perfectly right, bypassing the validation, but look the implementation very carefully 
I think this is applies on the password strong parameters
def user_params
    # NOTE: Using `strong_parameters` gem
    params.require(:user).permit(:password, :password_confirmation)
 end
is that mean we want to bypass validation for other attributes too?

Comment (patmbolger) [inline app/controllers/job_seekers_controller.rb]: @salem - not sure what you mean, but we do NOT want to bypass validation for any entered data.  What we do want is to prevent the model validation failing due to empty password and password_confirmation - when the use did not mean to change the password and hence those fields are correctly left blank in the form.

If the use entered something in the password field, then we absolutely will validate those values.

Comment (amba178) [inline app/controllers/job_seekers_controller.rb]: As I said earlier you are perfectly right, I was just wandering whether it will bypass the rest of validation assuming we are using devise validation since we are supplying more strong parameters to :user  
@patmbolger: what table/tables did you make change on?

Comment (patmbolger) [inline app/controllers/job_seekers_controller.rb]: I did not make any changes to any tables.  

If the password is not changed (empty field from the form), then we can strip out the password and password_confirmation from the params hash.  This prevents model validation for those fields when calling 'update_attributes'.  Any other fields *will* be validated - including User fields such as first_name, email, etc.

Comment (amba178) [inline app/controllers/job_seekers_controller.rb]: Probably not, I might be wrong. Good answer though.
Sorry for confusion,  the table I was talking about does not relate to devise, I was talking about the  png you posted on slack.

Comment (joaopapereira) [inline app/controllers/job_seekers_controller.rb]: I AM OK with the solution just wondering why we needed to sign in. But if it is a decide mandatory call i AM ok

Comment (patmbolger) [inline app/controllers/job_seekers_controller.rb]: Explored this question this morning (need to login).  What I found is that if a JobSeeker is updated, this occurs:

1) If the password is changed (or just overwritten with current password), the user parameters in the *session* are **removed**.

2) If the password is not updated, the *session* user parameters are updated to reflect the updated attribute values.

The first behavior causes Devise, upon next render, to determine that the user has logged out.

The second behavior results in Devise determining that the user is still logged in.

In the job seeker #update action, if we strip the 'password' and 'password_confirmation' parameters from the attribute hash, then no sign_in is required.  Otherwise, we need to call sign_in to force a refresh of the *session*.  

Either way, calling sign_in with 'bypass: true' is the recommended way to do this, as this option bypasses a bunch of callbacks that otherwise would be called.

Comment (vijayakarumudi) [inline app/controllers/job_seekers_controller.rb]: I used the recommended solution and tested the functionality. It's working as expected.

File: app/controllers/job_seekers_controller.rb (hunk 15 above)
Comment (vijayakarumudi) [inline app/controllers/job_seekers_controller.rb]: I will work on this story #153 later.

File: features/job_seeker.feature (hunk 13 above)
Comment (vijayakarumudi) [inline features/job_seeker.feature]: I removed the wait time.

File: config/environments/development.rb (hunk 11 above)
Comment (vijayakarumudi) [inline config/environments/development.rb]: I configured the application.yml file based on instructions sent through the mail.

File: config/environments/test.rb (hunk 12 above)
Comment (vijayakarumudi) [inline config/environments/test.rb]: I removed this line of code.

File: spec/controllers/job_seekers_controller_spec.rb (hunk 5 above)
Comment (vijayakarumudi) [inline spec/controllers/job_seekers_controller_spec.rb]: I added the post  and update tests.

**Comment by vijayakarumudi** [comment]
This is Jobseeker PR for create and update tests.

**Comment by vijayakarumudi** [comment]
JobSeeker Registration process provides complete CRUD actions, Rspec and Cucumber test scenarios.

File: app/views/devise/sessions/new.html.haml [hunk 16]
```diff
@@ -5,7 +5,7 @@
     %p{id: 'account-link'}
       Log In to your account
       %i{class: "fa fa-minus fa-4x fa-rotate-90"}  
-      #{link_to 'Register as jobseeker', '#'} 
+      #{link_to 'Register as jobseeker', 'job_seekers/new'}
```
patmbolger: Suggest using the helper method 'new_job_seeker_path' (instead of 'job_seekers/new') - in case we ever change the routing for that action.

vijayakarumudi: I added the helper method.

vijayakarumudi:
updates for the helper method

joaopapereira:
looks good.
good job Vijaya

PR #145, opened 2015-12-30
Closes AgileVentures/MetPlus_tracker#148

patmbolger:
Restructured JobSeeker <> AgencyPerson association to accommodate many-to-many relationships based on role of AgencyPerson (role defines the type of association between JS and AP).


joaopapereira:
I am OK with this change

PR #148, opened 2016-01-13
Enhance job seeker update#153

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#153.

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -22,9 +22,8 @@ class JobSeekersController < ApplicationController
  
   def update
     @jobseeker = JobSeeker.find(params[:id])
-    
     person_params = jobseeker_params
-    if person_params['password'] =='' 
+    if person_params['password'].to_s.length == 0 
        person_params.delete('password')
        person_params.delete('password_confirmation')
     end
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -22,7 +22,13 @@ class JobSeekersController < ApplicationController
  
   def update
     @jobseeker = JobSeeker.find(params[:id])
-    if @jobseeker.update_attributes(jobseeker_params)
+    
+    person_params = jobseeker_params
+    if person_params['password'] =='' 
+       person_params.delete('password')
+       person_params.delete('password_confirmation')
+    end
+    if @jobseeker.update_attributes(person_params)
        sign_in :user, @jobseeker.user, bypass: true
        flash[:notice] = "Jobseeker was updated successfully."
        redirect_to root_path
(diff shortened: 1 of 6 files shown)


app/controllers/job_seekers_controller.rb
```diff
@@ -22,14 +22,19 @@ def edit
...
   def update
     @jobseeker = JobSeeker.find(params[:id])
-
-    if @jobseeker.update_attributes(jobseeker_params)
-      sign_in :user, @jobseeker.user, bypass: true
-      flash[:notice] = "Jobseeker was updated successfully."
-      redirect_to root_path
+    
+    person_params = jobseeker_params
+    if person_params['password'] ==''
```
joaopapereira: Space missing between ++ and  ''
This could have been archived by using the length of the string, but is ok

app/views/job_seekers/_form.html.haml
```diff
@@ -29,7 +30,7 @@
 		=js.label :year_of_birth,'Year Of Birth',class:'col-sm-2 control-label text-right'
 	%div{class: 'form-group'}
 		=js.text_field :year_of_birth, class:'form-control'
-
+
```
joaopapereira: Do we need this change?

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -63,8 +63,9 @@
                
    context "valid attributes" do
      before(:each) do
-       @jobseeker =  FactoryGirl.create(:job_seeker) 
+       @jobseeker =  FactoryGirl.create(:job_seeker)
```
joaopapereira: Maybe we should have a new test added here to check the change

joaopapereira:
Please review my comments above

features/job_seeker.feature
```diff
@@ -40,8 +40,17 @@ Scenario: edit Js Registration
...
-     
 
+Scenario: edit Js Registration without password
+  Given I am on the home page
+  And I login as "vijaya.karumudi@gmail.com" with password "password"
+  Then I should see "Signed in successfully"
+  When I click the "vijaya" link
+  And I fill in "First Name" with "vijaya1" 
+  And I fill in "Password" with ""
+  And I fill in "Password Confirmation" with ""
```
patmbolger: These two lines violate the intent of the test - which is to confirm that the password (and confirmation) fields do *not* have to be touched at all unless the password is being changed.   That is, the edit view should show both of these fields as empty - and the user should not have to delete any content in those fields as a result.

vijayakarumudi:
Thanks  for your comments. I will work on these things.

app/controllers/job_seekers_controller.rb
vijayakarumudi: I fixed this one. I added the functionality of length of the string

app/views/job_seekers/_form.html.haml
vijayakarumudi: I added status field here. and I removed the status filed. It got changed.

features/job_seeker.feature
vijayakarumudi: I removed the those lines. we are not filling any password in this test.

spec/controllers/job_seekers_controller_spec.rb
vijayakarumudi: I added new test here and it passed.

vijayakarumudi:
close this request

vijayakarumudi:
close this request

PR #153, opened 2016-01-19
Closes AgileVentures/MetPlus_tracker#163

patmbolger:


app/controllers/agency_controller.rb
```diff
@@ -1,4 +1,9 @@
 class AgencyController < ApplicationController
+  # This controller manages activities for agency people who
+  # do not have admin rights.
```
joaopapereira: Isn't this comment a contradiction?

app/controllers/company_people_controller.rb
```diff
@@ -1,32 +1,48 @@
...
   end
 
   def update
+    @company_person = CompanyPerson.find(params[:id])
+
+    if @company_person.update_attributes(company_person_params)
+      flash[:notice] = "Company person was successfully updated."
+      redirect_to company_person_path(@company_person)
+    else
+      unless @company_person.errors[:company_admin].empty?
```
joaopapereira: Just a comment I prefer to see if statements in the code because it is harder to think in the negative form.

app/controllers/company_registrations_controller.rb
```diff
@@ -44,17 +48,40 @@ def create
...
+
   def approve
     # Approve the company's registration request.
     # There should be only one CompanyPerson associated with the company -
     # this is the 'company contact' included in the registration request.
     company = Company.find(params[:id])
-    company_person = company.company_people[0]
-
     company.status          = Company::STATUS[:ACT]
+    company.save
```
joaopapereira: What happens if this fails?

app/controllers/company_registrations_controller.rb
```diff
@@ -44,17 +48,40 @@ def create
...
+    if @company.update_attributes(reg_params)
+      flash[:notice] = "Registration was successfully updated."
+      redirect_to agency_admin_home_path
+    else
+      @model_errors = @company.errors
+      render :edit
+    end
+  end
+
   def approve
```
joaopapereira: Approve and deny look very alike maybe we should create a spike to redactor this into a function

joaopapereira:
Please review my previous comments

app/controllers/agency_controller.rb
patmbolger: The controller *agencies_controller.rb* manages the usual CRUD actions, for agencies, that are associated with an agency_admin (e.g. edit, update).  Eventually we will also add actions that are available to the "super" admin of PETS (create, delete an agency).

So, we need a controller that will manage actions associated with a regular agency person, which would not include CRUD actions on agencies.  The first one here is just a starting point - show a landing page for an agency person (routed to on login) that has information and actions available suitable to that person.  We will add other actions as we go.  For instance, show job seekers, for which this person is the job developer, who have been inactive for some period to time.  (there are lots of features we will need to add, consistent with the system requirements docs we developed earlier).

There is also an analog for this for companies.

app/controllers/company_registrations_controller.rb
patmbolger: They do look alike to some degree, in that they both set company and company contact attributes and send out notification emails (2 in the case of *approve*, one for *deny*).  One fundamental difference is that *approve* does a standard rendering whereas *deny* responds to an AJAX call with a rendered partial in the data payload (*approve* is called via standard HTTP and *deny* is called via HTTP XHR).

I'm afraid that one function for all this would have some many "if-else" paths as to be more confusing than helpful.

app/controllers/company_registrations_controller.rb
patmbolger: That is a good question - and is also applicable to many more controllers and controller actions.

Generally, the problem usually is "what happens if the object *find* method at the start of the action fails?".  We do this everywhere and do not account for failure of the *find* (   e.g. AgencyPerson.find(params[:id])    ).

I am OK with this because we are always (?) coming into the action from a link or button on a view that is showing (or doing something else) with the object we will be "finding" in the next (*this*) action.  So, for *this* action to fail the prior action would also have failed.  Any, bugs that might lead to this should (hopefully) be found by integration tests.

In this particular case, if the company *find* succeeds then setting it's 'status' attribute should not cause a save failure (unless we have a fundamental problem in our model, which should be caught by unit test).

Anyway, we should discuss as a team and adopt the same philosophy throughout.

patmbolger:
Thanks for the thoughtful comments .... please see my responses

app/controllers/agency_controller.rb
joaopapereira: I am talking about the comment. because it says:
people who
 +  # do not have admin rights

And then:
home' is the agency 'landing page', where admin people
 +  # are sent upon login

They are contradicting

patmbolger: Ah yes .... fixed.

PR #155, opened 2016-01-21

amba178:
seeding should work for production and development env without throwing errors.


Gemfile
```diff
@@ -2,6 +2,7 @@ source 'https://rubygems.org'
 
 
 ruby '2.1.8'
+# ruby '2.2.2'
```
patmbolger: Are we going back to Ruby 2.2.2? 

A little while ago we set our version to 2.1.8.  I thought that had to do with Heroku(?).

Anyway, I would rather be on the latest version but can we confirm that this is OK for Heroku?

amba178: I think, with my local repo deploying to Heroku works for both versions, but we should check with @joaopapereira. He has the cut.

joaopapereira: Was there any special reason for upgrading the version of Ruby?
What is the version you guys are using?

config/environments/production.rb
```diff
@@ -1,5 +1,4 @@
-Rails.application.configure do
```
joaopapereira: Why remove this?

joaopapereira:
Please review my comments

Gemfile
amba178: if I recall, we been working with ruby-2.2.2 version. Reason? I do not think there are any special reasons for using 2.2.2 other than other both have different version of dependencies for the most part. I prefer working with on most updated ruby version,  personally.

config/environments/production.rb
amba178: Is not removed. Still there, but maybe I deleted spaces either above or below the code. I am not sure why it shows as deleted.

Gemfile
patmbolger: We have not ben specifying a Ruby version in the Gemfile until a commit that occurred 10 days ago - https://github.com/AgileVentures/MetPlus_PETS/commit/1f50c19c5239c676984814da086cccb11dc8d780

I thought that might have had to do with the effort to stand up the app on Heroku (?).

I would prefer we do *not* specify a Ruby version if we do not have to.

(I was using 2.2.2 until this change occurred, current stable version appears to be 2.3)

PR #156, opened 2016-01-22
connects AgileVentures/MetPlus_tracker#162

amba178:
- Devise controllers overwrite not needed, as it provides helper method.
Had to modified to fit our need.


app/controllers/application_controller.rb
```diff
@@ -1,21 +1,24 @@
...
-    }
-  end
+    def configure_permitted_parameters
+      devise_parameter_sanitizer.for(:sign_up) << :first_name
+      devise_parameter_sanitizer.for(:sign_up) << :last_name    
+      devise_parameter_sanitizer.for(:sign_up) << :phone
+        
+      [:first_name,:last_name, :phone].each{ |field|
+        devise_parameter_sanitizer.for(:account_update)<<field  
+      }
```
patmbolger: Not a big deal but this might be easier to read if consolidates into one block:

```
[:first_name, :last_name, :phone].each  { | field |
  devise_parameter_sanitizer.for(:account_update) << field 
  devise_parameter_sanitizer.for(:sign_up) << field 
end

File: app/helpers/sessions_helper.rb [hunk 2]
```diff
@@ -0,0 +1,12 @@
...
+	def redirect_back_or(default)
+		redirect_to(session[:forwarding_url] || default)
+		session.delete(:forwarding_url)
+	end
+
+	# Stores the URL trying to be accessed.
+	def store_location
+		session[:forwarding_url] = request.url if request.get?
+	end
+end
```
patmbolger: Are we using this code anywhere?  (I can only see a redirect :back) for the login form *Cancel* action)

app/controllers/application_controller.rb
```diff
@@ -1,21 +1,24 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  protect_from_forgery with: :exception
+  protect_from_forgery with: :exception 
 
   before_action :configure_permitted_parameters, if: :devise_controller?
+  before_action :store_current_location, :unless => :devise_controller?
```
patmbolger: Are we using the stored location anywhere?

app/controllers/application_controller.rb
amba178: Yes. I agree.

app/helpers/sessions_helper.rb
amba178: In short NO. Initial, I wanted to overwrite the devise controllers, and created the above module,  sessionHelper methods end up not using it. I am not using it any where.

app/controllers/application_controller.rb
amba178: we are not using store_location anywhere.

patmbolger: Understood - I recommend removing this and any other unused code from this PR since we don't want unused and untested code in the codebase.

joaopapereira: I agree with Patrick

app/controllers/application_controller.rb
```diff
@@ -1,21 +1,25 @@
...
-    [:first_name,:last_name, :phone].each{ |field|
-      devise_parameter_sanitizer.for(:account_update)<<field  
-    }
-  end
+      [:first_name,:last_name, :phone].each{ |field|
+        devise_parameter_sanitizer.for(:account_update)<<field  
+      }
+
+      [:first_name,:last_name, :phone].each{ |field|
+        devise_parameter_sanitizer.for(:sign_up)<<field
```
joaopapereira: is these parameters will always be the same for signup and account_update we should use only 1 loop

config/routes.rb
```diff
@@ -4,6 +4,7 @@
                                      :sign_in => 'login' },
                      :controllers => { :invitations => 'people_invitations' }
 
+
```
joaopapereira: Is this line needed?

features/step_definitions/navigation_steps.rb
```diff
@@ -14,10 +16,11 @@
   end
 end
 
-Given(/I am on the (\w+) page$/) do |page|
+Given(/I am on the (.+) page$/) do |page|
```
joaopapereira: What is the reason behind this change?

joaopapereira: What is the reason for this change?

features/step_definitions/users_steps.rb
```diff
@@ -12,11 +12,15 @@
 end
 
 
-When(/^I visit profile for "(.*?)"$/) do |first_name|
+When(/^I visit profile for "(\w+)"$/) do |first_name|
    user = User.find_by_first_name!(first_name)
    visit  "/users/edit.#{user.id}"
 end
 
+# Given(/^I am on edit page for "(\w+)"$/) do |name|
```
joaopapereira: Why add this comment?

joaopapereira:
Please review my comments

features/step_definitions/navigation_steps.rb
amba178: I was trying to match something like this: 
"Company Registration" 
so (.+) or (.*?)  will do it for our purpose.
I think (\w+) will not work because the space between Company and Registration

config/routes.rb
amba178: Yes. I believe we need it for invitable module devise extension

PR #159, opened 2016-01-25

amba178:


config/environments/production.rb
```diff
@@ -80,4 +80,6 @@
   config.action_mailer.delivery_method = :sendmail
   #this will take care of devise setup default url in production
   config.action_mailer.default_url_options = { :host => 'metplus.org' }
-end
```
joaopapereira: Did you remove the end for any particular reason?

joaopapereira:
If in this PR we could get some users on the seed file it would help us to show the client the app in heroku, can you provide them @amba178 ?

config/environments/production.rb
amba178: Thanks, for pointing out.

amba178:
@joaopapereira: Sure, I can create some users.

amba178:
Error due to ruby version, try to close and open again.

PR #160, opened 2016-01-25
connects AgileVentures/MetPlus_tracker#162

amba178:


config/routes.rb
```diff
@@ -4,6 +4,7 @@
                                      :sign_in => 'login' },
                      :controllers => { :invitations => 'people_invitations' }
 
+
```
joaopapereira: Do we really need this line here?

db/seeds.rb
```diff
@@ -40,14 +40,14 @@
...
 
-# User.find_or_create_by(email: 'salemamba@gmail.com') do |user|
-#     user.first_name = "salem"
-#     user.last_name  = 'amba'
-#     user.password = 'secret123'
-#     user.password_confirmation = 'secret123'
-#     user.phone ='619-316-8971'
-#     user.confirmed_at = DateTime.now
-# end
+User.find_or_create_by(email: 'salemamba1@gmail.com') do |user|
```
joaopapereira: Why are you creating a generic user? Should it be associated directly with a role?

features/manage_users.feature
```diff
@@ -1,34 +1,51 @@
...
+	As a user 
+	I want to login 
+	So that I can edit my profiles
 
-#@focus
-Scenario Outline: Show edit link as salem
-	Given the following user records:
+	#@focus
+	Background:
+		Given the following user records:
```
joaopapereira: I believe this test is not 100% valid, because out users are never "User" are always of another type, so we should do tests with JS, JD, CM instead of plain users

joaopapereira:
I believe there are unnecessary changes in this PR, because the focus of this Story is just to add the back button.
Nevertheless please review my comments

config/routes.rb
amba178: if you referring to line 7, than No. I will delete it.

db/seeds.rb
amba178: I was seeding the file, after updating the jobs record and  forgot to comment it out. i will fix it . thanx.

features/manage_users.feature
amba178: You are definitely CORRECT, but as right now the edit link does render to respective USERS. I mean by USERS(JS, JD, CM). Once we fixed the edit link, than I can update the above story.   I think for now we can leave  as it is.  I will create a ticket for it.

amba178:
@joaopapereira , there are two parts for it. Not only back button.

amba178:
I have updated all comments.

joaopapereira:
Looks good going to merge

PR #162, opened 2016-01-28

amba178:


amba178:
I have add puma as our webserver in production instead of webrick. 
@joaopapereira: if I specify the ruby version 2.2.2, travis ci complains. You are very familiar with travis, but I was wondering if the travis is locked to ruby version 2.1.8?

joaopapereira:
I am OK with the change
For teachers you need to change the Travis file as well or else it will not work.
I will accept the PR as is but please do change the version to 2.2.2 @amba178

amba178:
Ok. I will make sure to look at the Travis file and make changes as needed on the next PR.

PR #163, opened 2016-01-28
Job seeker status field #142

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#142

Adding Jobseeker Status field to creating new jobseeker form  and edit jobseeker form and added related  test for Cucumber and Rspec.

### DIFF
--- a/app/views/job_seekers/_form.html.haml
+++ b/app/views/job_seekers/_form.html.haml
@@ -28,7 +28,7 @@
 	%div{class: 'form-group'}
 		=js.text_field :year_of_birth, class:'form-control'
 	%div{class: 'label-group'}
-		=js.label :value,'Status',class: 'col-sm-2 control-label text-right'
+		=js.label :job_seeker_status_id,'Status',class: 'col-sm-2 control-label text-right'
 	%div{class:'form-group'}
 		=js.collection_select(:job_seeker_status_id,JobSeekerStatus.all, :id, :value,{prompt: true})
 	%div{class: 'label-group'}
--- a/app/views/job_seekers/_form.html.haml
+++ b/app/views/job_seekers/_form.html.haml
@@ -28,6 +28,10 @@
 	%div{class: 'form-group'}
 		=js.text_field :year_of_birth, class:'form-control'
 	%div{class: 'label-group'}
+		=js.label :value,'Status',class: 'col-sm-2 control-label text-right'
+	%div{class:'form-group'}
+		=js.collection_select(:job_seeker_status_id,JobSeekerStatus.all, :id, :value,{prompt: true})
+	%div{class: 'label-group'}
 		=js.label :resume,'Resume',class:'col-sm-2 control-label text-right'
 	%div{class: 'form-group'}
 		=js.file_field :resume
--- a/features/job_seeker.feature
+++ b/features/job_seeker.feature
@@ -6,6 +6,13 @@ Feature: Jobseeker Management and performs different functions
 
 Background: seed data added to database
 
+  Given the following jobseekerstatus values exist:
+  | value                | description |
+  | Unemployedlooking    | A jobseeker without any work and looking for a job|
+  | Employedlooking      | A jobseeker with a job and looking for a job      | 
+  | Employednotlooking   | A jobseeker with a job and not looking for a job for now.|
+
+
   Given the following jobseeker exist:
   | first_name| last_name| email                      | phone       |password  |password_confirmation| year_of_birth |
   | vijaya      | karumudi     | vijaya.karumudi@gmail.com | 345-890-7890| password   |password               | 1990          |
@@ -20,6 +27,7 @@ Scenario: new Js Registration
   And I fill in "Password" with "password"
   And I fill in "Password Confirmation" with "password"
   And I fill in "Year Of Birth" with "1990"
+  Then I select "Unemployedlooking" in select list
   Then I click the "Create Job seeker" button
   Then I should see "A message with a confirmation and link has been sent  to your email address. Please follow the link to activate your account."
 
@@ -35,6 +43,7 @@ Scenario: edit Js Registration
   Then I should see "Signed in successfully"
   When I click the "vijaya" link
   And I fill in "First Name" with "vijaya1"
+  Then I select "Employedlooking" in select list
   And I fill in "Password" with "password"
   And I fill in "Password Confirmation" with "password"
   Then I click the "Update Job seeker" button
@@ -46,6 +55,7 @@ Scenario: edit Js Registration without password change
   Then I should see "Signed in successfully"
   When I click the "vijaya" link
   And I fill in "First Name" with "vijaya1" 
+  Then I select "Employednotlooking" in select list
   Then I click the "Update Job seeker" button 
   Then I should see "Jobseeker was updated successfully."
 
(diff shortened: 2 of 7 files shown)


features/job_seeker.feature
```diff
@@ -20,6 +27,7 @@ Scenario: new Js Registration
   And I fill in "Password" with "password"
   And I fill in "Password Confirmation" with "password"
   And I fill in "Year Of Birth" with "1990"
+  Then I select "Unemployedlooking" in select list
```
joaopapereira: Are we defaulting a value on this select if we do not have a default status what happens?

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -19,7 +19,8 @@
      before(:each) do
        @jobseeker = FactoryGirl.create(:job_seeker)
        @user = FactoryGirl.create(:user)
-       jobseeker_hash = FactoryGirl.attributes_for(:job_seeker).merge(FactoryGirl.attributes_for(:user))
```
joaopapereira: Shouldn't we have a test to check if the we do not add a status we should have an error?

joaopapereira:
The PR looks good
Please review my comments

features/job_seeker.feature
vijayakarumudi: We are defaulting  one of the values allowed for jobseeker status.

spec/controllers/job_seekers_controller_spec.rb
vijayakarumudi: Previously we do not status filed I wrote that test. Now I added Status value it is required and always has a default value in the form.

app/views/job_seekers/_form.html.haml
```diff
@@ -28,6 +28,10 @@
 	%div{class: 'form-group'}
 		=js.text_field :year_of_birth, class:'form-control'
 	%div{class: 'label-group'}
+		=js.label :value,'Status',class: 'col-sm-2 control-label text-right'
+	%div{class:'form-group'}
+		=js.collection_select(:job_seeker_status_id,JobSeekerStatus.all, :id, :value,{prompt: true})
```
amba178: you can add at the end as html_option to collection_select .........., {class: "form-control"}. Good work!

features/step_definitions/web_steps.rb
```diff
@@ -84,3 +84,10 @@ def search_text text
 When /^I reload the page$/ do
   visit current_path
 end
+
+Then(/^I select "(.*?)" in select list$/) do |item|
+  select(item)
+end
+
```
patmbolger: Please note that there was already a step definition in this file that could have been used (starting at line 77):

    When(/^(?:I|they) select "([^"]*)" in select list "([^"]*)"$/) do | item, list |
      find(:select, list).find(:option, item).select_option
    end

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -83,10 +87,12 @@
       before(:each) do
         @jobseeker =  FactoryGirl.create(:job_seeker)
         @user =  FactoryGirl.create(:user)
+        @jobseekerstatus =  FactoryGirl.create(:job_seeker_status)
         @jobseeker.valid? 
-        patch :update, job_seeker:FactoryGirl.attributes_for(:job_seeker, year_of_birth: '1980').merge(FactoryGirl.attributes_for(:user, first_name:'John',last_name:'Smith',phone:'780-890-8976')),id:@jobseeker
+        patch :update, job_seeker:FactoryGirl.attributes_for(:job_seeker, year_of_birth: '1980').merge(FactoryGirl.attributes_for(:user, first_name:'John',last_name:'Smith',phone:'780-890-8976')).merge(FactoryGirl.attributes_for(:job_seeker_status,value:'Employedlooking')),id:@jobseeker
         @jobseeker.reload
```
patmbolger: This test is to confirm that the user can change profile attributes - other than *password* - and not have to reenter *password* and *password_confirmation* each time.  

However, the parameters being sent to the controller do, in fact, include *password* and *password_confirmation*.

To fix this, you could probably add these arguments:

    password: nil, password_confirmation: nil

to FactoryGirl.attributes_for(:user, .....)

patmbolger: Also, in order to improve readability, I recommend that you 1) separate symbols from their values with white space, 2) separate multiple methods arguments with a space, and 3) use multiple lines for very long statements.  For example, the statement here is so long that Github just clips half of it half and I can't see the other half (unless there is a way to make Github wrap long lines):

    patch :update, job_seeker:FactoryGirl.attributes_for(:job_seeker, year_of_birth: '1980').merge(FactoryGirl.attributes_for(:user, first_name:'John',last_name:'Smith',phone:'780-890-8976')).merge(FactoryGirl.attributes_for(:job_seeker_status,value:'Employedlooking')),id:@jobseeker

A more readable way to write this is:

    patch :update, job_seeker:  FactoryGirl.attributes_for(:job_seeker, year_of_birth: '1980').
      merge(FactoryGirl.attributes_for(:user, first_name: 'John',  last_name: 'Smith',  phone: '780-890-8976')).
      merge(FactoryGirl.attributes_for(:job_seeker_status, value: 'Employedlooking')), id:  @jobseeker

Please be kind to my tired eyes :)

vijayakarumudi:
I will review the comments

features/job_seeker.feature
amba178: I do not thinking defaulting one of the values would be good idea, this can be misleading to employer or jobseeker. I would just prompt the jobseeker maybe asking to select the status options and without selecting would show error while submitting the form.

patmbolger: I think we're OK here for the form, as we are not actually setting a default status value in the job_seekers#new action, and we're using the "prompt: true" option for collection_select.  This latter puts a "Please select" message at the top of the pulldown menu for the selection.

(create Vitaya's branch in your local repo by:

    git fetch vijaya JobSeekerStatus-field-#142:JobSeekerStatus-field-#142
    git checkout JobSeekerStatus-field-#142

and go the /job_seekers/new and you can see that.  (in this command, Vitaya's github repo is is my remote repo named 'vijaya'))

The only thing that is missing from Salem's suggestion is that we are not forcing the presence of the status value in the JobSeeker model.  We should discuss whether we want to do this or not.  In any case, if we do want to do this it would probably be better to do it in another story.

joaopapereira: I believe that the status should be mandatory.
But we will manage that in a new story.

spec/controllers/job_seekers_controller_spec.rb
vijayakarumudi: I changed the update method for more readability.

features/step_definitions/web_steps.rb
vijayakarumudi: I saw this step definition. I wrote new step.

features/job_seeker.feature
vijayakarumudi: create new story for this one

vijayakarumudi:
I am working this PR I fixed most of the comments

vijayakarumudi:
I completed all the changes, now VM is down. I can not update those changes to PR.

joaopapereira:
When you finish the updated PR, please let us know

vijayakarumudi:
This is my updated PR for Jobseeker status

joaopapereira:
Looks good going to merge

PR #165, opened 2016-01-30
Closes AgileVentures/MetPlus_tracker#180

patmbolger:
The update action has these problems:

1. It was not allowing the email address for the company contact to be changed, and was failing silently
2. It was causing the Device confirmable module to send an "email address confirmation" email to the new email address.
This behavior is expected for a typical email address change by a user, since we have configured Devise to require both new and changed email addresses to be confirmed (via an email with a link that passes back a token to confirm the address).  (this is actually the root cause of the first problem noted above - Devise needs to confirm a changed email address before making the change to the User object).

However, we do not want confirmation behavior for a company contact email - while the company is in "pending approval" stage - because the company contact is not yet a bona fide user of PETS.

### DIFF
--- a/app/controllers/company_registrations_controller.rb
+++ b/app/controllers/company_registrations_controller.rb
@@ -62,8 +62,23 @@ class CompanyRegistrationsController < ApplicationController
       reg_params[:company_people_attributes]['0'].delete :password_confirmation
     end
 
+    changed_email = (@company.company_people[0].email !=
+            reg_params[:company_people_attributes]['0'][:email])
+
+    # Do not create a new email confirmation token and
+    # do not send an email confirmation email
+    # (If config.reconfirmable = true in Devise initializer file, the default
+    # behavior will be to create a new token and send confirmation email.)
+    @company.company_people[0].user.skip_reconfirmation!
+
     if @company.update_attributes(reg_params)
       flash[:notice] = "Registration was successfully updated."
+      # Send pending-approval email to company contact if the contact
+      # email address was changed
+      if changed_email
+        CompanyMailer.pending_approval(@company,
+                                @company.company_people[0]).deliver_now
+      end
       redirect_to agency_admin_home_path
     else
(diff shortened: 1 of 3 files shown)


app/controllers/company_registrations_controller.rb
```diff
@@ -62,8 +62,23 @@ def update
       reg_params[:company_people_attributes]['0'].delete :password_confirmation
     end
 
+    changed_email = (@company.company_people[0].email !=
```
joaopapereira: Will we allow the users to change email?
Not sure if that is a good feature

patmbolger: This feature (update company registration) will *only* be available to an agency admin.

The idea is that a site visitor who requests a company registration (registrant) may have entered incorrect information.  The registrant cannot delete the registration, so he or she has to communicate to the agency admin to make changes on their behalf (we have to work on how that communication occurs, but that will be simple).

Typical scenarios might be:

1. The registrant entered an incorrect email (they will be told that they will receive a 'pending registration' email, and when that does not come they might realize their mistake).  The agency admin will go in and fix the email address, at which point the registrant will receive the 'pending registration' email.

2. Registrant "A" creates a registration request, but enters the wrong EIN.  Then, registrant "B" (for another company) creates a registration request, with the correct EIN, but that EIN happens to be the EIN entered by registrant "A".  The agency admin can now go in and fix the EIN for "A", and then request that "B" reenter the registration request (RR).

The other approach to these problems - if we cannot update a RR, is for the agency admin to simply delete any problematic RR and ask the registrant to reenter the registration request.  That is probably an OK fall-back position if we decide that this feature is too complex relative to its value to the user (I'm not sure either way at this point).

joaopapereira: Ok, for now lets just accept this issue. We will need to talk about this and ask for Chet's input

joaopapereira:
Looks good, going to merge PR

PR #166, opened 2016-01-31
Closes AgileVentures/MetPlus_tracker#181

patmbolger:


joaopapereira:
Don't we have unit tests for this feature?

patmbolger:
No unit or functional tests for this.  I thought about but reached the conclusion that there were no functional tests that I could add that could not be done better as acceptance tests.  (no model or controller changes occurred - just behavior of Devise with resulting flash messages changing).

joaopapereira:
I am ok with the answer with the answer, but in a usual context for modules changes we should have UT to check things out

joaopapereira:
Going to merge this PR

PR #167, opened 2016-02-02
Closes AgileVentures/MetPlus_tracker#184

patmbolger:


app/views/agency_admin/home.html.haml
```diff
@@ -52,6 +49,10 @@
...
-      %a(data-toggle='collapse' href='#companies_table'
-          aria-expanded='true' aria-controls='companies_table')
-        Show/Hide Companies
+      %a(id='toggle_companies' href='#companies_table' class='text-muted')
+        Hide Companies
+
+%script
+  $(VisibilityToggle.setup_branches);
+  $(VisibilityToggle.setup_people);
+  $(VisibilityToggle.setup_companies);
```
amba178: Good usage of jQuery library, can you move it to js file? It is alright few lines of script in view, but I think in future we will be using a lot of jQuery/javascript and probably we do not want in view file.

patmbolger: I had many problems trying to make that work when I developed the company registration 'deny' feature (uses event callback and AJAX).  Putting the JS load call in the js file never worked, and I had to move it to the view file.

We'll need another story in the future to work out any issues in the asset pipeline.  In the meantime, putting the load operation in the view that contains the element that will fire the callback works for now.

joaopapereira: That is strange, maybe the problem is that you need to wait for the page to be loaded in order to set events on things, did you do that?

joaopapereira: Something like 
$( document ).ready(function() {
//  the code
});

amba178: @joaopapereira:  That is preferred way to doing it. Waiting until the page is loaded, but still he will have problem, because of turbolinks.  Maybe we can do separate story for it. There is work-round without having any script code in the views file.

patmbolger: João - That is essentially what my code is doing (https://api.jquery.com/ready/):

    All three of the following syntaxes are equivalent:

    $( document ).ready( handler )
    $().ready( handler ) (this is not recommended)
    $( handler )

But I've tried it explicitly as well.  

In any case, when the code is placed in the JS file it does not work.  Salem recommends trying the jquery-turblinks gem.  That might work but I'd rather have that done as a separate story rather than add anything else to this story (that story will also have to refactor the 'deny' JS structure).

joaopapereira: Pff forgot about the turbolinks and minification of the JS.....

PR #146, opened 2016-01-06
Js registration implementation

vijayakarumudi:
updates for post and update tests for spec.

PR #168, opened 2016-02-02
Update company person #146

sidhene:
Closes AgileVentures/MetPlus_tracker#146

Added self-edit functionality to Company and CompanyPerson. Added navbar logic for same.


app/views/companies/_company.html.haml
```diff
@@ -23,7 +23,12 @@
     %strong Status:
   .col-sm-8#company_status
     = render :partial => 'companies/company_status', :locals => {company: company}
-.row.margin-top
+
```
patmbolger: Please add this line back in.  It introduces a space that we need on the agency admin home page.

app/views/companies/_company.html.haml
```diff
@@ -23,7 +23,12 @@
...
+%h4 People at this company
+
+- if company.company_people.empty?
+  There are no people assigned to this company.
+- else
   = render partial: 'company_people/company_people',    |
```
patmbolger: These added lines should not be here - that logic has been moved to the _company_people partial.

Lines 26 through end of file should look like this:

    .row.margin-top
      = render partial: 'company_people/company_people',    |
          object: @company.company_people,                  |
          locals: {company_status: @company.status}         |

config/routes.rb
```diff
@@ -48,7 +48,12 @@
   # ----------------------- Company People -----------------------------------
   # Company admin (and agency admin) can edit and delete a company person
   resources :company_people, path: '/company_admin/company_people',
-                       only: [:show, :edit, :update, :destroy]
+                       only: [:create, :show, :edit, :update, :destroy]
+
```
patmbolger: I don't think we need a 'create' route for company_people.  

We handle the *create* logic in the company_people invite process/controller, and, in fact, the company_people_controller.rb does not have a #create action.

spec/controllers/company_people_controller_spec.rb
```diff
@@ -1,5 +1,73 @@
...
+     context "valid attributes without password change" do
+       before(:each) do
+         @companyperson =  FactoryGirl.build(:company_person)
+         @user =  FactoryGirl.create(:user)
+         @companyperson.company_roles <<
+             FactoryGirl.create(:company_role, role: CompanyRole::ROLE[:CA])
+         @companyperson.save
+         @companyperson.valid?
+         patch :update_profile, company_person:FactoryGirl.attributes_for(:company_person, title: 'Line Manager').merge(FactoryGirl.attributes_for(:user, first_name:'John',last_name:'Smith',phone:'780-890-8976')),id:@companyperson
+         @companyperson.reload
```
patmbolger: This is a test to confirm that company_person attributes can be changed without having to enter *password* and *password_confirmation* each time (unless password is being changed).

However, the params sent to the controller includes both *password* and *password_confirmation*.  Please remove those to take the test work as expected.

To fix this, you could probably add these arguments:

    password: nil, password_confirmation: nil

to FactoryGirl.attributes_for(:user, .....)

patmbolger: Also, this line is quite long and should br broken up into multiple lines.  You could write it like this:

    patch :update_profile, company_person:  FactoryGirl.attributes_for(:company_person, title: 'Line Manager').   
      merge(FactoryGirl.attributes_for(:user, first_name: 'John', last_name: 'Smith', phone:'780-890-8976')),  
      id:@companyperson

The Ruby interpreter is smart about this, so the period that ends the first line tells the interpreter that the statement continues on the next line (with a method invocation), etc.

app/views/companies/_company.html.haml
sidhene: done

app/views/companies/_company.html.haml
sidhene: done

config/routes.rb
sidhene: I thought only the agency admin could do that. How does the company admin add a contact?

patmbolger: Good question.

The company admin would click the 'Invite Person" button in the company *show* view (app/views/companies/show.html.haml).  This will route to:

    new_user_invitation GET    /users/invitation/new(.:format)   people_invitations#new

There, the invitee's name and email are entered, and an invitation to join PETS is sent to the invitee.  Also, at the same time the invite is sent, the company admin is redirected to the company contact *edit* view:

    edit_company_person GET    /company_admin/company_people/:id/edit(.:format) company_people#edit

Here, the company admin can set other attributes for the invitee (currently just 'company role', but probably other attributes in the future).

The invitee clicks on a link in the invitation email, which brings that person to a PETS page where they enter (and confirm) their password, after which they end up on the company's 'landing page'.

The invite process is facilitated by the devise_invitable gem.  The people_invitations_controller.rb contains some customization to make it work for us.

NOTE: anything the company admin can do, the agency admin will be able to do as well.  We'll impose these restrictions when we implement cancancan authorizations.

joaopapereira:
Just updated the comment so it can be merged in the Waffle view

Gemfile
```diff
@@ -39,7 +39,7 @@ gem 'figaro'
 
 gem 'cancancan', '~> 1.10'
 
-
+gem 'nokogiri', '~> 1.6', '>= 1.6.7'
```
joaopapereira: Do we need this gem in the all the environments or only on tests?

app/controllers/companies_controller.rb
```diff
@@ -6,13 +6,28 @@ def show
   end
 
   def destroy
-    company = Company.find(params[:id])
-    company.destroy
+    @company = Company.find(params[:id])
```
joaopapereira: Is this used in any view? if not we do not need to make them instance variables

app/views/company_people/edit_profile_a.html.haml
```diff
@@ -0,0 +1,43 @@
+%div
```
joaopapereira: What is the difference between this view and the previous one?

app/views/layouts/_menu.html.haml
```diff
@@ -17,23 +17,25 @@
       %ul{:class => 'nav navbar-nav navbar-right'}
```
joaopapereira: If you did not do any change in this file, can you please reset it?

app/views/companies/_form.html.haml
```diff
@@ -22,48 +28,26 @@
...
+      = f.text_field :ein, size: 30, class: 'form-control'
 
   .form-group
     = label :company, :description, 'Description', class: 'control-label col-sm-2'
-    .col-sm-8
-      = text_area :company, :description, cols: 90, rows: 5, class: 'form-control'
+    .col-sm-10
+      =text_area :company, :description, cols: 90, rows: 5, class: 'form-control'
 
-    %h3 Company Contact
```
joaopapereira: Did we lose the contact person here because we are doing this somewhere else?

config/routes.rb
sidhene: Ok, that makes sense. Done.

joaopapereira:
Please review my comments

spec/controllers/company_people_controller_spec.rb
sidhene: Good trick! Done and done.

Gemfile
sidhene: I don't know. I compared my version to upstream/development and it shows up as a dependency in Gemfile.lock. I'm not sure how it got there, but I'm pretty sure I put nokogiri into Gemfile as a fix to errors. They might have come from that dependency. 

I'll try removing the two instances (gemfile and gemfile.lock dependency) and see if anything breaks. But I'm going to hold off on that until all the other change requests are passing Travis, ok?

app/controllers/companies_controller.rb
sidhene: done

app/views/company_people/edit_profile_a.html.haml
sidhene: That file is obsolete. I've removed it.

app/views/layouts/_menu.html.haml
sidhene: Sorry, I don't understand what you need.

Also, let's have a conversation about rebase. It's introducing a lot of errors (such as that obsolete file I already deleted) and I don't think it does anything good for us.

app/views/companies/_form.html.haml
sidhene: Yes, @patmbolger moved it to _company.html.haml

Gemfile
joaopapereira: I am ok with that

app/views/companies/_form.html.haml
joaopapereira: Ok but we do not need it in this form as well?

sidhene: No, because then it would appear twice and that would look weird :)

app/views/layouts/_menu.html.haml
sidhene: I did change that file to include logic that redirects a company person to edit_profile when they click on their name link in the navbar. 

Do you still want this change and (if so) can you elaborate a bit? I'm not confident I understand what you want.

joaopapereira: From what I can see in the diff the only change we have there is spaces removed, am I correct?

sidhene:
Okay I think I've covered everything except nokogiri and Joao's CR for app/views/layouts/_menu.html.haml which I'm not sure is needed or not.

I'll resubmit the PR, wait for it to pass Travis, and see if removing 2 nokogiri mentions breaks anything

sidhene:
The nokogiri CR didn't break anything on my machine. Waiting for this Travis to finish so I can check against Travis.

app/views/layouts/_menu.html.haml
sidhene: I don't think so. When I started, the logic accommodated a jobseeker and "everyone else" when I was done it also accommodated a companyperson.

sidhene:
Nokogiri done. Final Travis passed. Ready for merge.

joaopapereira:
Looks good going to merge

PR #176, opened 2016-02-12
connects AgileVentures/MetPlus_tracker#173

amba178:


amba178:
Some code are either hard coded or stubbed for placeholder.
I have created story on  job-company association as next task and 
we have sitemap to track the work on job related task, in document section.

Gemfile
```diff
@@ -48,9 +48,20 @@ gem 'cancancan', '~> 1.10'
 # gem 'capistrano-rails', group: :development
 
 gem 'active_record-acts_as'
+gem 'ffaker' 
+gem 'faker'
+gem 'will_paginate'
+gem 'bootstrap-will_paginate'
+gem 'jquery-turbolinks' #will solve reload problems with javascript 
+gem 'bootstrap-sass'
```
joaopapereira: Do we really need this gem? We are including the bootstrap libs by hand

Gemfile
```diff
@@ -48,9 +48,20 @@ gem 'cancancan', '~> 1.10'
 # gem 'capistrano-rails', group: :development
 
 gem 'active_record-acts_as'
+gem 'ffaker' 
+gem 'faker'
```
joaopapereira: What do we use this gems for?

app/assets/stylesheets/application.css
```diff
@@ -124,23 +126,6 @@ body {
 /*end of homepage css*/
 
 
-/*PLEASE PUT COMMENT
-  WHERE EXACTLY
-  CSS APLLIES
-  TO
-*/
-.field_with_errors {
```
joaopapereira: Removing these tags isn't going to break other pages?

app/assets/stylesheets/jobs.css.scss
```diff
@@ -0,0 +1,122 @@
+@import "bootstrap-sprockets";
```
joaopapereira: I think we should be consistent and decide if we are using plain css or less

app/controllers/jobs_controller.rb
```diff
@@ -0,0 +1,55 @@
+class JobsController < ApplicationController
+	before_action :find_job,	only: [:show, :edit, :update, :destroy]
+
+	def index
+		# @jobs = Job.paginate(:page => params[:page], :per_page => 32).includes(:company)
+		@jobs = Job.paginate(:page => params[:page], :per_page => 32)
```
joaopapereira: Any particular reason to use 32?

app/models/job.rb
```diff
@@ -10,12 +10,19 @@ class Job < ActiveRecord::Base
   has_many   :nice_to_have_skills, -> {where job_skills: {required: false}},
                 through: :job_skills, class_name: 'Skill', source: :skill
   has_many   :skill_levels, through: :job_skills
-  
+
+  CONTENT_TYPES = ['morning', 'day', 'evening']
```
joaopapereira: This name is not very helpful

app/views/layouts/_footer.html.haml
```diff
@@ -1,4 +1,17 @@
...
-  %div{:class => 'container'}
-    %center
-      = "® Met Plus #{Date.today.year}"
+/ %footer{:class => 'footer'}
+/   %div{:class => 'container'}
+/     %center
+/       = "® Met Plus #{Date.today.year}"
+/ %div{class: "container", id: "footer-container-id"} 
+/ .navbar.navbar-fixed-bottom
+%div{class: "panel panel-default", id: "panel-default-id"}
```
joaopapereira: Shouldn't this be footer class ?

app/views/layouts/application.html.haml
```diff
@@ -12,7 +12,6 @@
     = javascript_include_tag 'https://maxcdn.bootstrapcdn.com/bootstrap/3.3.6/js/bootstrap.min.js'
   %body
     = render partial: 'layouts/menu'
-    %div{:id => 'loading', :style => "background: url('#{asset_path 'spinner.gif'}') no-repeat center center; display:none;"}
```
joaopapereira: Why did you remove this div?

app/views/layouts/application.html.haml
```diff
@@ -35,6 +34,7 @@
               %li= msg
       %div{:class => 'row'}
         = yield
+      / = debug(params) if Rails.env.development?
```
joaopapereira: Please remove the commented line

config/routes.rb
```diff
@@ -63,12 +63,14 @@
 
   root 'main#index'
 
+
```
joaopapereira: Why the extra line?

db/migrate/20160203220908_add_shitf_to_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class AddShitfToJobs < ActiveRecord::Migration
+  def change
+    add_column :jobs, :shift, :string
```
joaopapereira: Is this change in the model?

db/migrate/20160203221159_add_fulltime_to_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class AddFulltimeToJobs < ActiveRecord::Migration
+  def change
+    add_column :jobs, :fulltime, :boolean, default: nil
```
joaopapereira: Is this change in the model?

app/assets/stylesheets/application.css
amba178: Would not break anything. Good catch though. I moved to job.css.scss it is starts at line 117 and  it is nicely formatted with sass. I think, eventually we should  refactor out to separate css.scss file.

features/jobs.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Manage Jobs  
+
+As a user #not quite correct  
+I want to post, edit, delete jobs 
+
+Scenario: adding job to database
```
joaopapereira: Shouldn't this be a background?

app/views/jobs/_form.html.haml
```diff
@@ -0,0 +1,24 @@
+.col-sm-offset-3.col-sm-6
+	= render 'shared/error_messages', object: f.object 
+.clearfix 
+.form-group 
+	=f.label :title, class: "col-sm-3 control-label text-right"
+	%div{class: "col-sm-6"}
+		=f.text_field :title, class: "form-control" 
+.form-group 
+	=f.label :jobId,"Job id", class: "col-sm-3 control-label text-right"
+	%div{class: "col-sm-6"}
```
joaopapereira: What is this job is for?
Should it be editable?

features/step_definitions/fixtures_steps.rb
```diff
@@ -1,4 +1,4 @@
-Given(/^the following (.+) records:$/) do |factory, table|
+ Given(/^the following (.+) records:$/) do |factory, table|
```
joaopapereira: Why did you add a space there? And the new lines in the end of the file?

Gemfile
amba178: it help to generate  fake data like address(city name, street name, zipcode, state name), name, company name etc. We can populate the database with a lot of data using these gem.

features/step_definitions/jobs_steps.rb
```diff
@@ -0,0 +1,6 @@
+Given(/^the following jobs exists:$/) do |table|
+  table.hashes.each do |hash|
+    @job = Job.create!(hash)
+  end
```
joaopapereira: Is this step not accomplished by the factory generic method?

features/step_definitions/navigation_steps.rb
```diff
@@ -1,11 +1,15 @@
 When(/^I go to the (.+) page$/) do |page|
   case page
+    
+    when 'jobs/new'
```
joaopapereira: I personally prefer this more descriptive like 'Job creation'

features/step_definitions/users_steps.rb
```diff
@@ -17,6 +17,15 @@
    visit  "/users/edit.#{user.id}"
 end
 
+
+Then(/^I should verify the change of title "(.*?)", shift "(.*?)" and jobId "(.*?)"$/) do |title, shift, jobId|
```
joaopapereira: Why is this in user. Steps?

spec/controllers/jobs_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    it { should route(:get, "/jobs/1")
+              .to(action: :show, id: 1) }
+  end
+
+  describe "Route => edit" do
+    it{ should route(:get, "/jobs/1")
+            .to(action: :show, id: 1) }
+  end
+
+  describe 'GET #index' do
```
joaopapereira: We are missing the sad path in all endpoints

spec/models/job_spec.rb
```diff
@@ -27,9 +27,10 @@
     it { is_expected.to have_db_column :id }
     it { is_expected.to have_db_column :title }
     it { is_expected.to have_db_column :description }
-    it { is_expected.to have_db_column :company_id }
-    it { is_expected.to have_db_column :company_person_id }
-    it { is_expected.to have_db_column :job_category_id }
+  
+    xit { is_expected.to have_db_column :company_id }
+    xit { is_expected.to have_db_column :company_person_id }
+    xit { is_expected.to have_db_column :job_category_id }
```
joaopapereira: You added field but no validations

Gemfile
joaopapereira: Do we need the gem in production or in test?

app/assets/stylesheets/application.css
joaopapereira: If we are using these classes somewhere else they should be in application.css and not inside a specific css file

joaopapereira:
Please review my comments

Gemfile
amba178: Yes we do. otherwise we do not get nicely formatted and styled pagination. Tried without it and did not work for me.

app/assets/stylesheets/jobs.css.scss
amba178: Sure. Most of the files have sass, scss extention we should use sass if we can. Plus we can still use all css stuff with it. The advantage is that it let us to use less code, use nesting, and let us to set variable.

app/assets/stylesheets/application.css
amba178: Sure. since it has sass stuff that is why I did not included in the application.css.  Maybe we should have custom.css.css file,  we can put anything applies on application level together.

Gemfile
amba178: Yes. production and development. I am not quite sure how it works with test env though.

app/controllers/jobs_controller.rb
amba178: random number.I think the default # is 30. I could have used 20 record per page.

app/models/job.rb
amba178: any suggestion. That is what we came up with myself and @patmbolger.

app/views/jobs/_form.html.haml
amba178: You lost me. Sorry what line or lines are you referring to.

app/views/layouts/_footer.html.haml
amba178: I used footer class in the line 8 even though inside the div tag. I thought would be easy to style it without difficult.

app/views/layouts/application.html.haml
amba178: all css styling stuff for loading are gone. I thought there is no point of having it there. i can put back if you think we going to use it in the future time.

app/views/layouts/application.html.haml
amba178: Sure. I forgot to remove it.

config/routes.rb
amba178: I will fix it.

db/migrate/20160203220908_add_shitf_to_jobs.rb
amba178: Yes. we decided to add: I believe three columns to the tables.
-shift
-jobId
-fulltime

db/migrate/20160203221159_add_fulltime_to_jobs.rb
amba178: yes.

features/jobs.feature
amba178: right. fix it.

features/step_definitions/fixtures_steps.rb
amba178: No idea. maybe it is my text editor. I will try to align it again.

features/step_definitions/jobs_steps.rb
amba178: I do not think so, but I will look at it, again. thanks for pointing out.

features/step_definitions/navigation_steps.rb
amba178: Sure make sense. just want to be consistent with the form.

features/step_definitions/users_steps.rb
amba178: right, I will create separate step file for job.

spec/controllers/jobs_controller_spec.rb
amba178: sure. That will be a good idea.

spec/models/job_spec.rb
amba178: You are right. I guess overlooked it.  Thanks.

Gemfile
joaopapereira: Didnt work because you started using sass in the CSS right?

Gemfile
joaopapereira: I am not a fan of having gems in production that we are not going to use, is there any option to have the gem only if we need to seed?

app/assets/stylesheets/application.css
joaopapereira: This is a team decision we need to decide if we are going to use plain CSS or use SASS

app/models/job.rb
joaopapereira: SHIFT_TIMES
SHIFT_OPTIONS

app/views/layouts/_footer.html.haml
joaopapereira: Does this part brak the change @patmbolger did to solve the problem of the distance to the bottom?

app/views/layouts/application.html.haml
joaopapereira: if we no longer have Ajax and we do not need a loading spinner I am ok with that

app/views/jobs/_form.html.haml
joaopapereira: I do not understand what is the JOB ID for.
And I believe if is not an ID I believe we should not allow it to be edited

features/step_definitions/navigation_steps.rb
joaopapereira: I understand but if you see the rest of the names, it is not consistent with them

Gemfile
patmbolger: Why do we need "fake" data in production?

Also, we probably want to split our seeds.rb file into at least two files: one to contain baseline data for the system to startup with (initial agency, agency_admin, etc.) and one to contain test/dev data (where 'faker' would be useful).

app/assets/stylesheets/application.css
patmbolger: I think these css statements (#error-explanation, .field_with_errors) should stay in application.css as they are generic for most forms and are not jobs-specific.  This will be very confusing otherwise.

app/assets/stylesheets/jobs.css.scss
patmbolger: With some of these changes you are proposing restructuring of how we do css, and introducing another overlay language.  This is fine but really requires a short document outlining the proposed changes, followed by a team discussion.  Deciding on these kinds of changes with a little back-and-forth in these comments in not sufficient, IMO.

app/assets/stylesheets/jobs.css.scss
```diff
@@ -0,0 +1,122 @@
...
+@import "bootstrap";
+
```
patmbolger: I am running your feature branch now and I am not sure this import directive is actually working, since we are also accessing the bootstrap files via CDN.

Also, if we go with the bootstrap-sass gem, do we have to give up CDN import and just serve our local copy of bootstrap style files?

app/assets/stylesheets/jobs.css.scss
```diff
@@ -0,0 +1,122 @@
...
+		color: red;
+    	margin: 0 0 30px 0;
+  	}
+}
+.field_with_errors {
+  @extend .has-error;
+  .form-control {
+    color: $state-danger-text;
+  }
+}
```
patmbolger: Last two statements should go back to application.css.

app/models/job.rb
patmbolger: JOB_SHIFTS  ??

db/migrate/20160203221325_add_job_id_to_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class AddJobIdToJobs < ActiveRecord::Migration
+  def change
+    add_column :jobs, :jobId, :string
+  end
```
patmbolger: Rails naming conventions call for variables to be snake-case, not camel-case.  So, suggest changing name to 'job_id'.

db/seeds.rb
```diff
@@ -122,23 +83,128 @@
             description: 'Proficient in all aspects, requires little supervision')
 SkillLevel.create(name: 'Expert',
             description: 'Proficient in all aspects, able to work indepently')
-
-# Create all agency roles - this should stay in production version of this file
+#AgencyRole
```
patmbolger: Why delete the comment?  This is important information to retain.

db/seeds.rb
```diff
@@ -122,23 +83,128 @@
...
 AgencyRole::ROLE.each_value do |agency_role|
   AgencyRole.create(role: agency_role)
 end
-
-# Create all company roles - - this should stay in production version of this file
+#CompanyRole
```
patmbolger: Same comment as above.

features/jobs.feature
```diff
@@ -0,0 +1,42 @@
...
+	Given I am on the jobs/new page
+	And  I fill in the fields: 
+		| Title                  | cashier|  
+		| Job id                 |  |  
+		| Description            |  |  
+	And  I select "day" in select list "Shift"
+	And  I check  "job_fulltime" 
+	And  I press "new-job-submit"
+	Then  I should see "The form contains 2 errors"
+
```
patmbolger: Do you intend to add a test for job delete?

features/step_definitions/web_steps.rb
```diff
@@ -84,6 +84,10 @@ def search_text text
   check(item)
 end
 
+Then(/^I check  "(.*?)"$/) do |item|
+  check(item)
+end
+
```
patmbolger: This duplicates function of the step immediately above it (line 83).

spec/controllers/jobs_controller_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+    xit { should set_flash }
+  end
+
+  describe 'PATCH #update' do
+    let(:job){FactoryGirl.create(:job)}
+    before { get :update, :id => job.id }
+    xit { should set_flash }
+  end
+  
+end
```
patmbolger: There are a lot more tests that could be added here.  Check out job_seekers_controller_spec.rb, and similar, for some ideas.

Gemfile
amba178: I would not have said production, Sorry for poor usage of my word, production. I would say for staging or in development env.  By having more data we can predict how our apps behavior in real world. I do not think  would be good idea to split into two files. You could conditionally create data.

Gemfile
amba178: will_paginate goes with bootsrap-will_paginate.
for boostrap-sass, I have used variable like $light-blue, and for field errors  used some saas there without it, would not work. This is how I look at it, would not hurt us to have it here, and other team do not need to know saas, they can use plain css inside sass file.

app/assets/stylesheets/application.css
amba178: As I said earlier, we do not need to know sass. We can use css inside sass. Even though, I added a gem for it, by default rails will create css.scss extention unless you tell rails not to create when you when generate, for sass I believe. We could decouple it, to separate file, sure.

app/assets/stylesheets/jobs.css.scss
amba178: @patmbolger: You do need to know sass. Just Use css inside the file that is, we will be fine. Basically do not use it.

joaopapereira: @amba178 we need to discuss this as a team. We go with SASS or we go with CSS I do not think it is a good option to have them both o the site, it will just confuse things around.

app/assets/stylesheets/jobs.css.scss
amba178: should working I can demonstrate, probably you doing it wrong.  I would first try to understanding the gem, and its dependencies that will give you the overview.

app/models/job.rb
amba178: Oh the CONSTANT NAME, yes. right. Good suggestion. thanks

app/views/jobs/_form.html.haml
amba178: Oh. This is not the default job id where rails creates for us. it is column that we decided to add as jobId. Usually, as you know rails convention labels :id, job_id so I tried to give it different name, as jobId.

app/views/layouts/_footer.html.haml
amba178: it does not. But the footer is relative to the page content. Instead of fixing it at the bottom of the page whether the page has content or not, But it does not overlap on the page content.  it just behaves like navbar...

db/migrate/20160203221325_add_job_id_to_jobs.rb
amba178: This might have name collision with the rails foreign job id. leave  as it is or change to something else. any suggestion?

db/seeds.rb
amba178: I thought, I moved to the top of the page since as you said we need it in production env. I will double check again. thanks.

app/views/layouts/_footer.html.haml
patmbolger: I have run the PR code and don't see any problem with the footer - it behaves correctly.

features/jobs.feature
amba178: Bad wording should be updating instead of creating. thanks.

features/step_definitions/navigation_steps.rb
amba178: ok. thanks

features/step_definitions/web_steps.rb
amba178: did not work for me, but i will double check the regex syntax

db/migrate/20160203221325_add_job_id_to_jobs.rb
patmbolger: Yes, good point - it could be confusing to have a 'job_id' in the jobs table and also 'job_id' in the job_skills table.

Perhaps we should call it 'company_job_id', since that is what it is - the identifier for the job used by the company's internal system for tracking job openings.

db/seeds.rb
patmbolger: Actually, I just meant for those comments to apply to those specific roles creations (not the whole file).  In production, we need to populate the roles tables with these fixed values.  The user cannot add more roles since we actually use these to switch code paths in out logic, so our code needs to be aware of the codes and the codes cannot be deleted or missing.

spec/controllers/jobs_controller_spec.rb
amba178: Remember jobs is still in progress and these are simple crud actions. it will  be added as more associations implention is done.. If you have specific one let me know.

amba178:
Thanks guys for feedback.

db/migrate/20160203221325_add_job_id_to_jobs.rb
amba178: That would be a good option. thanks

features/step_definitions/web_steps.rb
patmbolger: There are 3 scenario step statements that have an extra space that is preventing you from being able to use the more generic version of checking a box.  For instance, in line 17:

    Then I check  "Full-time"

There are two spaces between 'check' and "Full-time".  If you change this to one space the step at line 83 in web_steps.rb will work.

Similarly for lines 27 and 39 in the feature file.

features/jobs.feature
patmbolger: I suggest adding a test for deleting a job.  You'll have to use the @javascript switch for that test since it involves a modal.

spec/controllers/jobs_controller_spec.rb
patmbolger: A lot of the other controller tests try to isolate atomic operations of controllers, ala the idea of 'test on thing at a time'.  So, for instance, its good to test for "did it render the correct template", and "did it return the expected status" for all actions.

For the "create" and "update" action, it is useful to test the controller when it is presented with *valid* form data, as well as *invalid* form data.  So, for each of those we set up a params hash, invoke the action, and then test specific action responses.  

This is a good example from job_seekers_controller_spec.rb:

```
describe "POST #create" do
      
   context "valid attributes" do
     before(:each) do
       @jobseeker = FactoryGirl.create(:job_seeker)
       @user = FactoryGirl.create(:user)
       @jobseekerstatus = FactoryGirl.create(:job_seeker_status)
       jobseeker_hash = FactoryGirl.attributes_for(:job_seeker).merge(FactoryGirl.attributes_for(:user)).merge(FactoryGirl.attributes_for(:job_seeker_status))
       post :create, job_seeker: jobseeker_hash 
     end
        
     it 'sets flash message' do
        expect(flash[:notice]).to eq "A message with a confirmation and link has been sent to your email address. Please follow the link to activate your account."
     end
                
     it 'returns redirect status' do
        expect(response).to have_http_status(:redirect)
     end
     it 'redirects to mainpage' do
       expect(response).to redirect_to(root_path)
     end
          
   end
   context 'invalid attributes' do
     before(:each) do
       @jobseeker = FactoryGirl.create(:job_seeker)
       @user = FactoryGirl.create(:user)
       @jobseekerstatus = FactoryGirl.create(:job_seeker_status)
       @jobseeker.assign_attributes(year_of_birth: '198')
       @user.assign_attributes(first_name:'John',last_name:'Smith',phone:'890-789-9087')
       @jobseekerstatus.assign_attributes(description:'MyText')
       @jobseeker.valid?
       jobseeker1_hash = FactoryGirl.attributes_for(:job_seeker, year_of_birth: '198').merge(FactoryGirl.attributes_for(:user,first_name:'John',last_name:'Smith', phone:'890-789-9087')).merge(FactoryGirl.attributes_for(:job_seeker_status, value=nil, description:'MyText'))
       post :create, job_seeker: jobseeker1_hash 
        
     end
     it 'assigns @model_errors for error display in layout' do
       expect(assigns(:model_errors).full_messages).to eq @jobseeker.errors.full_messages
     end
     it 'renders new template' do
        expect(response).to render_template('new')
     end
     it "returns http success" do
        expect(response).to have_http_status(:success)
     end
    end
 end
```
(note that when a flash message is set up by the controller, we could match that against the actual expected message.  Also, model errors can be compared against expected as well).

When you have added associated models to the Job creation, your delete action test should ensure that the appropriate associated models are destroyed if the job is destroyed.  For instance, the *job_skills* record(s) should also be destroyed if the *job* is destroyed.  There is an example of this testing in company_registrations_controller_spec.rb:

```
describe 'DELETE registration also deletes associated objects' do
    let(:company_person) do
      $cp = FactoryGirl.create(:company_person)
      $cp.company_roles << FactoryGirl.create(:company_role,
                                role: CompanyRole::ROLE[:CA])
      $cp.save
      $cp
    end
    let(:address1) { FactoryGirl.create(:address) }
    let(:address2) { FactoryGirl.create(:address, city: 'Detroit') }
    let!(:company) do
      $comp = FactoryGirl.build(:company)
      $comp.company_people << company_person
      $comp.addresses << address1 << address2
      $comp.save
      $comp
    end
    it ' delete company person(s)' do
      expect { delete :destroy, id: company }.
          to change(CompanyPerson, :count).by(-1)
    end
    it ' delete company address(s)' do
      expect { delete :destroy, id: company }.
          to change(Address, :count).by(-2)
    end
  end
```
Of course, this last testing is beyond the scope of this initial PR.

features/step_definitions/web_steps.rb
amba178: Right. Thanks

spec/controllers/jobs_controller_spec.rb
amba178: Most of test will make sense once the associations are implemented.   Good suggestions. thanx

amba178:
I have update the PR....I took out the boostrap sass gem, and replaced the scss code with css version.  I hope will make sense to @team.

PR #177, opened 2016-02-15
connects AgileVentures/MetPlus_tracker#173

amba178:


Gemfile
```diff
@@ -48,9 +48,19 @@ gem 'cancancan', '~> 1.10'
 # gem 'capistrano-rails', group: :development
 
 gem 'active_record-acts_as'
+gem 'ffaker'
```
joaopapereira: Do we really have no way to use those gem in heroku and only have them installed for dev and test?

app/assets/stylesheets/application.css
```diff
@@ -124,23 +124,6 @@ body {
 /*end of homepage css*/
 
 
-/*PLEASE PUT COMMENT
-  WHERE EXACTLY
-  CSS APLLIES
-  TO
-*/
-.field_with_errors {
```
joaopapereira: Are we sure we are not using this css in any other place?

app/controllers/jobs_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+
+	def destroy
+		@job.destroy
+		flash[:alert] = "#{@job.title} has been deleted successfully."
+		redirect_to jobs_url 
+	end
+
+	private 
+		def find_job
+			@job = Job.find_by_id(params[:id])
```
joaopapereira: Shouldn't we have errors handling if a job is not found?

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
```diff
@@ -0,0 +1,5 @@
+class AddCompanyJobIdToJobs < ActiveRecord::Migration
+  def change
+    add_column :jobs, :company_job_id, :string
```
joaopapereira: Is this value going to be unique? Is it going to be editable? I do have some reservations about this field because it looks like we are keeping 2 ids for each job.
Is this just informative?

features/jobs.feature
```diff
@@ -0,0 +1,42 @@
...
+As a user #not quite correct, 
+I want to post, edit, delete jobs 
+
+background: adding job to database 
+	Given the following jobs records:
+	| title               | jobId   | shift  | fulltime | description | 
+	| software developer  | KRK12K  | evening| true     | internship position with pay| 
+Scenario: Creating and Updating Job successfully	
+	Given I am on the job creation page 
+    And  I fill in the fields:
```
joaopapereira: Indentation on the file looks a bit odd

features/step_definitions/fixtures_steps.rb
```diff
@@ -50,7 +50,6 @@
     agency_person.save
   end
 end
-
```
joaopapereira: Are the changes in this file needed?

features/step_definitions/users_steps.rb
```diff
@@ -17,6 +17,7 @@
    visit  "/users/edit.#{user.id}"
 end
 
+
```
joaopapereira: Is the change in this file needed?

spec/controllers/jobs_controller_spec.rb
```diff
@@ -0,0 +1,78 @@
+require 'rails_helper'
+
+RSpec.describe JobsController, type: :controller do
+  
+  describe "Route => index" do
+    it { should route(:get, "/jobs")
```
joaopapereira: I believe we are missing some tests for the case were save fails

joaopapereira:
Generally the PR looks good. Please review the comments and I believe we should have a team talk bout css

Gemfile
amba178: Of course we are not using it for production, it will be very helpful for staging purpose to my opinion. I think there is away to let the data to be populated on staging env if we do not want it on the production.

app/assets/stylesheets/application.css
amba178: There is edited version of .field_with_errors at the bottom of  the same file,  I believe.

app/controllers/jobs_controller.rb
amba178: This is the same as doing Job.find(params[:id) inside the actions and error handling take place inside the actions. good point though.

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
amba178: This is external id for job, provided by employer. Employers usually have their own job id to a identify specific job title with that is company_job_id for.

spec/controllers/jobs_controller_spec.rb
amba178: Since the job association with the other models are not implemented yet and writing posting, editing through form does not make sense now. I will make sure to add  tests as we implement the association.You right need to have that, but whether I provided valid or not valid data will still fail.

amba178:
Note: The test file will be updated as we implement the job association  with other models.

db/seeds.rb
```diff
@@ -122,23 +83,130 @@
...
+companypeople = CompanyPerson.all.to_a 
+companies = Company.all.to_a 
+#job
+200.times do |n|
+  title = FFaker::Job.title 
+  description = Faker::Lorem.paragraph(3,false, 4 )
+  shift = ["day", "evening", "morning"][r.rand(3)]
+  fulltime =  [false, true][r.rand(2)]
+  jobId = ((1..9).to_a + ('A'..'Z').to_a).shuffle[0..7].join 
+
```
patmbolger: *jobId* is not *company_job_id*

db/seeds.rb
```diff
@@ -122,23 +83,130 @@
...
+  Job.create!(title: title,
+             description: description,
+             shift: shift,
+             jobId: jobId,
```
patmbolger: *jobId* is not *company_job_id*

features/jobs.feature
```diff
@@ -0,0 +1,42 @@
+Feature: Manage Jobs  
+
+As a user #not quite correct, 
+I want to post, edit, delete jobs 
+
+background: adding job to database 
+	Given the following jobs records:
+	| title               | jobId   | shift  | fulltime | description |
```
patmbolger: recommend changing column name to *company_job_id* to make it clearer what data is being populated.

features/jobs.feature
```diff
@@ -0,0 +1,42 @@
...
+	When I click the "jobs-edit-link" link 
+	And  I fill in the fields: 
+		| Title                  | cab-driver|  
+		| Job id                 | KRT123 |  
+		| Description            | Atleast two years work experience|  
+	And  I select "day" in select list "Shift"
+	Then I check "Fulltime" 
+	And   I press "Update"
+	Then  I should see "cab-driver has been updated successfully."
+	And   I should verify the change of title "cab-driver", shift "day" and jobId "KRT123"
```
patmbolger: *jobId* is not *company_job_id*

db/seeds.rb
```diff
@@ -226,6 +294,3 @@
               year_of_birth: '1970', resume: 'text',
           job_seeker_status: @jss3, confirmed_at: Time.now)
 
-
-
-# Think we have enough users(JS, AA, JD, CM, CP). We can login in productin env with created creditials.
```
patmbolger: Running rake db:seed fails due to misnamed 'jobId' field. See comments above.

app/views/jobs/_form.html.haml
```diff
@@ -0,0 +1,24 @@
...
+	%div{class: "col-sm-6"}
+		=f.text_field :title, class: "form-control" 
+.form-group 
+	=f.label :company_job_id,"Job id", class: "col-sm-3 control-label text-right"
+	%div{class: "col-sm-6"}
+		=f.text_field :company_job_id, class: "form-control" 
+		
+.form-group 
+	=f.label :shift, class: "col-sm-3 control-label text-right"
+	-shift_opts = ['morning', 'day', 'evening']
```
patmbolger: This line should be
```
-shift_opts = Job::SHIFT_OPTIONS
```

db/seeds.rb
amba178: yes. we should change it to company_job_id since we already updated on of the jobs columns to company_job_id.....

features/jobs.feature
amba178: Should be fine as a label...on the form.

features/jobs.feature
amba178: yes. as column name.

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
patmbolger: We can't rely on *company_job_id* being unique, since it is another company's ID, not ours, so there could be collisions in numbering schemes between different companies.

However, it does not matter since we are not relying on the field to be unique (in fact, PETS does not do anything with the field other than store and display it).

app/views/jobs/_form.html.haml
amba178: you could, that is one way doing it.

patmbolger: We should not have 'shift options' defined in two places.  Please change this.

app/assets/stylesheets/application.css
joaopapereira: The changes are different, because in that line you do not change the field_with_error, but the form-control inside it. That is why I am asking

app/controllers/jobs_controller.rb
joaopapereira: What I am asking is, what happens if you access job/12312312312312

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
joaopapereira: Is it going to be editable?

app/assets/stylesheets/application.css
amba178: well, when error occurs the red border around the field form is applied that is what says basically. It does not break anything.  I can do demo if you want.

app/controllers/jobs_controller.rb
amba178: in short  if there no such record in database,  will return nil.  First of all who has access to edit url?  I think this can be taken care of when when we implement cancancan or devise also has some authentication too.

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
amba178: if company has job id and may use to identify certain job with. That can be unique to them to differentiate one job to another. On our side, the database dose not care the uniqueness of the external job id. Yes can be  editable through the form

app/views/jobs/_form.html.haml
amba178: How can i word this? ..... First of all I do not like to put ruby code in the view unless it is very short or necessary. Now question for you, in term of performance which  lines of code do you think does better?

app/controllers/jobs_controller.rb
joaopapereira: everyone have access to the job/:jobId URL, this is not something we need to take care with cancancan.
If we cannot find the job we have 2 options:
1. Send the user back to the list of jobs and tell them the job do not exists
2. Show an empty page with the error message

But we cannot should an empty form or empty page

app/assets/stylesheets/application.css
joaopapereira: Is this the behavior that we had before? Does it work in all the forms?

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
joaopapereira: I do not agree that a Job ID Can be editable. If they have a mistake they should create a new job.

app/assets/stylesheets/application.css
amba178: Yes.the idea was the same, but displaying errors look bad,  the edited version works  properly to all forms.

app/controllers/jobs_controller.rb
amba178: If you think is not good idea to do it through cancancan or devise, than options 2 would be ok. We can display the error message 404 not found than redirect whatever page want. I am ok with that. I can change Job.find_by_id to Job.find as you know that will give us 404 error if the record not found in the database.

db/migrate/20160203221325_add_company_job_id_to_jobs.rb
patmbolger: Regarding the ability to edit a job once it is saved - I was thinking that this what we want, because only authorized people will be able to edit/update a job (typically, Job Developers and Company Contacts ..... maybe the agency admin(s) as well).  Otherwise, someone may lose a lot of work and have to reenter that work if they want to, for instance, adjust the job description to be more accurate.

amba178: Interesting point, @joa. In that case, can a job have expiration date?  if yes, we can look at two scenarios:
 signal the jobs tables:
-the position is no longer available(employer hired a jobseeker for that position)
or the time period ended(maybe we set the expiration date) 
Any suggestions? 
-

amba178: if i am not wrong, @joaopapereira specifically talking about just the external job ID column. Of course we can edit, update the job. I do not know why he wants to keep the external job id uneditable.

PR #179, opened 2016-02-17
WIP: Event and Notifications management (waffle #186) (PR not to be merged)

patmbolger:
See architecture description here: https://docs.google.com/document/d/11Y1hzcpz4OUc30C6-pAD9EAbaMFKyw9cmUrdvw-QSrg/edit#

### DIFF
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -12,20 +12,14 @@ var PusherControl = {
       alert('From: ' + data.name + ', Message: ' + data.message);
     });
 
-    // pusher_control channel is used to toggle bindings appropriate
-    // to the user (e.g., when a user logs in as an agency_person,
-    // that client is subscribed to channel(s) appropriate for that role)
+    // pusher_control channel is used to send server-side events to client
     var channel = pusher.subscribe('pusher_control');
 
-    channel.bind('user_logged_in', function(data) {
-      alert('User ID: ' + data.person_id + ' has logged in.');
-      $(window).unload( function(data) {
-        alert('Now the page should be fully loaded ...');
-      })
-    });
-
     channel.bind('js_registered', function(data) {
-      alert('Job Seeker: ' + data.name + ' has joined PETS.');
+      // Process event if I am logged in and am an agency_person
+      if (Cookies('person_id') && (Cookies('person_type') === 'AgencyPerson') ) {
+        alert('Job Seeker: ' + data.name + ' has joined PETS.');
+      }
     });
   }
 };
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -8,7 +8,7 @@ var PusherTest = {
     });
     var channel = pusher.subscribe('pusher_test');
     channel.bind('js_register', function(data) {
-      alert('Job Seeker ' + data.name + ' has joined PETS');
+      alert('Job Seeker: ' + data.name + ' has joined PETS');
     });
   }
 };
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -1,13 +1,12 @@
 var PusherTest = {
 
   setup: function () {
-    var pusher = new Pusher('1d3d6764932fac0ac318');
+    var pusher = new Pusher('<%=ENV['PUSHER_APP_KEY']%>');
     var channel = pusher.subscribe('my_channel');
     channel.bind('new_message', function(data) {
-      alert('An event was triggered with message: ' + data.message);
+      alert('From: ' + data.name + ', Message: ' + data.message);
     });
   }
 };
 
 $(PusherTest.setup);
-$(PusherTest) {alert('this is a test')};
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -6,10 +6,14 @@ var PusherControl = {
   setup: function () {
     var pusher = PusherControl.get_pusher();
 
-    // the following is used only for test purposes
-    var channel = pusher.subscribe('my_channel');
+    var channel = pusher.subscribe('test_channel');
     channel.bind('new_message', function(data) {
-      alert('From: ' + data.name + ', Message: ' + data.message);
+      noty({text: "TEST: <a href='/'>Go Home</a>",
+          layout: 'bottomLeft',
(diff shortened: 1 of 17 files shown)


app/assets/javascripts/pusher.js.erb
```diff
@@ -0,0 +1,43 @@
+var PusherControl = {
+  get_pusher: (function () {
+    var pusher = new Pusher('<%=ENV['PUSHER_APP_KEY']%>');
```
joaopapereira: Is there a way to do not expose the API key?

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+        cookies[:person_type] = current_user.actable_type
+      end
+    end
+  end
+
+  # DELETE /resource/sign_out
+  def destroy
+    super
+    # Remove cookies data
+    cookies.delete :person_id
```
joaopapereira: If the person as remember me should we destroy the cookie?

joaopapereira:
Looks good, have some ideas:
do you think we should have the notifications like Facebook near the name with number?
Maybe the notification format should be a partial or something instead of being directly in JS?
Do we have a way to save the notifications in the database or something, to show the users if they are not online or is that supported by pusher?
Eventually we might need to create a filter so you do not receive all the notifications from the system, even if you are allowed to.

app/controllers/users/sessions_controller.rb
patmbolger: Good question.  For our current config, the answer is yes.  But this is due to this setting for Devise:

    config.expire_all_remember_me_on_sign_out = true

So, when the user signs out all remember tokens are invalidated.

However, I think my code will break if we change that setting to false.  Will take a look ....

app/assets/javascripts/pusher.js.erb
patmbolger: Pusher does not seem to support sending the key in encrypted format.  

I suppose one could implement an AJAX request from client to server to have the encrypted key sent to the client, which then decrypts, using a shared secret, before connecting to Pusher.  I don't think we currently have security needs that rise to that level, however.

FYI - Pusher does support user authentication, which is their answer to the need for greater security.  User auth is required for both *private* channels as well as *presence* channels. (the latter allows the app to see who is connected, etc. - you could actually use this to implement advanced features - such intra-app chat - if desired)

patmbolger:
João - we could certainly experiment with how to best show multiple alerts to the user, ala Facebook or other concepts.

Not sure how to implement a partial for the message construction.  I can look at that.  On the other hand, the message is only constructed and used in one place, so I don't think we'd have the benefit of sharing (such as a partial provides).

Pusher does not support event/notifications persistence - we'll have to do that ourselves, but it is certainly doable.  This spike effort intentionally did not include:

1. Event retention for delivery later (not as a Pusher message but in some "events" queue for each user, so the user could see any messages they may have missed because not logged in at the time)
2. Email delivery of the same message (this would ensure message receipt even if the user is not logged in at time of the event) (with this maybe we don't need item #1)
3. Creation of action tasks based on event - for example, a task is added to an agency admin's task list to associate a job seeker with a job developer and a case manager - when the 'js_registered' event happens.

And, yes, we can filter notifications by using *private* channels if needed.

app/assets/javascripts/pusher.js.erb
joaopapereira: ok

joaopapereira:
There are things that we need to discuss, specially how should we handle the notification at server side, because we will also receive them there and how to understand if the user already saw it or not.
In terms of front end I believe we got a good prototype, now we need to start thinking how can we architect this in the server side

app/controllers/users/sessions_controller.rb
patmbolger: I have found that the behavior of Devise upon logging out is that same regardless of the setting of that config variable - that is, once the user is logged out, they *have* to sign in again.

So, I think the current logic will work fine regardless.

patmbolger:
Generally, Pusher is not architected to send events to the app server.  You can see the idea in the first diagram here: https://pusher.com/docs.

Clients can trigger events, but those would go to all of the other connected clients, not the server. (https://pusher.com/docs/client_api_guide/client_events#trigger-events).  This would require the use of *private* or *presence* channels - which would add complexity and thus should be avoided unless we have to.

If a client needs to raise an event, we could accomplish this via a post to a server controller, which would then take action (such as sending an event to all other clients).

The architecture is pretty robust to handling any scenario we might come up.

I think we should consider going ahead with putting this prototype into the production codebase, since it actually serves a useful business purpose.  Then, grow and evolve as we tackle additional stories.

joaopapereira:
Looking at my comment I realised that we are the ones generating the events in the backend.
In order to have control over the displayed events we might need to do Ajax calls to ensure wich users received which events and stuff. Like this we might not need to store in the cookie the type of user we have

patmbolger:
Yes, I had thought about AJAX calls instead of using a cookie, on order for the client to determine: am I logged in?, what kind of "person" am I?, etc.  However, cookies seemed to be a much cleaner way to do this, at least for this prototype functionality.

Another way to answer those questions is to perform a jQuery select for the presence of an ID that indicates the user is logged in (an ID associated with the user name and link in the top right corner of the page).  That seems a little inelegant but could work, but I'd rather use one of the other methods.

If we want to know which clients received an event, AJAX from the client would work.  Also, the Pusher REST API allows for querying which clients are currently on *presence* channels (so, assuming a client will receive an event if on the channel, this list of users would be the complete set who received an event sent via that channel).

patmbolger:
BTW, it is interesting to see that Rails 5 will include WebSockets pub/sub capability: https://github.com/rails/rails/tree/master/actioncable

joaopapereira:
I am a little bit divided here because qe should know in the backend who got the notification. not saying we need that for now in the proof of concept but maybe in the first release

patmbolger:
For our initial notification feature I don't see the need to have the backend keep track of who received a real-time alert (and, of course, who did not) - at least relative to the added complexity that would require.

The first feature could just to enable alerts to all agency people when a new job seeker joins the agency.  I propose we 1) send a R/T notification to all agency people who are logged in, and 2) send an email to *all* agency people at the same time.   That way we ensure that agency people are made aware of the new JS.

Then, we get some feedback from the client on how that works for him.

joaopapereira:
If the idea is to lessen the work maybe we should do only not real time and consider the real time after

PR #185, opened 2016-02-23
Closes AgileVentures/MetPlus_tracker#197

patmbolger:


app/assets/javascripts/agency_admin.js
```diff
@@ -12,13 +12,41 @@ var VisibilityToggle = {
...
-    $('#toggle_companies').click(VisibilityToggle.toggle);
-  }
+    $('#toggle_companies').click(AgencyData.toggle);
+    $('#companies_table').on('click', '.pagination a', AgencyData.update_data);
+  },
 };
+
+$(AgencyData.setup_branches);
+$(AgencyData.setup_people);
+$(AgencyData.setup_companies);
```
amba178: Probably, I would do: 
$(function() {
    put the last last three lines in here!
});
will make sure the DOM is loaded before the execution takes place.
Looks like you decouple the javascript from views.... nice move.

app/assets/javascripts/agency_admin.js
```diff
@@ -12,13 +12,41 @@ var VisibilityToggle = {
 
     return(false);
   },
+  update_data: function () {
+    var link_url = $(this).attr('href');
```
joaopapereira: Maybe you could have done this without the if with something like
$(this).parent()
This would get you the table object

app/assets/javascripts/agency_admin.js
joaopapereira: Good comment

app/views/agency_admin/home.html.haml
```diff
@@ -8,26 +8,42 @@
...
+  .col-sm-9
+    - if @branches.empty?
+      %h5 There are no branches.
+    - else
+      = render 'branches/branches'
 
 .row
-  - if @agency.branches.empty?
+  - if @branches.empty?
     .col-sm-2.col-sm-offset-10#table_action_button
```
joaopapereira: We shouldn't have this code repeated

app/views/agency_admin/home.html.haml
```diff
@@ -35,7 +51,7 @@
           method: :get, class: 'btn btn-primary btn-sm pull-right'  |
```
joaopapereira: We shouldn't have this code repeated

joaopapereira:
I beloved we already have the test to support this right?

joaopapereira:

app/views/agency_admin/home.html.haml
patmbolger: I have tried to refactor so as not to repeat some code.  For branches, this works for initial display:

```
.row
  %hr
  .col-sm-3
    %h4 Agency Branches

  - if @branches.empty?
    .col-sm-9
      %h5 There are no branches.
  - else
    .col-sm-9
      = render 'branches/branches'
    %div
      .col-sm-3.col-sm-offset-3
        %a(id='toggle_branches' href='#branches_table' class='text-danger')
          Hide Branches

  .col-sm-2.col-sm-offset-4#table_action_button
    = button_to 'Add Branch', new_agency_branch_path(@agency.id), |
        method: :get, class: 'btn btn-primary btn-sm pull-right'  |
```

However, when the 'Hide Branches' link is clicked, the arrangement of the 'Show Branches' link and the 'Add Branch' button are skewed because they are not appearing on a separate row (separate from the branches table).

I might be able to do something in javascript to counter this, but that would be ugly in and of itself.

Maybe I can come back to this later but for now I need to maintain the current structure.  It may not be pretty but it works, and the duplicated bits are all on the same page and in proximity to each other, so I don't think this is a maintenance problem.

app/assets/javascripts/agency_admin.js
patmbolger: Actually, that would give me the div that contains the table.  I need the actual table ID.

patmbolger:
I am going to look into adding functional tests for the XHR actions.  Will do that later today.

patmbolger:
PR updated.

app/assets/javascripts/agency_admin.js
joaopapereira: I am ok with your comment, mine was a suggestions, because I would not like to have to change that if every time that we need to add something.
I am ok with leaving that as is, but if we need to add a new branch then we will need to change that into a different javascript format

app/views/agency_admin/home.html.haml
joaopapereira: Have you tried to keep the .col-sm-9 outside of the if?

joaopapereira:
The new tests look good

app/views/agency_admin/home.html.haml
patmbolger: Yes, tried that and a number of other things.

patmbolger:
Refactored agency_admin.js - this should be good to go now.

joaopapereira:
Ok then, please just change the suggestion by Salem and I believe we can merge this PR

patmbolger:
PR updated.

joaopapereira:
Looks good, going to merge

PR #189, opened 2016-02-29
Closes AgileVentures/MetPlus_tracker#183

sidhene:


sidhene:
Reviewers: you can mostly disregard the changes to jobseekers controller and controller spec files. The only substantive change is on line 105 on the spec.

patmbolger:
Looks OK to me.

Related to this, could you add to this PR a change in spec/factories/job_seeker_statuses.rb from:

```
FactoryGirl.define do
  factory :job_seeker_status do
    value "MyString"
    description "MyTextString"
  end
end
```
to:
```
FactoryGirl.define do
  factory :job_seeker_status do
    value "Employedlooking"
    description "Currently employed but looking for better opportunity"
  end
end
```

sidhene:
Sure, np. Done. Pushed.

joaopapereira:
Do we have any test to check when we try to save the model without the status?

joaopapereira:
I understand this pr undo a lot of space and stuff but the changes are not related with this PR we should try to minimize these changes that are unrelated

sidhene:
@joaopapereira - that test is line 18 of "spec/models/job_seeker_spec.rb" as described above

Since git tracks any and all changes, I'm not sure that combing through "spec/controllers/job_seekers_controller_spec.rb" to make sure every line of code is in the same position it was before will improve the situation. Most likely,  that will only introduce even more changes which will then require yet another review.

spec/models/job_seeker_spec.rb
```diff
@@ -15,9 +15,10 @@
 
   	it {is_expected.to validate_presence_of(:year_of_birth)}
   	xit {is_expected.to validate_presence_of(:resume)}
+    it {is_expected.to validate_presence_of(:job_seeker_status_id)}
```
joaopapereira: Is the indentation correct?

joaopapereira:
@sidhene Do not understand your comment

sidhene:
Meaning: "spec/controllers/job_seekers_controller_spec.rb" has a lot of pink and green lines that reflect code that was added and removed. So the net is no change, or minor hard-to-trace changes like spaces. Since they're not "real changes" (like adding, replacing, or deleting code) they are not human-readable. They only show up in Git. If I try to do as you ask, I can't guarantee that the net effect won't be only more of the same.

joaopapereira:
I understand that but we should not be reformatting the code per commit and editor.

sidhene:
I'm not. I started to write some tests. They were the wrong tests. I deleted them. So I've already removed the changes.

spec/models/job_seeker_spec.rb
joaopapereira: Is the indentation correct in this line?

joaopapereira:
@sidhene can you please check my comment in job_Seeker_spec.RB ?

spec/models/job_seeker_spec.rb
sidhene: it's good

joaopapereira:
Going to merge, thanks for checking

PR #193, opened 2016-03-08
Connects to AgileVentures/MetPlus_tracker#210

patmbolger:
This is the first phase for job_category management.  It implements "create" capability.

To be delivered later are "delete", "edit" and "search" capability.

These actions will all take place on one page (/agency_admin/job_properties/) in order to provide a reasonable user experience.

### DIFF
--- a/app/assets/javascripts/agency_admin.js
+++ b/app/assets/javascripts/agency_admin.js
@@ -18,51 +18,30 @@ var AgencyData = {
     // (anchor is contained in 'li' element which will have 'disabled' class)
     if ($(this).parent().hasClass('disabled')) { return false; };
 
+    // 'event.currentTarget' is the div element that delegated the event
+    var div_id = '#' + $(event.currentTarget).attr('id')
+
     // 'this' is anchor element that recieved the event
     var link_url = $(this).attr('href');
 
-    // 'event.currentTarget' is the div element that delegated the event
-    var table_id = '#' + $(event.currentTarget).attr('id')
-
-    $.ajax({type: 'GET',
-            url: link_url,
-            timeout: 5000,
-            error: function (xhrObj, status, exception) {
-                              alert('Server Timed Out');},
-            success: function (data, status, xhrObject) {
-                      alert('In Success Function');
-                                 $(table_id).html(data);}
-            });
+    AgencyData.get_updated_data(div_id, link_url);
     return(false);
   },
-  get_update_data: function(div_id) {
-    // 'this' is anchor element that recieved the event
-    var link_url = $(this).attr('href');
-
+  get_updated_data: function(div_id, link_url) {
+    // calls link_url to get data and replace content of div with that
     $.ajax({type: 'GET',
             url: link_url,
             timeout: 5000,
             error: function (xhrObj, status, exception) {
                               alert('Server Timed Out');},
             success: function (data, status, xhrObject) {
-                      alert('In Success Function');
                                  $(div_id).html(data);}
             });
-    return(false);
   },
   add_job_category: function () {
-    // set up action_url (no id needed so can hard-wire)
-    // get category name
-    // get category description
-    // send AJAX request (controller adds category and returns success)
-    // - on success:
-    //   - determine 'active' pagination anchor
-    //   - 'click' that anchor link
-    //   - return true
-    // - on error (for status = 422)
-    //   - get the model error messages
-    //   - render the error messages in the modal window
-    //   - return false
+    // this function is bound to 'click' event on 'Add Category' button
+    // in 'agency_admin/_add_job_category.html.haml'
+    // Create the job category .....
     $.ajax({type: 'POST',
             url: '/job_categories/create/',
             // Get the data entered by the user in the dialog box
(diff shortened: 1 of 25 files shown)


joaopapereira:
Is this a mergeable pull request?

app/assets/javascripts/agency_admin.js
```diff
@@ -18,21 +18,71 @@ var AgencyData = {
...
-                                 $(table_id).html(data);}
+                                 $(div_id).html(data);}
             });
-    return(false);
+  },
+  add_job_category: function () {
+    // this function is bound to 'click' event on 'Add Category' button
+    // in 'agency_admin/_add_job_category.html.haml'
+    // Create the job category .....
+    $.ajax({type: 'POST',
```
joaopapereira: Can't we use rails function for address?

patmbolger:
Yes it is

app/assets/stylesheets/application.css
```diff
@@ -49,7 +49,7 @@
 
 /* this is for a button associated with a table above it. It is set close
    to the table. See agency admin home view for example. */
-#table_action_button {
+.table_action_button {
   margin-bottom: 30px;
```
joaopapereira: I assume the current change do not break any place that is currently using this css

app/views/agency_admin/_add_category_modal.html.haml
```diff
@@ -0,0 +1,20 @@
+.modal.fade#add_job_category(tabindex='-1' role='dialog')
```
joaopapereira: I have seen modal in a bunch of places can't we have a template for the modal where we just change the interior so we have consistency?

config/routes.rb
```diff
@@ -4,7 +4,7 @@
       :path_names => {sign_up: 'new', sign_out: 'logout',
                       sign_in: 'login' },
       :controllers => {invitations: 'people_invitations',
-                       confirmations: 'users/confirmations' }
```
joaopapereira: We should be consistent with our indentation. If you see the path_names is different

features/agency_admin.feature
```diff
@@ -163,8 +193,21 @@ Scenario: cannot remove sole agency admin
...
 Scenario: non-admin does not see 'admin' in menu
+  And I click the "Agency and Partner Companies" link
   Given I log out
   Given I am on the home page
   And I login as "jane@metplus.org" with password "qwerty123"
   Then I should see "Signed in successfully."
   And I should not see "Admin"
+
+@javascript
+Scenario: add job category
```
joaopapereira: Maybe we need to check some failure scenario don't you think?

app/assets/javascripts/agency_admin.js
patmbolger: Not sure what you mean ... this is on the client side so Rails helpers not available

app/assets/stylesheets/application.css
patmbolger: no breakage

app/views/agency_admin/_add_category_modal.html.haml
patmbolger: I intend to make this modal into a generic partial that can be called in other instances - that will occur when I add "Skills" management to the Job Properties admin page.

This modal is unlike other modals in that it will act as a form for simple objects - including the display of model validation objects.  This is useful for simple objects that should not require a completely separate page for create/update actions.

config/routes.rb
patmbolger: Please explain in more detail what you mean.  The indentation looks OK to me.

joaopapereira: Isn't confirmations strtin at the space instead under the first letter? Like the previous lines?

app/assets/javascripts/agency_admin.js
joaopapereira: You are right

app/views/agency_admin/_add_category_modal.html.haml
joaopapereira: What i was proposing is to have the modal structure separated and the modal body would load the template needed. So it can be a form or a simple message

config/routes.rb
joaopapereira: Strtin == starting :)

app/views/agency_admin/_add_category_modal.html.haml
patmbolger: I'm still not quite sure of what you're getting at.  I have refactored some of the "scaffolding" around the modal dialog creation.  In particular, there is this helper in application_helper.rb:

```
def text_modal_button(button_text, target_div_id, btn_class='btn btn-warning btn-xs')
    haml_tag('button', button_text, class: btn_class,
              data: {toggle: 'modal', target: "\##{target_div_id}"} )
  end
```

This allows us to just call this helper when we want a modal dialog associated with an activation button (which is most if not all of the time).  An example from job_properties.html.haml:

```
.col-sm-3.col-sm-offset-3.table_action_button
    - text_modal_button 'Add Job Category', 'add_job_category', |
                        'btn btn-primary btn-sm pull-right'     |

= render 'add_category_modal'
```

The last line above loads the modal partial, which is very specific to the particular context and cannot be made more generic (at least, as far as I can determine).  In fact, the very first line of the partial includes the div ID that was passed in to the helper function (in this case '#add_job_category').

Can we discuss when you're back home?  If we can achieve a better structure I'm all for it .... but would love to get this PR merged in the meantime if possible.

config/routes.rb
patmbolger: I understand what you meant now.

I have adopted an approach for this kind of construct that includes aligning multiple lines by the visual break between identical data types.  That is, using this:

```
 :controllers => {invitations: 'people_invitations',
                confirmations: 'users/confirmations'
```

instead of this:

```
 :controllers => {invitations: 'people_invitations',
                  confirmations: 'users/confirmations'
```

I find it easier to visually scan up and down and find what I'm looking for that way (especially in a statement with many lines).  However, I think I am an outlier here so will change this to reflect the other style.

patmbolger:
Changes made - ready for review.

joaopapereira:
Looks good, going to merge

PR #199, opened 2016-03-14
Closes AgileVentures/MetPlus_tracker#210

patmbolger:
This is the job_category "delete" feature.

### DIFF
--- a/app/assets/javascripts/agency_admin.js
+++ b/app/assets/javascripts/agency_admin.js
@@ -80,6 +80,46 @@ var AgencyData = {
           });
     return(false);
   },
+  delete_job_category: function () {
+    $.ajax({type: 'DELETE',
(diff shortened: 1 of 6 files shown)


joaopapereira:
Generally looks good, but I am afraid we are starting to have some code in JavaScript and we are not testing it thoroughly. Maybe we should start using Jasmine to test JavaScript parts as well

patmbolger:
I agree we need to start doing functional tests with Jasmine.  

How about I create a separate story for that and we merge this PR now.  I want to get job_categories in place so that development of jobs can continue without interruption.

joaopapereira:
Just wanted to point this out, but for next PR's I believe we should start doing so

joaopapereira:
Going to merge the PR

PR #201, opened 2016-03-16
Closes AgileVentures/MetPlus_tracker#221

patmbolger:


spec/javascripts/fixtures/company_registration.html
```diff
@@ -0,0 +1,131 @@
+<div class='row'>
```
joaopapereira: What is this file used for?

patmbolger: This is a fixture file contain HTML that the javascript will operate on for the tests.

joaopapereira:
Looks good going to merge

PR #202, opened 2016-03-17
connects AgileVentures/MetPlus_tracker#192

amba178:
- job< >company,  for job creation, editing, and deleting implicitly company person is involved in this relationship.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,13 +1,20 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
-  include ApplicationHelper 
-  protect_from_forgery with: :exception 
+  
+  protect_from_forgery with: :exception
+  helper_method :correct_user_type 
+
+  
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :store_current_location, :unless => :devise_controller?
   
   protected
+
+    def correct_user_type(current_user)
+       current_user.try(:actable).nil? ? current_user : current_user.actable
+    end
     def store_current_location
       store_location_for(:user, request.url)
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,12 +1,12 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
+  include ApplicationHelper 
   protect_from_forgery with: :exception 
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :store_current_location, :unless => :devise_controller?
- 
-
+  
   protected
     def store_current_location
       store_location_for(:user, request.url)
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -4,22 +4,22 @@ class JobsController < ApplicationController
 
 	def index
 		if company_person?
-			@jobs=@company_person.company.jobs.paginate(:page => params[:page],
+			@jobs = @company_person.company.jobs.paginate(:page => params[:page],
 				                                        :per_page => 32)
 		else
-			@jobs=Job.paginate(:page => params[:page],
+			@jobs = Job.paginate(:page => params[:page],
 				               :per_page => 32).includes(:company)
 		end
 	end	
 
 	def new
-		@job=Job.new(:company_id => params[:company_id])
+		@job = Job.new(:company_id => params[:company_id])
 	end 
 
 	def create
-		@job=Job.new(job_params)
+		@job = Job.new(job_params)
 		if @job.save 
-			flash[:notice]="#{@job.title} has been created successfully."
+			flash[:notice] = "#{@job.title} has been created successfully."
 			redirect_to jobs_url  
 		else
 			render :new 
@@ -34,7 +34,7 @@ class JobsController < ApplicationController
 
 	def update
 		if @job.update_attributes(job_params)
-			flash[:info]="#{@job.title} has been updated successfully."
+			flash[:info] = "#{@job.title} has been updated successfully."
 			redirect_to @job 
 		else
 			render :edit 
@@ -43,7 +43,7 @@ class JobsController < ApplicationController
 
 	def destroy
 		@job.destroy
-		flash[:alert]="#{@job.title} has been deleted successfully."
+		flash[:alert] = "#{@job.title} has been deleted successfully."
 		redirect_to jobs_url 
 	end
 
@@ -51,13 +51,13 @@ class JobsController < ApplicationController
 
 		def authentication_for_post_or_edit 
 			if !company_person?
-			   flash[:alert]="Sorry, You are not allowed to post or edit a job!"
+			   flash[:alert] = "Sorry, You are not allowed to post or edit a job!"
 			   redirect_to :back 
 			end  
 		end
 
 		def set_company_person
-			@company_person=correct_user_type(current_user) 
+			@company_person = correct_user_type(current_user) 
 		end
 
 		def company_person?
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -1,18 +1,25 @@
 class JobsController < ApplicationController
 	before_action :find_job,	only: [:show, :edit, :update, :destroy]
+	before_action :authentication_for_post_or_edit, only: [:new, :edit] 
 
 	def index
-		@jobs = Job.paginate(:page => params[:page], :per_page => 32).includes(:company)
+		if company_person?
+			@jobs=@company_person.company.jobs.paginate(:page => params[:page],
+				                                        :per_page => 32)
+		else
+			@jobs=Job.paginate(:page => params[:page],
+				               :per_page => 32).includes(:company)
+		end
 	end	
 
 	def new
-		@job = Job.new
-	end
+		@job=Job.new(:company_id => params[:company_id])
+	end 
 
 	def create
-		@job = Job.new(job_params)
+		@job=Job.new(job_params)
 		if @job.save 
-			flash[:notice] = "#{@job.title} has been created successfully."
+			flash[:notice]="#{@job.title} has been created successfully."
 			redirect_to jobs_url  
 		else
 			render :new 
@@ -27,7 +34,7 @@ class JobsController < ApplicationController
 
 	def update
 		if @job.update_attributes(job_params)
-			flash[:info] = "#{@job.title} has been updated successfully."
+			flash[:info]="#{@job.title} has been updated successfully."
 			redirect_to @job 
 		else
 			render :edit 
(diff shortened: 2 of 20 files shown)


app/helpers/application_helper.rb
```diff
@@ -5,6 +5,10 @@ def full_title(page_title = '')
     page_title.empty? ? base_title : "#{page_title} | #{base_title}"
   end
 
+  def correct_user_type current_user
```
joaopapereira: I beloved this should live inside the model maybe add functions like is_a_job_Seeker

app/controllers/jobs_controller.rb
```diff
@@ -37,11 +43,32 @@ def update
...
 		@job.destroy
-		flash[:alert] = "#{@job.title} has been deleted successfully."
+		flash[:alert]="#{@job.title} has been deleted successfully."
 		redirect_to jobs_url 
 	end
 
 	private 
+
+		def authentication_for_post_or_edit 
+			if !company_person?
```
joaopapereira: Shouldn't this be done with cancancan?

app/controllers/jobs_controller.rb
```diff
@@ -1,19 +1,25 @@
...
+		else
+			@jobs=Job.paginate(:page => params[:page],
+				               :per_page => 32).includes(:company)
+		end
 	end	
 
 	def new
-		@job = Job.new
-	end
+		@job=Job.new(:company_id => params[:company_id])
```
joaopapereira: What is the reason for removing the spaces around the equal signal?

app/controllers/jobs_controller.rb
```diff
@@ -1,19 +1,25 @@
...
+	end
```
joaopapereira: Do we need the space here?

app/models/company.rb
```diff
@@ -2,7 +2,7 @@ class Company < ActiveRecord::Base
   has_many :company_people, dependent: :destroy
     accepts_nested_attributes_for :company_people
 
-  has_many :jobs
+  has_many :jobs, dependent: :destroy
```
joaopapereira: Not 100%sure this should be done

app/models/job.rb
```diff
@@ -20,10 +20,11 @@ class Job < ActiveRecord::Base
   validates_length_of   :title, maximum: 100
   validates_presence_of :description
   validates_length_of   :description, maximum: 10000
+  validates_presence_of :company_id
 
   #These will be uncommented once 
   # we start implementing the features
-    # validates_presence_of :company_id
+
```
joaopapereira: I beloved we can remove this line

app/views/jobs/show.html.haml
```diff
@@ -2,19 +2,19 @@
 
 %div{class: "col-sm-offset-3 col-sm-6"}
 	%div{class: "pull-right", id: "job-show-links-id"}
-		=link_to "Delete", job_path(@job), method: :delete, 
-		data: {confirm: "Are you sure you want to delete the following job: #{@job.title}"}
-		|
-		=link_to "Edit Job", edit_job_path(@job)
-		|
+		-if correct_user_type(current_user).is_a?(CompanyPerson)
```
joaopapereira: This is done in a different way on _job.HTML.haml

joaopapereira:
Generally looks good, please review my comments
Wood be best if someone else would review the code as well

app/helpers/application_helper.rb
amba178: I think that would be ideal if it is specific to a jobseeker. For example in devise we have current_user instance which we can access in  any controllers and views, right. So instead of doing every time current_user.actable which will tell return either JobSeeker, CompanyPerson etc we just have to pass current_user to correct_user_type method.  It is like refactoring out current_user.actable from the controllers and views basically. I hope make sense.

app/models/company.rb
amba178: Maybe we need it. let me explain....let say we have a company A, has many jobs in the database, and for some reason a company A disassociate itself from agency that is deleted from database. So the jobs which are related to company A will be floating around the database, there are orphan. I think we do not want that to happen. What do you think?

app/models/job.rb
amba178: Sure.

app/models/company.rb
joaopapereira: And what if a Job Seeker Applied to it, what happen to that connection will we loose traceability on it?
What we need to do is create a defunct company or something to associate the jobs to it, but when a company is removed we should mark the jobs as none applicable.
Or another solution is to mark the Company as defunct and do not remove it from the database, but just disable all access to it, so we can keep any type of information.
This should be something that we need to bring to Chet attention to see what he think about it

app/controllers/application_controller.rb
```diff
@@ -1,12 +1,12 @@
 class ApplicationController < ActionController::Base
   # Prevent CSRF attacks by raising an exception.
   # For APIs, you may want to use :null_session instead.
+  include ApplicationHelper
```
patmbolger: I am a little leery of including these helper methods into a controller class.  As we know, the helpers are meant to bi visible to views, and to contain logic needed by the views (and prevent that logic from being in the views).

Including this module here makes all of the methods in the module instance methods of the controller.  That is probably not what we want, and mixes view code with controller code.

app/controllers/jobs_controller.rb
```diff
@@ -1,19 +1,25 @@
 class JobsController < ApplicationController
 	before_action :find_job,	only: [:show, :edit, :update, :destroy]
+	before_action :authentication_for_post_or_edit, only: [:new, :edit] 
 
 	def index
-		# @jobs = Job.paginate(:page => params[:page], :per_page => 32).includes(:company)
-		@jobs = Job.paginate(:page => params[:page], :per_page => 32)
+		if company_person?
```
patmbolger: We have quite a few class methods in the User model that are meant to support this kind of logic.  For instance, for this  case we could use the existing methods:
```
if User.is_a_company_admin?(current_user) || User.is_a_company_contact?(current_user)
```
If that is too verbose, we could define another method in user:
```
def self.is_a_company_person?
   User.is_a_company_admin?(current_user) || User.is_a_company_contact?(current_user)
end

File: app/models/company.rb (hunk 5 above)
Comment (patmbolger) [inline app/models/company.rb]: We already have a company status that can be set to indicate the company is no longer active.

I think we do need traceability to past jobs for an inactive company - if for no other reason than to support reporting job placement activity by the agency.

File: db/seeds.rb [hunk 10]
```diff
@@ -130,9 +129,7 @@
   200.times do |n|
     name = FFaker::Job.title
     description = FFaker::Lorem.sentence
-    if !JobCategory.create(name: name, description: description)
-      JobCategory.create!(name: "#{name}_#{n}", description: description)
-    end
+    JobCategory.create!(name: "#{name}_#{n}", description: description)
```
patmbolger: Bot sure why those 3 lines were deleted .... I put those in there because FFaker does not guarantee uniqueness of it's return values.  Thus, I was seeing intermittent failure of JobCategory create because of this.  However, since we were using "create" rather than "create!" these were ignored.  Then, later in the file, the array of Job Categories was looped though and each JC was assigned to another object.   That loop would fail if the JobCategory table had less than 200 records.  I added these 3 lines to more reliably ensure unique 'name' attribute, so the create! would not fail.

features/jobs.feature
```diff
@@ -1,37 +1,38 @@
...
-As a user #not quite correct,
+As a company person
 I want to create, update, and delete jobs
 
 background: adding job to database
 	Given the following jobs records:
-	| title               | company_job_id  | shift  | fulltime | description |
-	| software developer  | KRK12K  | evening| true     | internship position with pay|
+	| title               | company_job_id  | shift  | fulltime | description | company_id |
+	| software developer  | KRK12K  | evening| true     | internship position with pay| 3  |
```
patmbolger: What you might want to do here is add the creation of a real company to the DB in the background, then reference that company in the jobs table.

Then, include one or more views in the testing where the company name should appear and confirm that it does, in fact, appear correctly in the view.

app/models/company.rb
amba178: A very good point, this is the  case when a job seeker had had applied to the job before the company deactivated or deleted, and as far as I know there is no direct or indirect  relationship,  job<>jobseeker. we can ask ourselves does job seeker applies in person, at MetPlus site or redirect to company site to fill out application. After we sorted out, we can try to implement the traceability functionality. Moreover, we can ask the client the options for job seeker to take to apply to the job with specific company and if I remember Chet saying that we need to track all the job seekers who applied to the specific job.

features/step_definitions/devise_steps.rb
```diff
@@ -13,4 +13,11 @@
 Then(/^I log(?: ?)out$/) do
   step %{I click the "Log out" link}
 end
+
+
+Given(/^I am logged in as company person$/) do
+	company_person = FactoryGirl.create(:company_person)
```
patmbolger: Not sure of the purpose of this statement creating a company_person (?)

spec/controllers/jobs_controller_spec.rb
```diff
@@ -117,18 +156,9 @@
...
-    it "is a success" do 
-      expect(response).to have_http_status(:ok)
-    end
-
-    it "renders 'new' template" do
-      expect(response).to render_template('new')
-    end
-
-    it { should_not set_flash }
+  describe "Route => edit, sad path" do
```
patmbolger: isn't this the happy path?

app/models/company.rb
joaopapereira: I believe then that we are all in agreement that we need to revert this change

app/controllers/jobs_controller.rb
joaopapereira: Why cant the object respond to this instead of the class? doesnt make much sense to have the functions in the class level because we have the user object itself, maybe current_user should return the correct object instead of the User object

db/seeds.rb
amba178: Right, but by just doing  the following code, there is zero probability of duplication:  JobCategory.create!(name: "#{name}_#{n}", description: description) because of n which are not included in faker when generating title. But with deleted version there some probability of duplication thus < 200 records. Let me know if I am wrong.

app/controllers/jobs_controller.rb
amba178: I think is not that complicated, and no need to hit DB. we have cached version of user, which is current_user so we can just say, current_user.actable.is_a?(CompanyPerson) will do it. The same thing with company_person?

db/seeds.rb
patmbolger: What I had was this:
```
if !JobCategory.create(name: name, description: description)
      JobCategory.create!(name: "#{name}_#{n}", description: description)
end
```

which could definitely be shortened to one line as you suggest.  I did not do that in order prevent every company name from have "_<N>" appended to it.  But, since this is just fake data for dev, that does not really matter.

app/controllers/jobs_controller.rb
patmbolger: Yes, either:
```
current_user.actable_type == 'CompanyPerson'
```
or
```
current_user.actable.is_a?(CompanyPerson)
```
should do it.

The first way might not need to go to the DB, since it just uses 'User' attributes.  The second probably will, since user.actable will return the CompanyPerson object retrieved from the DB.

db/seeds.rb
amba178: I think if I am not wrong the  if !JobCategory.creat(....) part have a chance of throwing error while running seed otherwise I would left it original form.

app/controllers/jobs_controller.rb
amba178: Yes, we could. The reason I chose to do this way because it is fairly simple, and specific to a jobscontroller only, and  later on we can take care of the implementation NoMethodError and other errors on application level once we implement cancancan on the rest of controllers. What do you think?

app/controllers/jobs_controller.rb
amba178: No reason.

app/controllers/jobs_controller.rb
amba178: fix it. it looks better with spaces I guess.

app/models/company.rb
amba178: For now. Yes.

app/views/jobs/show.html.haml
amba178: Yes. I could have used just like show view, but here do not have access to instance of companyperson.

PR #204, opened 2016-03-22
Closes AgileVentures/MetPlus_tracker#235

patmbolger:
All features for managing 'Skills' in the agency admin page: add, edit, update, delete.


app/controllers/agency_admin_controller.rb
```diff
@@ -25,13 +25,34 @@ def home
...
+      when 'skills'
+        @skills = Skill.order(:name).
+                    page(params[:skills_page]).per_page(10)
+
+        render partial: 'job_properties', object: @skills,
+                locals: {data_type:  'skills',
+                         partial_id: 'skills_table',
+                         edit_property_path:   :edit_skill_path,
+                         delete_property_path: :skill_path}
+
```
joaopapereira: Shouldn't we render error when we do not know the data_type?

app/controllers/skills_controller.rb
```diff
@@ -0,0 +1,53 @@
+class SkillsController < ApplicationController
+  def create
+    skill = Skill.new(skill_params)
```
joaopapereira: Is this a Ajax request?

joaopapereira:
Once again this is a Huge change, and looks like it as things that are not related with this Story. Due to the size of the story I cannot judge it all by myself so I need someone else to review the code.

joaopapereira:
Another thing that I just saw, this story isn't even in the scope for the current sprint........

patmbolger:
Actually, everything in this PR is related to the story.

I used my discretion to move it into this sprint.  I can discuss at today's scrum.

app/controllers/skills_controller.rb
patmbolger: Yes.  I added a filter to confirm that only XHR requests are sent to the controller.

app/controllers/agency_admin_controller.rb
patmbolger: Added an exception if data_type is not recognized.

joaopapereira:
I assume this is correct and if no one add any comment till tomorrow afternoon I will merge this PR

PR #205, opened 2016-03-23
Closes AgileVentures/MetPlus_tracker#238

joaopapereira:
Changed email format and added tests to check it

### DIFF
--- a/app/views/devise/mailer/confirmation_instructions.html.haml
+++ b/app/views/devise/mailer/confirmation_instructions.html.haml
@@ -1,4 +1,5 @@
 %p
   Welcome #{@resource.first_name} #{@resource.last_name}!
+%p Thank you for registering in PETS.
 %p You can confirm your account email through the link below:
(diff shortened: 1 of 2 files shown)


app/views/devise/mailer/confirmation_instructions.html.haml
```diff
@@ -1,4 +1,4 @@
 %p
-  Welcome #{@email}!
+  Welcome #{@resource.first_name} #{@resource.last_name}!
 %p You can confirm your account email through the link below:
```
patmbolger: How about adding a line above "you can confirm your account ....":

"Thank you for registering with #{Agency.this_agency(@resource)} in PETS."

This gives the receiver more of an idea of what the email is about.

joaopapereira: Cannot be done because Agency.this_agency only work for AgencyPerson, for that we would need to create a function in each type of User to check what agency the user is associated with

PR #208, opened 2016-03-28
Connects to AgileVentures/MetPlus_tracker#133

patmbolger:
also,

Connects to AgileVentures/MetPlus_tracker#186


patmbolger:
This PR is the first phase of event management.

In this phase, events result in real-time popups.  The events are 1) job seeker joins PETS, 2) company registers request to be a partner with agency.  Any agency person logged in at the time will see a popup for each event.

Second phase will add emails sent for each event as well.  This may be via simple blocking code.  If so, I will defer async (background) email sending to a 3rd phase.

app/controllers/job_seekers_controller.rb
```diff
@@ -25,11 +25,11 @@ def update
     @jobseeker = JobSeeker.find(params[:id])
 
     person_params = jobseeker_params
-    if person_params['password'].to_s.length == 0
+    if person_params['password'].to_s.length == 0
```
joaopapereira: Why change this?

app/controllers/job_seekers_controller.rb
```diff
@@ -25,11 +25,11 @@ def update
     @jobseeker = JobSeeker.find(params[:id])
 
     person_params = jobseeker_params
-    if person_params['password'].to_s.length == 0
+    if person_params['password'].to_s.length == 0 
        person_params.delete('password')
        person_params.delete('password_confirmation')
     end
-
+
```
joaopapereira: Why change this?

app/controllers/job_seekers_controller.rb
```diff
@@ -39,7 +39,7 @@ def update
        render 'edit'
     end
   end
-
+
```
joaopapereira: Why change this?

app/models/event.rb
```diff
@@ -0,0 +1,13 @@
+class Event
+  include ActiveModel::Model
+
+  def self.create(evt_type, evt_data)
+    case evt_type
+    when :JS_REGISTER     # job seeker registered
+      Pusher.trigger('pusher_control', 'js_registered', evt_data)
```
joaopapereira: Maybe we should put these string into static class variables so it is easier to change if we need

config/routes.rb
```diff
@@ -1,14 +1,14 @@
 Rails.application.routes.draw do
 
-  devise_for :users,
-      :path_names => {sign_up: 'new', sign_out: 'logout',
-                      sign_in: 'login' },
-      :controllers => {invitations: 'people_invitations',
-                       confirmations: 'users/confirmations' }
+  devise_for :users, :path_names => {:sign_up => "new", :sign_out => 'logout',
```
joaopapereira: Any particular reason behind the change of formatting?

joaopapereira:
Looks good, please review my comments, and I would prefer that someone else would review the code as well

app/controllers/job_seekers_controller.rb
patmbolger: I don't know how these changes were introduced in this file.

config/routes.rb
patmbolger: Not really other than I had to add a key-value pair so fiddled a little with alignment at the same time.

app/controllers/job_seekers_controller.rb
joaopapereira: If you didnt change the file, please revert the change

joaopapereira:
Looks good going to merge the PR

PR #210, opened 2016-03-29
Close AgileVentures/MetPlus_tracker#202

sidhene:
Created a home page for JobSeeker that shows personal information, AgencyPerson relationship information, and job search results. Includes test coverage.

I'm holding off on the rebase until I can get all the change requests in, since multiple rebasing has been causing problems.

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -40,6 +40,13 @@ class JobSeekersController < ApplicationController
     end
   end
 
+  def home
+    @jobseeker = JobSeeker.find(params[:id])
+    @newjobs = Job.new_jobs(@jobseeker.last_sign_in_at).paginate(:page => params[:page], :per_page => 5)
+    @past_week = Job.new_jobs(Time.now - 3.weeks).paginate(:page => params[:page], :per_page => 5)
+
+  end
+
   def index
     @jobseeker = JobSeeker.all
   end
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -22,15 +22,6 @@ class Agency < ActiveRecord::Base
     user.actable.agency
   end
 
-  def agency_people_on_role role
-    users = []
-    agency_people.each do |person|
-      users << person if person.agency_roles &&
-          person.agency_roles.pluck(:role).include?(role)
-    end
-
-    users
-  end
 
   # MULTIPLE AGENCIES: the code below needs to change
   def self.all_agency_people_emails
@@ -38,11 +29,17 @@ class Agency < ActiveRecord::Base
   end
   ###################################################
 
-
   private
 
   def self.find_users_with_role(agency, role)
-    agency.agency_people_on_role role
+    users = []
+    agency.agency_people.each do |ap|
+                users << ap if ap.agency_roles &&
+                               ap.agency_roles.pluck(:role).include?(role)
+    end
+
+    users
   end
+
 
 end
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -42,13 +42,6 @@ class Agency < ActiveRecord::Base
 
   def self.find_users_with_role(agency, role)
     agency.agency_people_on_role role
-    users = []
-    agency.agency_people.each do |ap|
-                users << ap if ap.agency_roles &&
-                               ap.agency_roles.pluck(:role).include?(role)
-    end
-
-    users
   end
 
 
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -22,7 +22,16 @@ class Agency < ActiveRecord::Base
     user.actable.agency
   end
 
+  def agency_people_on_role role
+    users = []
+    agency_people.each do |person|
+      users << person if person.agency_roles &&
+          person.agency_roles.pluck(:role).include?(role)
+    end
 
+    users
+  end
+  
   # MULTIPLE AGENCIES: the code below needs to change
   def self.all_agency_people_emails
     first.agency_people.pluck(:email)
--- a/app/models/agency.rb
+++ b/app/models/agency.rb
@@ -31,7 +31,7 @@ class Agency < ActiveRecord::Base
 
     users
   end
-  
+
   # MULTIPLE AGENCIES: the code below needs to change
   def self.all_agency_people_emails
     first.agency_people.pluck(:email)
@@ -41,6 +41,7 @@ class Agency < ActiveRecord::Base
   private
 
   def self.find_users_with_role(agency, role)
+    agency.agency_people_on_role role
     users = []
     agency.agency_people.each do |ap|
                 users << ap if ap.agency_roles &&
--- a/app/models/agency_relation.rb
+++ b/app/models/agency_relation.rb
@@ -8,4 +8,10 @@ class AgencyRelation < ActiveRecord::Base
   def self.in_role_of role_key
     where(agency_role_id: AgencyRole.find_by_role(AgencyRole::ROLE[role_key]).id)
   end
+
+  # Helper methods for associating job seekers with agency people
+  # These business rules are enforced:
+  # 1) A job developer can have only one case manager
+  # 2) A job developer can have only one job developer ('primary' JD)
+
 end
--- a/app/views/job_seekers/home.html.haml
+++ b/app/views/job_seekers/home.html.haml
@@ -1,0 +1,45 @@
+%h3 Your Information
+%div{:class => 'container'}
+  %div{class: 'div_align_left'}
+    %b First Name:
+    = @jobseeker.first_name
+    %br
+    %b Last Name:
+    = @jobseeker.last_name
+    %br
+    %b Status:
+    = JobSeekerStatus.find(@jobseeker.job_seeker_status_id).value
+    %br
+    %b Last Login:
+    - unless @jobseeker.last_sign_in_at.nil?
+      = @jobseeker.last_sign_in_at.strftime('%x')
+
+  %div{class: 'div_align_left'}
+    %b Case Manager:
+    - if @jobseeker.case_manager.nil?
+      %i None assigned
+    - else
+    = @jobseeker.case_manager.full_name
+    %br
+    %b Job Developer:
+    - if @jobseeker.job_developer.nil?
+      %i None assigned
+    - else
+      = @jobseeker.job_developer.full_name
+      %br
+      %p
+  %div{class: 'div_align_left'}
+    = link_to 'Update your information', edit_job_seeker_path, class: 'btn btn-primary'
+
+
+%h3 Your Applications
+%i This feature has not been built yet
+
+%h3 Job Opportunities - Best Match
+%div
+  = render partial: 'jobs/compact_list', locals: {jobs: @past_week}
+.clearfix
+%h3 Job Opportunities - New
+%div
+  = render partial: 'jobs/compact_list', locals: {jobs: @newjobs}
+.clearfix
--- a/app/views/jobs/index.html.haml
+++ b/app/views/jobs/index.html.haml
@@ -10,16 +10,16 @@
 		%div{class: 'col-xs-4 lead'}
 			%h4{id: 'list-header'} List of job opportunities
 		%div{class: 'col-xs-8 lead text-right'}
-			%h5{id: 'hire-header'} 
+			%h5{id: 'hire-header'}
 				=link_to 'Are you hiring? Become a hiring partner!',
-				new_company_registration_path 
-- if @jobs.empty? 
+				new_company_registration_path
+- if @jobs.empty?
 	%div{class: "well well-lg col-sm-offset-2 col-sm-8", id: "sub-list-id"}
 		%strong{class: "text-center"}
 			%h3 There are no jobs available.
-- else 
+- else
 	- @jobs.each do |job|
 		%div{class: "well well-lg col-sm-offset-2 col-sm-8", id: "sub-list-id"}
-			= render job 
-%div{class: "col-sm-offset-4 col-sm-6"} 
-	= will_paginate @jobs 
+			= render job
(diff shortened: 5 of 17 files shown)


joaopapereira:
If this is going to be the final PR, nevertheless can be change please change the name to Close AgileVentures/MetPlus_tracker#202 so it can be tracked in waffle

app/controllers/job_seekers_controller.rb
```diff
@@ -40,6 +40,13 @@ def update
     end
   end
 
+  def home
+    @jobseeker = JobSeeker.find(params[:id])
```
joaopapereira: What happen if we cannot find the JobSeeker?

app/models/agency_relation.rb
```diff
@@ -8,4 +8,60 @@ class AgencyRelation < ActiveRecord::Base
   def self.in_role_of role_key
     where(agency_role_id: AgencyRole.find_by_role(AgencyRole::ROLE[role_key]).id)
   end
+
+  # Helper methods for associating job seekers with agency people
+  # These business rules are enforced:
+  # 1) A job developer can have only one case manager
+  # 2) A job developer can have only one job developer ('primary' JD)
+
+  def self.case_manager_for_job_seeker(job_seeker)
```
joaopapereira: As we discussed on the meeting I do not believe this is the correct place to put these functions, they should be in AgencyPeople, because we are trying to retrieve AgencyPeople objects

app/models/agency_relation.rb
```diff
@@ -8,4 +8,60 @@ class AgencyRelation < ActiveRecord::Base
...
+  end
+
+  def self.assign_job_developer_to_job_seeker(job_seeker, job_developer)
+    assign_agency_person_to_job_seeker(job_seeker, job_developer, :JD)
+  end
+
+  def self.assign_agency_person_to_job_seeker(job_seeker, agency_person, role_key)
+    raise 'Not a job seeker'   if not job_seeker.is_a? JobSeeker
+    raise 'Not an agency person' if not agency_person.is_a? AgencyPerson
+    raise 'AgencyPerson does not have appropriate role' if
```
joaopapereira: We should not have this logic here. Isn't there a method in the AgencyPerson to check if that Person as a specific role?

app/models/agency_relation.rb
```diff
@@ -8,4 +8,60 @@ class AgencyRelation < ActiveRecord::Base
...
+            agency_person.agency_roles.empty? ||
+            !agency_person.agency_roles.
+                  include?(AgencyRole.find_by_role(AgencyRole::ROLE[role_key]))
+
+    if (ap_relation = job_seeker.agency_relations.in_role_of(role_key)[0])
```
joaopapereira: Is this statement correct? assignation in the if?

joaopapereira:
Did you do integration tests?

joaopapereira:
Not 100% convinced specially in the relationship part of the PR.
Please review my comments above

app/views/job_seekers/home.html.haml
```diff
@@ -0,0 +1,44 @@
...
+    = @jobseeker.first_name
+    %br
+    %b Last Name:
+    = @jobseeker.last_name
+    %br
+    %b Status:
+    = JobSeekerStatus.find(@jobseeker.job_seeker_status_id).value
+    %br
+    %b Last Login:
+    = @jobseeker.last_sign_in_at.strftime('%x')
```
patmbolger: If the user accepts the link in the confirmation email, but then subsequently never logs in, I think this attribute (last_sign_in_at) will be nil - in which case this statement will raise an exception.

Please look into that and harden this statement if needed.

app/views/jobs/index.html.haml
```diff
@@ -10,16 +10,16 @@
...
+- else
 	- @jobs.each do |job|
 		%div{class: "well well-lg col-sm-offset-2 col-sm-8", id: "sub-list-id"}
-			= render job 
-%div{class: "col-sm-offset-4 col-sm-6"} 
-	= will_paginate 
\ No newline at end of file
+			= render job
+%div{class: "col-sm-offset-4 col-sm-6"}
+	= will_paginate
```
patmbolger: I don't see any pagination on this page.  Please fix that.

Also, since we'll have many job seekers and multiple pages, we'll need to sort the displayed list by last name.

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -168,6 +168,27 @@
...
+    end
+    it "returns http success" do
+      expect(response).to have_http_status(:success)
+    end
+
+    it "returns jobs posted since last login" do
+      @newjob = FactoryGirl.create(:job)
+      @newjob.assign_attributes(created_at: Time.now)
+      @jobseeker.assign_attributes(last_sign_in_at: (Time.now - 1.week))
+      expect(Job.new_jobs(@jobseeker.last_sign_in_at)).to include(@newjob)
```
patmbolger: This is a good test.

I suggest adding a few more jobs that were created within the last week.  And, also a few jobs that were created more than a week ago.  Then, the 'expect' can match returned jobs against an array that contains *only* the jobs that were created within the last week, and excludes the others.

app/models/agency_relation.rb
patmbolger: I wrote that code and, yes, the assignment within the if statement is intentional.  (perfectly valid Ruby)

app/views/job_seekers/home.html.haml
```diff
@@ -0,0 +1,44 @@
+%h3 Your Information
+%div{:class => 'container'}
+  %div{class: 'div_align_left'}
```
patmbolger: OK, I need to confess that I introduced the 'div_align_left' and 'div_align_right' styles back when I was first trying to get a column of field names (on the left) to line up with a column of values (on the right).  Subsequently I realized that this is a pretty bad way to do this for a number of reasons, including the fact that if one of the field descriptions is long and thus wraps, the next (lower) field and value pair will not be aligned with each other.

I have been using tables for this kind of thing from some point in the past on (for example, see app/views/agencies/_agency.html.haml).

If you don't mind a little extra work it would be good if you could convert this to use tables.  It is a much better technique and reduces that we'll repeat my original sin with further copy-and-paste.

app/models/agency_relation.rb
patmbolger: I struggled with this decision for the reason you cite, but chose to put these methods here because the two "heavy lifting" methods noted below are very tied up with the AgencyRelation model, and I did not feel it was appropriate for AgencyPerson methods to know that much about the internals of the AgencyRelation model.

The two "heavy lifting" methods are:

```
self.assign_agency_person_to_job_seeker
self.find_agency_person_for_job_seeker
```
How about we move the other methods to AgencyPerson and leave those two in AgencyRelation?

joaopapereira: When I see objects being passed to Class functions so many times, for me automatically smells.
I believe this Class should be only the relationship and not a logic placeholder

app/models/agency_relation.rb
joaopapereira: I understand the Ruby allow you to do this, as almost all the languages do, but:
irb
2.2.1 :001 > if(ab = 100)
2.2.1 :002?>   puts 'bam'
2.2.1 :003?>   else
2.2.1 :004 >     puts 'bammm'
2.2.1 :005?>   end
(irb):1: warning: found = in conditional, should be ==
bam
 => nil 

I think the code should be as clean as possible and having assignations in the if's make it more cryptic, and we the only win there is 1 line.

patmbolger: I agree it is a bit more cryptic.  I have no problem with having a separate assignment statement.

app/controllers/job_seekers_controller.rb
sidhene: This is the JobSeeker's own home page. Although we don't have permissions yet (right?) it shouldn't be visible to anyone else. And they should get our standard 'you are not authorized' error with a redirect to the main page.

app/views/job_seekers/home.html.haml
sidhene: I've inserted an 'unless nil' clause

app/views/jobs/index.html.haml
sidhene: I need to revert this to whatever it is in our current branch. This file was not supposed to be changed

app/models/agency_relation.rb
sidhene: This function (and associated tests) have been moved

app/models/agency_relation.rb
sidhene: moved

app/models/agency_relation.rb
sidhene: Done yesterday in parking lot

app/controllers/job_seekers_controller.rb
joaopapereira: I believe this is not a permission issue, if I go to /job_seekers/-1 what happens? 
or /job_seekers/10000000000?

spec/controllers/job_seekers_controller_spec.rb
sidhene: I've only added one job (to test the negative case) but it failed properly (when I tried to use "assign_attributes" to change the created date, turns out you have to use "update_attributes"). I think this is good enough for now.

app/controllers/job_seekers_controller.rb
amba178: Probably it will gives us NoRecordFound error, we could hand exceptions at application level and maybe redirect gracefully to any page we want. This can happen to other controllers so it would be nice to handle it at application level.

sidhene:
@joaopapereira - Regarding the cucumber tests, I'm reviewing what we have now and they're all workflows, with several actions and usually a series of views. This story has just one view and not a lot of action, so I'm not clear on what your'e looking for other than confirming that the page loads (which we're already doing with Rspec) and the expected information is there. Will you be at standup today?

app/controllers/job_seekers_controller.rb
joaopapereira: Maybe we can do that at application level, but nevertheless some sort of information should come out of here, like a string with the error that should be shown or something similar, that is the least in order to be able to have this setup at application level

joaopapereira:
So some ideas come to mind:
* Now we have a link to access the edit form for job seeker, we should update the cukes that edit JobSeeker to come through this new page
* Are we testing presence and not presence of Job Developer and Case Manager display?
* Are we checking when there are no jobs no jobs popup and when there are jobs they show up?

app/views/job_seekers/home.html.haml
sidhene: I spent some time working on this and remembered that the reason why I used %div is because that's the only way I could get that to come out looking ok. If you really want this, we can try to solve it Monday, but I'm going to change it back.

sidhene:
@joaopapereira - in the interest of getting this story and the next delivered, let's put these new cucumber tests in a different story.

joaopapereira:
The story associated with this PR is:
As a registered job seeker
When I login to PETS
I should be directed to a personalized 'home page'

Is this story working?
Is this PR the latest code?

sidhene:
Yes, except that we're using a navbar option instead of custom redirects on these stories.

app/models/agency.rb
```diff
@@ -31,18 +31,24 @@ def agency_people_on_role role
...
     first.agency_people.pluck(:email)
   end
   ###################################################
 
-
   private
 
   def self.find_users_with_role(agency, role)
-    agency.agency_people_on_role role
+    users = []
```
joaopapereira: This change need to be reverted

spec/models/job_seeker_spec.rb
```diff
@@ -21,22 +21,79 @@
...
-      expect(JobSeeker.acting_as? :user).to be true
-      expect(JobSeeker.acting_as? User).to  be true
-    end
-
-    it "returns false for anything other than supermodel" do
-      expect(JobSeeker.acting_as? :model).to be false
-      expect(JobSeeker.acting_as? String).to be false
-    end
-  end
+	context "#acting_as?" do
```
joaopapereira: Is this change of indentation correct?

joaopapereira:
@sidhene Please do the 2 changes I commented on.
After I will merge the PR

spec/models/job_seeker_spec.rb
sidhene: Yes, it was indented too far and I fixed it.

sidhene:
Done

app/models/agency.rb
sidhene: Done. Pushed.

app/models/agency.rb
```diff
@@ -38,11 +38,18 @@ def self.all_agency_people_emails
   end
   ###################################################
 
-
   private
 
   def self.find_users_with_role(agency, role)
     agency.agency_people_on_role role
+    users = []
```
joaopapereira: Previously I say to revert this change, it was to remove what you added and keep what was there before.
You need to remove this added code, there is no need to you to even change this file

sidhene: @joaopapereira - that's not my code. It's either from the Agency work Patrick did mid-story, or the interim work that happened on other branches that I merged before placing the pull request. 

Are you asking me to remove the entire method? I tried commenting it out to see what would happen and it causes 30 test failures.

sidhene: Ok, it looks like the change is limited to inside this one method - a longer and shorter version. I swapped out to the shorter version and nothing's failing. I'm pretty sure this is Patrick's work, so I'm going to ask him what the difference is. It sounds like the shorter version is the one you want, so I'm going to check with him to ensure there's not a reason to use the longer one.

sidhene: Confirmed (shorter). Done. Pushed.

joaopapereira:
Going to merge.
@sidhene do not forget to follow up with the 2 stories needed to end this story

PR #212, opened 2016-03-30
connects AgileVentures/MetPlus_tracker#192

amba178:


app/controllers/application_controller.rb
```diff
@@ -1,13 +1,20 @@
...
+  
 
   before_action :configure_permitted_parameters, if: :devise_controller?
   before_action :store_current_location, :unless => :devise_controller?
- 
-
+  
   protected
+
+    def correct_user_type(current_user)
```
joaopapereira: Looks good, but instead of calling it correct_user_type I would call it pets_user and instead of receiving a parameter I would create the function like this:
def pets_user
   return nil if current_user.nil?
   current_user.try(:actable).nil? ? current_user : current_user.actable
end
Using the helper from devise we should not have any problem

app/controllers/jobs_controller.rb
```diff
@@ -42,6 +48,27 @@ def destroy
...
+			   flash[:alert] = "Sorry, You are not allowed to post or edit a job!"
+			   redirect_to :back 
+			end  
+		end
+
+		def set_company_person
+			@company_person = correct_user_type(current_user) 
+		end
+
+		def company_person?
```
joaopapereira: I prefer this logic to be in a instance method of User for example

app/helpers/application_helper.rb
```diff
@@ -5,6 +5,10 @@ def full_title(page_title = '')
     page_title.empty? ? base_title : "#{page_title} | #{base_title}"
   end
 
+  def correct_user_type current_user
```
joaopapereira: Why is this function defined twice?

app/views/layouts/_footer.html.haml
```diff
@@ -8,7 +8,7 @@
 	%div{class: "panel panel-default panel-footer footer text-center"}
 		%div{class: "row text-center", id: "footer-text-color"}
 			For employers:
-			=link_to "Post jobs", new_job_path 
+			=link_to "Post jobs", new_job_path(company_id: correct_user_type(current_user).try(:company_id))
```
joaopapereira: Shouldnt this be guarded in the same way as the previous templates? Like only show the button when you are a company person or eventually we can keep it and redirect the user to login

joaopapereira:
Generally looks good, but please review my comment.
It would be nice if someone else could review this PR as well

app/controllers/jobs_controller.rb
```diff
@@ -42,6 +48,27 @@ def destroy
 	end
 
 	private 
+
```
patmbolger: This did not change with this PR, but I was wondering how #destroy works.  The first line is:
```
@job.destroy
```
I am not seeing how @job is being assigned before destroy (?).   Perhaps this was meant to be
```
find_job.destroy
```

app/controllers/application_controller.rb
amba178: I like that!.

app/controllers/jobs_controller.rb
amba178: well, we already have current_user available for us, it wouldn't be a good idea to deal with model. But it can be done. I will look at it.

joaopapereira: The problem here is that we have functions at class level that should be at instance level.
We can create a function in User like
def is_company_person?
   self.actable_type == "CompanyPerson"
end

app/views/layouts/_footer.html.haml
amba178: yeah. good idea.

app/controllers/jobs_controller.rb
amba178: before_action filter will set the instance of @job, will be available anywhere in the action, than we can do @job.destroy, @job.update_att..... etc Good question though.

app/helpers/application_helper.rb
amba178: well initial, I had in the helper module,and I made available everywhere by include at application level. But one of you guys did like the idea, I understand the risk  of expose all helper methods to rest of application so I create one for each for controllers and views. Perhaps I should have named it different.

app/controllers/jobs_controller.rb
patmbolger: Ah yes .... I should have looked for that!

app/controllers/jobs_controller.rb
```diff
@@ -42,6 +48,27 @@ def destroy
...
+		def authentication_for_post_or_edit 
+			if !company_person?
+			   flash[:alert] = "Sorry, You are not allowed to post or edit a job!"
+			   redirect_to :back 
+			end
```
patmbolger: Don't we also have to confirm that the specific job is owned by the company that the company_person is associated with?

app/helpers/application_helper.rb
patmbolger: Could you put that or similar method in User model?  That way it would be visible to both controllers, helpers and views.   Perhaps there is an architectural constant that I am not seeing.

patmbolger:
I reviewed the PR it is a good body of work.

I do share some concerns re user authorization as well as where we identify if the user is a CompanyPerson or not.   See my comments ...

app/controllers/jobs_controller.rb
amba178: address<>job  comes in handy, in here.

app/controllers/jobs_controller.rb
amba178: That will work correctly, but why don't we just check  current_user it's  actable_type?

amba178:
Updated the PR with relevant comments.

app/views/layouts/_footer.html.haml
```diff
@@ -7,11 +7,12 @@
 %div{class: "panel panel-default", id: "panel-default-id"}
 	%div{class: "panel panel-default panel-footer footer text-center"}
 		%div{class: "row text-center", id: "footer-text-color"}
-			For employers:
-			=link_to "Post jobs", new_job_path 
-			|
-			=link_to "Search Resumes", '#'
-			|
-			=link_to "Partner with Us", new_company_registration_path 
+			-if current_user.try(:actable).is_a?(CompanyPerson)
```
joaopapereira: why dont you use the pets_user helper?

app/views/jobs/show.html.haml
```diff
@@ -2,19 +2,19 @@
 
 %div{class: "col-sm-offset-3 col-sm-6"}
 	%div{class: "pull-right", id: "job-show-links-id"}
-		=link_to "Delete", job_path(@job), method: :delete, 
-		data: {confirm: "Are you sure you want to delete the following job: #{@job.title}"}
-		|
-		=link_to "Edit Job", edit_job_path(@job)
-		|
+		-if current_user.try(:actable).is_a?(CompanyPerson)
```
joaopapereira: Why dont you user the pets_user helper?

app/controllers/jobs_controller.rb
```diff
@@ -42,6 +48,27 @@ def destroy
...
+
+		def authentication_for_post_or_edit 
+			if !company_person?
+			   flash[:alert] = "Sorry, You are not allowed to post or edit a job!"
+			   redirect_to :back 
+			end  
+		end
+
+		def set_company_person
+			@company_person = pets_user
```
joaopapereira: if we have the pets_user do we really need this function anymore?

joaopapereira:
Please review my new comments

app/views/layouts/_footer.html.haml
amba178: Right, thanks.

app/views/jobs/show.html.haml
amba178: Same here.

app/controllers/jobs_controller.rb
amba178: if referring to set_company_person, yes. I am using it in the partial job.html.haml for job listing.

app/controllers/jobs_controller.rb
patmbolger: I pulled this feature branch and stood it up in my environment.  If I am logged in as a company person I can go to:
```
/jobs/<n>
```
where <n> is the ID of any job and I can edit and delete the job - regardless of whether I am a company person for the owning company or not.

I think it's probably OK to leave this as is for now, since we'll refactor in any case when we implement Cancancan auth.

On the other hand, if you put a fix in now that will be a good pattern to follow when switching to cancancan.

patmbolger:
I reviewed the PR again and other than a minor suggestion I am fine with it.  Good work.

app/controllers/jobs_controller.rb
amba178: Yes. at this point as long as you are company_person they do have access to any actions regardless of the company or a company at the specific address for now. it would be better if you review the association jobs<> with other models, and down the line this will be fixed. At a low level, if you can do with cancancan that would be great. Good point though.

joaopapereira:
looks good going to merge

PR #213, opened 2016-03-30
Closes to AgileVentures/MetPlus_tracker#204

joaopapereira:
Connects with AgileVentures/MetPlus_tracker#204
Initial approach for the tasks creation


app/models/agency.rb
```diff
@@ -21,6 +21,16 @@ def self.this_agency(user)
...
+
+  def agency_people_on_role role
+    users = []
+    agency_people.each do |person|
+      users << person if person.agency_roles &&
+          person.agency_roles.pluck(:role).include?(role)
+    end
+
+    users
+  end
```
patmbolger: This duplicates code in the class method immediately below it.  We should refactor to use one or then other.

app/models/company.rb
```diff
@@ -43,6 +44,16 @@ def other_company_admin?
...
 
+  def people_on_role role
+    users = []
+    company_people.each do |person|
+      users << person if person.company_roles &&
+          person.company_roles.pluck(:role).include?(role)
+    end
+
+    users
+  end
```
patmbolger: Same comment as above for agency.

app/models/task.rb
```diff
@@ -0,0 +1,48 @@
+class Task < ActiveRecord::Base
+  belongs_to :owner, class_name: 'User', foreign_key: 'target_user_id'
+
+  belongs_to :owner_agency, class_name: 'Agency', foreign_key: 'target_agency_id'
+
+  belongs_to :owner_company, class_name: 'Company', foreign_key: 'target_company_id'
+
```
patmbolger: Wondering why the owning references are referred to as targets?  Why not use foreign_key(s) that include 'owner' in the name instead of targets?

I was thinking it would since to use 'target' for the object that the task operates on, e.g, the agency admins all own a task to 'assign JD to JS', and the 'target' object for that task would be the JS object.

app/models/task.rb
```diff
@@ -0,0 +1,48 @@
...
+  def task_owner
+    return owner.pets_user if owner != nil
+    return owner_agency.agency_people_on_role AgencyRole::ROLE[target_agency_role.to_sym] if owner_agency != nil and target_agency_role != nil
+    return owner_company.people_on_role CompanyRole::ROLE[target_company_role.to_sym] if owner_company != nil and target_company_role != nil
+    nil
+  end
+
+  def task_owner=(user: nil, agency: {agency: nil, role: nil}, company: {company: nil, role: nil})
+    self.owner = nil
```
patmbolger: Not to be picky on naming convention, but we have generally adopted a style that includes referencing variables that are of a people-type as 'person', as distinct from the underlying base class User.  This makes for more natural reading of the code.  Thus instead of:
```
self.owner = user.user if user != nil
```
we could do:
```
self.owner = person.user if person != nil
```

joaopapereira:
When someone have some time, please review this PR. It is a bit big because it implements the full Task system :dart:

joaopapereira:
if no one see any problem with this pr I will merge it tomorrow morning

patmbolger:
I am going to look at it again tomorrow morning (not sure I can get to it today)

app/models/task.rb
```diff
@@ -0,0 +1,91 @@
...
+
+  belongs_to :user
+  belongs_to :company
+  belongs_to :job
+
+  belongs_to :task_setting
+
+  validates_with TaskOwnerValidator
+
+  scope :today_tasks, -> {where('deferred_date IS NULL or deferred_date < ?', Date.today)}
```
patmbolger: should that be ' ..... or deferred_date <= ?'  ?

app/models/task.rb
```diff
@@ -0,0 +1,91 @@
...
+      when Job
+        self.person = nil
+        self.company = nil
+        self.job = target
+      when Company
+        self.person = nil
+        self.company = target
+        self.job = nil
+    end
+  end
```
patmbolger: We might want to have multiple targets for some tasks.  For instance, a JS applies to a job.  The primary target would be the JS, of course.  And, looking at the JS activity will show a job application (event) having occurred.  However, what if the JS has applied for multiple jobs on the same day?   Then, we would just see multiple tasks for these events without any distinction as to which company the specific event is associated with.

patmbolger: Spoke too soon .... looks like you're using TaskSetting to manage the above scenario.

lib/task_manager/business_logic.rb
```diff
@@ -0,0 +1,67 @@
...
+##     def done_my_event person
+##         done_default 'no_events'
+##         # your code here
+##     end
+##   The 'no_events' tell the function to do not set events for this specific task or if you want to set the events
+##   using your code
+##
+
+  DESCRIPTIONS = {:need_job_developer => 'Job Seeker have no assigned Job Developer',
+                  :need_case_manager  => 'Job Seeker have no assigned Case Manager'}
```
patmbolger: should be 'Job Seeker *has* no assigned .....'

lib/task_manager/task_manager.rb
```diff
@@ -0,0 +1,150 @@
...
+      task = Task.new
+      task.task_owner = owner
+      task.user = target_user
+      task.job = target_job
+      task.company = target_company
+      task.task_type = task_type
+      task.status = STATUS[0]
+      task.save
+      task
+    end
```
patmbolger: Seems strange to see this functionality here instead  of in the Task model #initialize method (?)

lib/task_manager/business_logic.rb
```diff
@@ -0,0 +1,67 @@
...
+##
+##
+## If we need to overload the assign method for the event "my_event"
+## 1 - Add a function in InstanceMethods called that looks like this:
+##     def assign_my_event person
+##         assign_default person, 'no_events'
+##         # your code here
+##     end
+##   The 'no_events' tell the function to do not set events for this specific task or if you want to set the events
+##   using your code
```
patmbolger: The reference to events seems to obscure what is happening here.  Is ti more accurate to say:
```
If we need to overload the assign method for the task "my_task"
  - Add a function in InstanceMethods called that looks like this:

      def assign_my_task person
          assign_default person, 'no_events'
          # your code here
      end

The 'no_events' tell the function to not set events for this specific task.  Remove that if you want an event to be set, or you can set the event for this task in this method.
```

lib/task_manager/task_manager.rb
```diff
@@ -0,0 +1,150 @@
+module TaskManager
+
+  STATUS = ['New', 'Work in progress', 'Done', 'Assigned']
```
patmbolger: Are we sure we'll never need to access the STATUS constant outside of this module?  If so, we should use a hash structure ( { NEW: 'New, WIP: 'Work in Progress', etc. } )

patmbolger:
done with my review ... please see comments

app/models/task.rb
joaopapereira: Pff, not sure about task settings, I though I had removed it.....
About the target I see 2 options here keep everything as is and if you want to define only 1 then you use the target method or else you need to use the other setters( user, company, job )
or I can create something like task_owner that would receive possibly a hash with each value

joaopapereira: And I am going to remove TaskSettings because it is not used for anything

lib/task_manager/task_manager.rb
joaopapereira: This is our factory method that is why it is here, because it can also be overriden

patmbolger:
PR looks good to me.

PR #214, opened 2016-03-30
Connects to AgileVentures/MetPlus_tracker#133

patmbolger:
also,

Connects to AgileVentures/MetPlus_tracker#186

This is 2nd of 3 phases for these stories.  This add notification emails (delivered in blocking mode - 3rd phase will be background/async delivery).

### DIFF
--- a/app/mailers/agency_mailer.rb
+++ b/app/mailers/agency_mailer.rb
@@ -1,9 +1,22 @@
 class AgencyMailer < ApplicationMailer
 
-  def job_seeker_registered
+  def job_seeker_registered(agency_person, js_name, js_id)
+    send_notification_mail(agency_person, js_name, js_id, 'Job Seeker')
   end
 
-  def company_registered
+  def company_registered(agency_person, company_name, company_id)
+    send_notification_mail(agency_person, company_name, company_id, 'Company')
   end
-  
+
+  private
+
+  def send_notification_mail(agency_person, obj_name, obj_id, obj_type)
+    @agency_person = agency_person
+    @obj_name = obj_name
+    @obj_id   = obj_id
+    @obj_type = obj_type
+    mail to: agency_person.email,
+         template_name: 'agency_notification'
+  end
+
 end
--- a/app/mailers/agency_mailer.rb
+++ b/app/mailers/agency_mailer.rb
@@ -1,21 +1,20 @@
 class AgencyMailer < ApplicationMailer
 
-  def job_seeker_registered(agency_person, js_name, js_id)
-    send_notification_mail(agency_person, js_name, js_id, 'Job Seeker')
+  def job_seeker_registered(agency_emails, js_name, js_id)
+    send_notification_mail(agency_emails, js_name, js_id, 'Job Seeker')
   end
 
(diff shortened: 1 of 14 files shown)


app/mailers/agency_mailer.rb
```diff
@@ -0,0 +1,21 @@
+class AgencyMailer < ApplicationMailer
+
+  def job_seeker_registered(agency_emails, js_name, js_id)
```
joaopapereira: Wouldnt it be best to pass the JS object instead of parts of it?

app/mailers/agency_mailer.rb
```diff
@@ -0,0 +1,21 @@
...
+    send_notification_mail(agency_emails, js_name, js_id, 'Job Seeker')
+  end
+
+  def company_registered(agency_emails, company_name, company_id)
```
joaopapereira: Wouldnt it be best to pass the Company object instead of parts of it?

app/models/event.rb
```diff
@@ -9,9 +10,20 @@ def self.create(evt_type, evt_data)
     when :JS_REGISTER     # job seeker registered
       Pusher.trigger('pusher_control',
                      EVT_TYPE[:JS_REGISTER], evt_data)
+
+      # Send notification email to all agency personnel
+      agency_emails = AgencyPerson.where(agency: Agency.first).pluck(:email)
```
joaopapereira: Shouldn't this logic be inside the Agency to retrieve all the emails from AgencyPerson?

app/mailers/agency_mailer.rb
joaopapereira: Hmm I see you dont have the object there, shouldn't we get it and use it?

joaopapereira:
Generally looks good, please review my comments

app/mailers/agency_mailer.rb
patmbolger: Good question.  I designed *Event.create* to take two arguments: 1) a key for which event is to be created and 2) a task containing attributes needed for both Pusher and email notifications.  This allows us to add other events that may not be associated with a distinct object that contains all the attributes needed for notification.

For these first two events, I could certainly pass in the object (JS, Company) and be fine.  But other events may have different data attributes that don't lend themselves to that model.  So, a hash of attributes seems more generic for growth.

app/models/event.rb
patmbolger: You're right .... my head must have still been in prototype mode when I wrote that

app/models/agency.rb
```diff
@@ -10,27 +10,31 @@ class Agency < ActiveRecord::Base
...
     raise RuntimeError, 'Logged in user is not an agency person' unless
             user.actable.is_a? AgencyPerson
-            
+
     user.actable.agency
   end
-  
+
+  def self.all_agency_people_emails
+    first.agency_people.pluck(:email)
```
joaopapereira: This need to be changed in the future when we have multiple agencies, please add a comment or something similar

joaopapereira:
Please just review my latest commit and I believe we are ready for merge, unless anyone else have some question

joaopapereira:
Looks good going to merge

PR #218, opened 2016-04-07
Connects to AgileVentures/MetPlus_tracker#232

joaopapereira:
WIP to have a look into the implementation of the tasks.
Long PR because a bunch of templates and files needed to be created


app/assets/javascripts/task.js
```diff
@@ -0,0 +1,166 @@
...
+                      $("#assignTaskModal").modal('hide');
+                  },
+                  error: function (xhrObj, status, exception) {
+                      noty({
+                          text: xhrObj['message'],
+                          theme: 'bootstrapTheme',
+                          layout: 'bottomRight',
+                          type: 'error'
+                      });
+                  }
```
patmbolger: There is an opportunity to extract out the 'noty' functions into one utility function.  (that could also be used in pusher.js.erb)

app/helpers/application_helper.rb
```diff
@@ -31,5 +31,10 @@ def text_modal_button(button_text, target_div_id, btn_class='btn btn-warning btn
               data: {toggle: 'modal', target: "\##{target_div_id}"} )
   end
 
+  def show_person_path person
+    return job_seeker_path person if person.is_job_seeker?
+    return edit_profile_agency_person_path person if person.is_a? AgencyPerson
```
patmbolger: The user should not be directed to the edit-profile view for the task owner or target.  That view is (will be) limited to the actual person, and allows changing of user-level attributes like email and password.

I think this should just go to agency_person_path for an AgencyPerson.

app/helpers/application_helper.rb
```diff
@@ -31,5 +31,10 @@ def text_modal_button(button_text, target_div_id, btn_class='btn btn-warning btn
...
+    return edit_company_person_path person if person.is_a? CompanyPerson
```
patmbolger: See above comments.

I think this should direct to company_person_path.

app/controllers/task_controller.rb
```diff
@@ -0,0 +1,80 @@
...
+          if all_users.nil? or all_users.size == 0
+    all_users.each do |user|
+      if user.full_name.downcase =~ /#{term}/
+        list_users << {id: user.user.id, text: user.full_name}
+      end
+    end
+    render json: {:results => list_users}
+  end
+
+end
```
patmbolger: Is there a reason why this controller is named as a singular resource rather than the standard pluralized naming convention?  (task_controller.rb vs tasks_controller.rb).  

Probably does not make much difference, but the routes will look different and might be a source of confusion.

lib/task_manager/business_logic.rb
```diff
@@ -59,6 +61,15 @@ def description
     DESCRIPTIONS[task_type.to_sym]
   end
 
+  def assignable_list
+    return [] if not owner.nil?
```
patmbolger: Might be easier to understand if:

 ```return [] unless owner.nil?```

spec/javascripts/tasks_spec.js
```diff
@@ -0,0 +1,84 @@
+describe('Tasks', function () {
+    var taskManager = TaskManagerHolder('mine-open-tasks', 'mine-open');
+    beforeEach(function () {
+        loadFixtures('tasks/task_list.html');
+        taskManager.init();
+        TaskModal.setup();
+    });
+    describe("Check pagination", function () {
+        it('calls ajax to add job category', function () {
```
patmbolger: title should probably be 'calls ajax to load next page'

spec/javascripts/tasks_spec.js
```diff
@@ -0,0 +1,84 @@
...
+            spyOn($, 'ajax');
+            $('#next-page').triggerHandler('click');
+            expect($.ajax).toHaveBeenCalled();
+            expect($.ajax.calls.mostRecent().args[0]["url"]).toEqual("http://localhost:8888/task/tasks/mine-open?tasks_page=2");
```
patmbolger: I suspect most of us are running the app on the default port 3000.

patmbolger:
Finished review - please see comments.

spec/javascripts/tasks_spec.js
joaopapereira: This is jasmine URL, I can change this to the format of the other test and start using jasmine-ajax

joaopapereira:
Closed all the comments

joaopapereira:
This PR is no longer a WIP, but the implementation of the Story

patmbolger:
Reviewed PR again - looks good.

PR #219, opened 2016-04-07
Connects to AgileVentures/MetPlus_tracker#189

patmbolger:
Initial architecture for interacting with backend 'cruncher' service.  Primary components of PR:

1. CruncherService:
 - establishes communication channel
 - persists authorization token (class var)
 - handles communication errors
 - logs communication events

2. ResumeCruncher:
 - will provides specific API's for actions (post resume, retrieve resume, find jobs matching resume, etc.)
 - Interacts with CruncherService to effect communication with Cruncher backend
 - (anticipate adding an analogous 'JobCruncher' in future)

3. Resume model:
 - now a model as it needs to be validated and most of its attributes are persisted in DB
 - included in this PR to illustrate interaction with ResumeCruncher
 - (only minimal testing of this model in this PR)


app/services/cruncher_service.rb
```diff
@@ -0,0 +1,48 @@
...
+    raise "Invalid MIME type for file: #{file_name}" if mime_type.empty?
+    raise "Unsupported file type for: #{file_name}" if
+          not Resume::FILETYPES.include? mime_type.first.preferred_extension
+
+    result = RestClient.post(service_url + '/curriculum/upload',
+      { 'file'   => file,
+        'name'   => file_name,
+        'userId' => user_id },
+      { 'Accept' => 'application/json',
+        'X-Auth-Token' => auth_token,
```
joaopapereira: You don't need to retrieve the token for every request, in the Java side the session is persisted.
Just need to make that if you get a authorization error you need to reauthenticate

joaopapereira:
Looks good.
Some considerations:
 * You might need to reauthenticate after 30 minutes, so we need to check for that
 * In terms of testing we should stub the service or if you want you can use VCR gem to emulate the answer
 * Maybe I would user Concern instead of Model on ResumeCruncher

patmbolger:
I have converted this WIP to a real PR.

It is ready for review.

I will stub the requests to the Cruncher service in a separate (next) story.  I want to get this in soon in case someone wants to move ahead with résumé upload to the Cruncher.

app/services/cruncher_service.rb
```diff
@@ -0,0 +1,62 @@
...
+                'X-Auth-Token' => auth_token,
+                'Content-Type' => mime_type.first.content_type })
+
+      return JSON.parse(result)['resultCode'] == 'SUCCESS'
+
+    rescue RestClient::Unauthorized   # most likely expired token
+      # Retry and force refresh of cached auth_token
+      self.auth_token = nil
+      if retry_upload
+        retry_upload == false
```
joaopapereira: shouldn't this be 
`retry_upload = false`
?

joaopapereira:
Are you stubbing the Cruncher integration?
I think we should stub it because we cannot rely 100% on the server being up and for the UT I think we can use VCR gem to solve this

app/services/cruncher_service.rb
patmbolger: Yes, that is a mistake.

patmbolger:
My intent it to stub the Cruncher in my next story (maybe delivered today or tomorrow latest).

joaopapereira:
Looks good, going to merge

PR #223, opened 2016-04-08
Closes AgileVentures/MetPlus_tracker#258

joaopapereira:


patmbolger:
PR looks good to me.

patmbolger:
Actually, I forgot to mention that with the addition of 'state' to Address we should update method *single_line_address* in application_helper.rb.

joaopapereira:
I do not understand why this PR as that many commits, but I will open a Bug for the line, because it is not related with the current story

PR #224, opened 2016-04-09

amba178:
I have change the Job<>Address relation in the model. We need to updated our ERD, I think.


app/controllers/jobs_controller.rb
```diff
@@ -13,11 +15,14 @@ def index
 	end	
 
 	def new
-		@job = Job.new(:company_id => params[:company_id])
+		@job = Job.new(:company_id => params[:company_id],
+			           :company_person_id => params[:company_person_id])
```
joaopapereira: Can't a JD create a job?

app/controllers/jobs_controller.rb
```diff
@@ -13,11 +15,14 @@ def index
...
 	end 
 
 	def create
 		@job = Job.new(job_params)
+		 
+		@job.address = @company_person.address
```
joaopapereira: Is this assumption something that Chet asked for? A company person can only create a job at his address?

app/controllers/jobs_controller.rb
```diff
@@ -51,9 +58,12 @@ def destroy
 
 		def authentication_for_post_or_edit 
 			if !company_person?
-			   flash[:alert] = "Sorry, You are not allowed to post or edit a job!"
```
joaopapereira: Revert this change, we don't need spaces there

app/controllers/jobs_controller.rb
```diff
@@ -69,6 +79,14 @@ def company_person?
 			end
 		end
 
+		def is_company_person_job
+			 set_company_person
+			 if !(@job.address == @company_person.address)
```
joaopapereira: Not sure this if it's accurate

app/controllers/jobs_controller.rb
amba178: From my understanding JD work for agency, and we know that has some kind of ties with company that is help to match jobs with jobseekers, but  he/she not permitted to create a job, same applies to case manager. What do you think?

app/controllers/jobs_controller.rb
amba178: Ok.

app/controllers/jobs_controller.rb
amba178: I think so, walmart A  jobs can not be created, edited or delete by walmart B company_person vise verser. Otherwise there will be some kind of confusion. Can you send me Chet's email through slack private messaging/direct messaging if you can. I have some questions to ask.

app/views/layouts/_footer.html.haml
```diff
@@ -9,7 +9,7 @@
 		%div{class: "row text-center", id: "footer-text-color"}
 			-if pets_user.is_a?(CompanyPerson)
 				For employers:
-				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id))
+				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id), company_person_id: current_user.try(:actable_id))
 				|
```
joaopapereira: Does it make sense to show this link if you cannot post jobs?

app/views/layouts/application.html.haml
```diff
@@ -40,7 +40,7 @@
             - @model_errors.full_messages.each do |msg|
               %li= msg
       %div{:class => 'row'}
-        = yield
+        = yield
```
joaopapereira: We don't need this space

db/migrate/20151107120000_create_database.rb
```diff
@@ -4,8 +4,7 @@ def change
       t.string :street
       t.string :city
       t.string :zipcode
-      t.references :location, polymorphic: true, index: true
-
+      t.references :location, polymorphic: true, index: true
```
joaopapereira: This is not a good practice. If you need to change a table do a new migration. Before we go live we will merge all changes but in the meanwhile create new migrations

spec/controllers/jobs_controller_spec.rb
```diff
@@ -249,17 +247,21 @@
...
                              company_id: '3',
-                             :shift => "Evening", company_job_id: "WERRR123"}
+                             address_id: '2',
+                             company_person_id: 3,
+                             :shift => "Evening", 
+                             company_job_id: "WERRR123"}
       expect(response).to redirect_to(:action => 'index')
       should set_flash 
-      expect(Job.count).to eq(2)
+      expect(Job.count).to eq(3)
```
joaopapereira: This code doesn't look good. You should get the count before the post and then compare after

spec/factories/jobs.rb
```diff
@@ -3,8 +3,8 @@
     title "MyString"
     description "MyString"
     company_id 1
-    # company_person_id 1
-    # address
+    company_person_id 1
```
joaopapereira: Shouldn't we call the factories of company and address here?

joaopapereira:
Looks good please review my comments

app/views/layouts/_footer.html.haml
amba178: Only shows if you are company_person instance

app/views/layouts/application.html.haml
amba178: ok

db/migrate/20151107120000_create_database.rb
amba178: you are right. But here I do not why there are two return lines, and will be fixed.

spec/controllers/jobs_controller_spec.rb
amba178: yes. Good suggestion. thanks

spec/factories/jobs.rb
amba178: actually, should be foreign key for address, instead of instance of address.

app/controllers/jobs_controller.rb
patmbolger: This is something we need to check with the client on.

The client wrote this:

```
 I want a standard format that employers and job developers can input job leads in which will be recognizable by the PETS software.
```
in this document: http://www.agileventures.org/projects/metplus/documents/public-employment-tracking-system-description-1

Then, at the start of the project we wrote this requirement for the interaction between job developers and employers:
```
JD:E/3	JD [job developer] creates a job opportunity from the E [employer] description
```
in this document: http://www.agileventures.org/projects/metplus/documents/roles-and-interactions

app/controllers/jobs_controller.rb
patmbolger: I think this is way too restrictive.

A company can have many addresses - reflecting locations where the company may have available jobs.

A company person has an address.  We don't restrict that address to being in the set of company addresses, nor should we necessarily do that.  IMO, this is for assistance in contacting a company person if we need to.

A job address might be restricted to being one of the company addresses - that seems reasonable to me.

However, this code would prevent an HR company person (whose address is the company's HQ, or their home if they are a remote employee, or anywhere really) from posting jobs for other locations.

app/models/address.rb
```diff
@@ -1,6 +1,6 @@
 class Address < ActiveRecord::Base
   belongs_to :location, polymorphic: true
-  
+  has_many :jobs
```
patmbolger: yes, we need to change the ERD to reflect this.  I will do that.

Also, while you're at it you may want to add
```
has_many :company_people
```
which has been in the ERD but not reflected in the Address model yet.

patmbolger:
My review complete - please see comments.

app/models/address.rb
amba178: I will. Let set PP time just to work on ERD I think we might need to change or add associations on other models too that way we are on the same page with each other.

app/controllers/jobs_controller.rb
amba178: it shows changes on flash message , I think. not spaces.

PR #227, opened 2016-04-11
Closes AgileVentures/MetPlus_tracker#261

patmbolger:


spec/models/resume_spec.rb
```diff
@@ -53,6 +74,11 @@
     end
 
     it 'fails with invalid model and valid file type' do
+
+      stub_request(:post, CruncherService.service_url + '/curriculum/upload').
```
joaopapereira: I am not following this test, because the name says it should fail and you stub success from the cruncher.

joaopapereira:
For the spec/models/resume_spec.rb you could have just mocked the service, instead of having the process flowing through it

spec/services/cruncher_service_spec.rb
```diff
@@ -24,16 +24,36 @@
...
+        stub_request(:post, CruncherService.service_url + '/curriculum/upload').
+            to_return(body: "{\"resultCode\":\"SUCCESS\"}", status: 200,
+            :headers => {'Content-Type'=> 'application/json'})
+
         expect(auth_result.code).to eq 200
       end
 
       it 'returns authorization token' do
+
+        stub_request(:post, CruncherService.service_url + '/curriculum/upload').
```
joaopapereira: Why do you stub the upload in this call?

spec/services/cruncher_service_spec.rb
```diff
@@ -24,16 +24,36 @@
...
+    before(:each) do
+      stub_request(:post, CruncherService.service_url + '/authenticate').
+          to_return(body: "{\"token\": \"12345\"}", status: 200,
+          :headers => {'Content-Type'=> 'application/json'})
+    end
+
     describe 'authorization' do
       it 'returns HTTP success' do
+
+        stub_request(:post, CruncherService.service_url + '/curriculum/upload').
```
joaopapereira: Why do you stub the upload in this call?

spec/services/cruncher_service_spec.rb
```diff
@@ -24,16 +24,36 @@
...
         expect(auth_result.code).to eq 200
       end
 
       it 'returns authorization token' do
+
+        stub_request(:post, CruncherService.service_url + '/curriculum/upload').
+            to_return(body: "{\"resultCode\":\"SUCCESS\"}", status: 200,
+            :headers => {'Content-Type'=> 'application/json'})
+
         expect(JSON.parse(auth_result)['token']).not_to be_nil
```
joaopapereira: Here now you can check the value of the token to be the correct one instead of not nil

joaopapereira:
Looks good, but have some questions in some parts, please review my comments

spec/models/resume_spec.rb
patmbolger: On resume.save there are two "saves" occurring - one is the model instance and one is the file upload to Cruncher.  This is now of 4 tests for each of the possible outcomes of each "save".

spec/services/cruncher_service_spec.rb
patmbolger: There is no attempt to get an auth token until the first actual operation requires it - in this case, the file upload triggers the authentication to occur.

patmbolger: That is not needed - removed.

spec/services/cruncher_service_spec.rb
patmbolger: That is not needed - removed.

joaopapereira:
Looks good going to merge

PR #230, opened 2016-04-12
Closes AgileVentures/MetPlus_tracker#262

patmbolger:


db/seeds.rb
```diff
@@ -3,18 +3,21 @@
 
 puts "Seeding wait......."
 #incase rerun comes in. not necessary
-AgencyRelation.delete_all
```
joaopapereira: This change need to be reverted, because in PG the order used to delete record matter.
In PG foreign_key constraint are enforced, so we need to remove the objects in a specific order

patmbolger:
Removed DB table deletes from seeds.rb

joaopapereira:
Looks good going to merge

PR #231, opened 2016-04-13
Closes AgileVentures/MetPlus_tracker#264

patmbolger:


lib/task_manager/task_manager.rb
```diff
@@ -48,10 +48,10 @@ def unschedule_event event_type, task, role_to_notify
 
     private
     ## Method that will create the task itself
-    def new_task task_type, owner, target_user = nil, target_job = nil, target_company = nil
+    def new_task task_type, owner, target_person = nil, target_job = nil, target_company = nil
       task = Task.new
       task.task_owner = owner
-      task.user = target_user
+      task.user = target_person.user if target_person
```
joaopapereira: use 
task.person = target_user
should be enough

spec/models/task_spec.rb
```diff
@@ -166,30 +166,37 @@
     it 'job seeker user' do
       @task.target = @job_seeker
       expect(@task.target).to eq @job_seeker
+      expect(@task.user).to eq @job_seeker.user
```
joaopapereira: we shouldnt use directly the user method but the person method, I will make user private for tasks

patmbolger:
Updated per comments.

joaopapereira:
Looks good going to merge

PR #233, opened 2016-04-13
Connects to AgileVentures/MetPlus_tracker#253 (WIP-pundit)

joaopapereira:
Option using pundit

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -45,7 +45,8 @@ gem 'bootstrap-will_paginate'
 
 # gem 'ajax_pagination'
 
-gem 'cancancan', '~> 1.10'
+#gem 'cancancan', '~> 1.10'
+gem "pundit"
 
 gem 'pusher'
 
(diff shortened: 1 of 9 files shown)


patmbolger:
This looks pretty good .... I really like:

1. You were able to cleanly set up auth for non-CRUD controller actions (assign, in_progress, done)
2. The policy scope stuff will be very useful for building views in which the authorized content will vary based on type of person, which we will have (e.g. AgencyPerson home page with scopes that show different content based on role)
3. I like that it is plain-old Ruby, and not tied solely to AR models (e.g. Event class is not an AR model, but needs to have auth wrapped around it).

Admittedly, I have not looked closely at cancancan, so that tool may have the same or similar advantages as Pundit.  In any case, Pundit looks useable and powerful without getting in the way.

João - you're the only person on the team who's worked with both.  Curious as to your opinion ...

PR #234, opened 2016-04-14
Connects to AgileVentures/MetPlus_tracker#253 (WIP-cancancan)

joaopapereira:


patmbolger:
This spike shows cancancan working cleanly.

Looking further, I think cancancan Ability class will get fairly big and complex, because we have so many authorizations that depend on not only user and action but also role. And, possibly, the role-base relation between a user (say, job developer) and a person on the other side of that relationship (e.g. job seeker).

On the other hand, Pundit will cause us to have lots of policy files that instantiate discrete authorizations per class.  Having lots of files is probably ok, and effectively breaks up a large set of authorization methods into small sets related to the specific class policy.

Also, cancancan seems more complicated in its DSL than Pundit.  And, there is a lot of "magic" going on that may be powerful but also may make things more obscure if trying to chase down a bug (for example, read this: https://github.com/CanCanCommunity/cancancan/wiki/authorizing-controller-actions).

joaopapereira:
Closing this PR because we are going to use pundit

PR #237, opened 2016-04-14
Closes AgileVentures/MetPlus_tracker#273

patmbolger:


spec/services/cruncher_service_spec.rb
```diff
@@ -33,7 +33,7 @@
         expect(auth_result.code).to eq 200
       end
       it 'returns authorization token' do
-        expect(JSON.parse(auth_result)['token']).to eq '12345'
+        expect(JSON.parse(auth_result)['token']).not_to be_nil
```
joaopapereira: I do not understand this change, because the token should be 12345 because it is what you stubbed it with

patmbolger: This code is what I had before, and now I remember why .... if we are running tests without stubbing the service calls the test will fail.

joaopapereira: But didn't we agree that for now we would be stubbing everything?

patmbolger: yes, everything is stubbed but it is useful to turn off stubbing for a test in order to confirm that it is interacting with the *service* correctly.  (otherwise, stubbing is useful for testing behavior within the PETS code but not for confirming correct external interaction nor correct behavior expectations from the RestClient middleware)

joaopapereira: If you are stubbing the socket then you are testing the middleware.
I get what you say about testing the service, nevertheless I believe that unit tests are unit tests and should never touch the outside world, for that we have Integration Tests

joaopapereira:
Nevertheless I think we can live with this in the meanwhile, and when we can have some real IT that got all the way through we can revisit this stubbing

joaopapereira:
Going to merge this PR

PR #242, opened 2016-04-15
Closes AgileVentures/MetPlus_tracker#120

patmbolger:
Simple, initial implementation of résumé upload upon job seeker registration.

This PR does not:

1. Cache résumé file for later upload if cruncher service is unavailable
2. Address résumé change upon job seeker profile edit
3. Accommodate more than one résumé per job seeker.

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -21,6 +21,7 @@ class JobSeekersController < ApplicationController
         unless resume.save
           models_saved = false
           @jobseeker.destroy
+          @jobseeker.errors.messages.merge! resume.errors.messages
         end
       end
     end
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -5,8 +5,27 @@ class JobSeekersController < ApplicationController
   end
 
   def create
+    jobseeker_params = form_params
+    dispatch_file    = jobseeker_params.delete 'resume'
+
     @jobseeker = JobSeeker.new(jobseeker_params)
-    if @jobseeker.save
+    models_saved = @jobseeker.save
+
+    if models_saved
+      if dispatch_file          # If there is a résumé, try to save that
+        tempfile = dispatch_file.tempfile
+        filename = dispatch_file.original_filename
+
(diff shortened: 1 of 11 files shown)


app/models/resume.rb
```diff
@@ -5,9 +5,24 @@ class Resume < ActiveRecord::Base
 
   validates_presence_of :file, on: :create
   validates_presence_of :file_name, :job_seeker_id
+  validate :acceptable_file_type
 
-  # Valid file types for résumé files:
-  FILETYPES = ['pdf', 'doc', 'docx']
+  def acceptable_file_type
```
joaopapereira: eventually we might want to send this to a validator class

app/controllers/job_seekers_controller.rb
```diff
@@ -5,16 +5,35 @@ def new
   end
 
   def create
+    jobseeker_params = form_params
+    dispatch_file    = jobseeker_params.delete 'resume'
+
     @jobseeker = JobSeeker.new(jobseeker_params)
-    if @jobseeker.save
+    models_saved = @jobseeker.save
```
joaopapereira: we save the user information even if we cannot save the resume. Is this behavior as expected? or should we just save the information if we can save the resume?
another thing that pops out is if we cannot save the resume we display a error message but we saved the user info. shouldn't we say "User profile saved, but was unable to upload the resume"?

joaopapereira:

app/controllers/job_seekers_controller.rb
patmbolger: Actually, I do save the JS but, if the resume fails to save, I destroy the JS record.  (I save the JS first because I need the JS id for the resume record).

I chose to have all saves succeed rather than allow the JS save to succeed and resume save to fail.  I think its cleaner that way and it is good to get as much valid info as we can while we have the job seeker's attention.  Also, if the user does not try to upload a resume there is no validation failure due to that and the JS / User gets saved if valid.

joaopapereira: Your are write, didn't saw the destroy. Looks good then

joaopapereira:
Looks good, going to merge

PR #243, opened 2016-04-16
Closes chore for cucumber tests on JobSeeker homepage (#255)

sidhene:


sidhene:
oops, wrong branch. Any way to make "development" the default btw?

PR #245, opened 2016-04-16
Closes chore for cucumber tests on JobSeeker homepage (#255)

sidhene:


features/job_seeker.feature
```diff
@@ -34,6 +34,16 @@ Scenario: login action as jobseeker
   And I login as "vijaya.karumudi@gmail.com" with password "password"
   Then I should see "Signed in successfully."
   And I should be on the Job Seeker 'vijaya.karumudi@gmail.com' Home page
```
patmbolger: This step is defined in navigation_steps.rb:
```
Then(/^I should be on the (.+) page$/) do |page|
  expect(current_path).to eq(page_translator(page))
end
```
In method *page_translator* (in the same file), there is this *when* clause in the case statement:
```
when /Job Seeker '.+' Home/
      user = name.match(/'(.+)'/)
      return home_job_seeker_path(User.find_by_email(user[1]))
```
This is sending a User instance ID to the path helper methods, and should be sending a JobSeeker ID.

So, need to call *pets_user* on the User instance to get the JobSeeker instance associated with the User instance (see the clause for Company Person home page in the same method).

(the test is not failing because there are is only one  User and one JobSeeker in the test setup, and they both would have ID == 1.)

patmbolger:
Reviewed PR ... looks good except for one step definition to be fixed.

features/job_seeker.feature
sidhene: so, in other words 
features/step_definitions/navigation_steps.rb:15 
should say "(User.find_by_email(user[1]).pets_user)"
like features/step_definitions/navigation_steps.rb:18
?

Assuming that is true, changed and pushed :)

joaopapereira:
Looks good, after the conflicts are solved I will merge the PR

sidhene:
I wasn't able to figure out what the conflicts are. Is this a github site
thing or do I need to try merging locally?

joaopapereira:
You need to merge locally, get your branch do a merge from development and push again

sidhene:
Ok, done.

spec/controllers/concerns/user_parameters_spec.rb
```diff
@@ -1,37 +1,37 @@
-require 'rails_helper'
```
joaopapereira: I believe this part of the code should be kept

sidhene: Done

joaopapereira:
Looks good I will merge after freeze

PR #247, opened 2016-04-17

amba178:
No restriction on job_developer for now.


app/controllers/jobs_controller.rb
```diff
@@ -50,25 +55,42 @@ def destroy
...
 
-		def company_person?
-			if pets_user.is_a?(CompanyPerson)
-				set_company_person
+	
+		def company_p_or_job_d?
+			if pets_user.is_a?(CompanyPerson) 
+				set_company_p_or_job_d
 				return true
+			elsif pets_user.is_a?(AgencyPerson) && pets_user.is_job_developer?(pets_user.agency)
```
joaopapereira: You don't really need to check if the pets_user is a AgencyPerson here

app/controllers/jobs_controller.rb
```diff
@@ -50,25 +55,42 @@ def destroy
...
+				set_company_p_or_job_d
+				return true 
 			else
 				return false
 			end
 		end
 
+		def is_right_company_person
+			if @cp_or_jd.is_a?(CompanyPerson)
+				if !(@cp_or_jd.address==@job.address)
```
joaopapereira: as we talked before not 100% sure if this step is needed

amba178:
I will fixed the error...I know what is causing it.

app/models/address.rb
```diff
@@ -1,6 +1,7 @@
 class Address < ActiveRecord::Base
   belongs_to :location, polymorphic: true
-  
+  has_many :jobs
```
joaopapereira: do we really need the connection back the address with Jobs? are we going to use this?

app/models/address.rb
```diff
@@ -9,7 +10,10 @@ class Address < ActiveRecord::Base
     message: 'should be in form of 12345 or 12345-1234',
     allow_blank: true
 
-
+  def full_address
```
joaopapereira: I think we have this function in one helper, do we really need it here?

app/views/layouts/_footer.html.haml
```diff
@@ -9,10 +9,17 @@
 		%div{class: "row text-center", id: "footer-text-color"}
 			-if pets_user.is_a?(CompanyPerson)
 				For employers:
-				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id))
+				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id), company_person_id: current_user.try(:actable_id))
 				|
 				=link_to "Search Resumes", '#'
 				|
 				=link_to "Partner with Us", new_company_registration_path
```
joaopapereira: This doesn't make sense when you already are registered as a company

app/views/layouts/_footer.html.haml
```diff
@@ -9,10 +9,17 @@
...
 				=link_to "Search Resumes", '#'
 				|
 				=link_to "Partner with Us", new_company_registration_path 
+			-if pets_user.is_a?(AgencyPerson) && pets_user.is_job_developer?(pets_user.agency)
+				For employers:
+				=link_to "Post jobs", new_job_path
+				|
+				=link_to "Search Resumes", '#'
+				|
+				=link_to "Partner with Us", new_company_registration_path
```
joaopapereira: As an agency person this link doesn't make sense

features/step_definitions/devise_steps.rb
```diff
@@ -16,8 +16,16 @@
 
 
 Given(/^I am logged in as company person$/) do
-	company_person = FactoryGirl.create(:company_person)
+	FactoryGirl.create(:company_person)
```
joaopapereira: Is this correctly indented? In this diff looks strange

app/controllers/jobs_controller.rb
amba178: Yes, you are right. thanks.

app/controllers/jobs_controller.rb
amba178: Are we assuming that any company person can edit any job at any location? if that is the case than we do not need that step.

app/models/address.rb
amba178: Bi-directional, is there need to know  job created at specific location, maybe?

app/models/address.rb
amba178: I think so. Even though look similar, this operate on the instance of address and the other one I guess you pass the address to it.

app/views/layouts/_footer.html.haml
amba178: Maybe I am missing your point... to register a company at least you have to be company person or someone from agency.

app/views/layouts/_footer.html.haml
amba178: if it is just agency person, yes. But here he/she is agency person and has a role of job developer. What do you think?

features/step_definitions/devise_steps.rb
amba178: I think it is correct. I removed the assignment part.

app/controllers/jobs_controller.rb
patmbolger: I entered a comment about this is a prior PR (or maybe this one?) - can't find that now.

In any case, we have no business requirement to control how a company manages their job postings.  And, as I said in prior comment, the company person is likely to be a person in HQ (e.g. an HR employee) posting jobs for various location of the company.   Limiting job posting like this would cripple that mode of working.

app/models/address.rb
patmbolger: I think it might be useful to list all jobs at a specific company location.

In any case, is there any downside to keeping this 'has_many' even if we don't use it?

app/views/layouts/_footer.html.haml
patmbolger: Actually, *anyone* can register (or, really, request registration of) a company.  You don't need a PETS account to do so.

The person who registers a company is added as the "default" company admin for the company.

That person is not able to do anything in PETS until and if the registration is accepted by the agency.

patmbolger: Anyway, I think João's point here is that the menu item 'Partner with us' should not appear if the logged in user is a person from a registered company.

Similar for the comment immediately below.

app/models/address.rb
joaopapereira: Eventually not, because we are not eagerly retrieving the object, but nevertheless doesn't look correct to have this in the Address.
Lets imagine the address of JobSeeker, does it make sense to have these 2 fields here?
I don't like it , but is more a conceptual issue from my part because it doesn't really affect the code itself

app/views/layouts/_footer.html.haml
amba178: Thanks, for point out....I totally forgot that part, 'Partner with us'. Definitely  right. thxs.

app/views/layouts/_footer.html.haml
amba178: I got your point.

app/models/address.rb
amba178: If we think, we are not going to retrieving it at all than it is better to remove it to avoid confusion in future.

amba178:
I have updated the PR with relevant comments.

joaopapereira:
I think it looks good, need an extra set of eyes to make sure it is ok

db/seeds.rb
```diff
@@ -114,8 +114,8 @@
                       first_name: first_name,
                        last_name: last_name,
                     confirmed_at: confirmed_at,
-                      company_id: companies.pop.id,
-                      address_id: addresses.pop.id,
+                      company_id: companies[n].id,
+                      address_id: addresses[n].id,
```
patmbolger: I like this change a lot .... it makes the seeds processing much faster.

db/seeds.rb
```diff
@@ -214,16 +214,15 @@
   end
 
   branch = Branch.create(code: '001', agency: agency)
-
-  branch.address = Address.create!(city: 'Detroit', state: 'Michigan',
+  branch.address = Address.create!(city: 'Detroit', state: "MI",
```
patmbolger: We're actually storing the long version of state names in the DB, not the abbreviation.  

I don't think it affects our logic, but for consistency you may want to use long names for states.

patmbolger:
I reviewed a second time - looks good.

joaopapereira:
@amba178 were you able to solve the comments?

amba178:
All needed changes are done. Yes.

amba178:
All needed changes are done. Yes.

db/migrate/20151107120000_create_database.rb
```diff
@@ -89,6 +88,7 @@ def change
     create_table :jobs do |t|
       t.string :title
       t.string :description
+      t.integer :address_id, index: true
```
joaopapereira: Please create a new migration for this

spec/controllers/jobs_controller_spec.rb
```diff
@@ -300,9 +302,166 @@
...
+      
+      post :create, :job => {:title => "Ruby on Rails", 
+                             :fulltime => true,
+                             description: "passionate", 
+                             :shift => "Evening",
+                             :company_id => 1,
+                             :address_id => 1, 
+                             company_person_id: nil, 
+                             company_job_id: "WERRR123"}
+      # byebug
```
joaopapereira: Remove this comment

app/controllers/jobs_controller.rb
joaopapereira: You forgot to remove this check

app/views/layouts/_footer.html.haml
```diff
@@ -7,12 +7,20 @@
...
 			-if pets_user.is_a?(CompanyPerson)
-				For employers:
-				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id))
+				
+				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id), company_person_id: current_user.try(:actable_id))
 				|
 				=link_to "Search Resumes", '#'
 				|
-				=link_to "Partner with Us", new_company_registration_path 
+			-elsif pets_user.is_a?(AgencyPerson) && pets_user.is_job_developer?(pets_user.agency)
```
joaopapereira: You do not need the check for agency person

joaopapereira:
Please review my comments. If you do it today I will merge in the morning

app/views/layouts/_footer.html.haml
amba178: I think, I do. without it @ jobseeker will throw noMethodError if I do not put pets_user.is_a?.  is_job_developer? instance method does not handle error, other comparing

amba178: I think I do. otherwise will throw, noMethodError.

PR #250, opened 2016-04-19
Closes AgileVentures/MetPlus_tracker#280

patmbolger:


app/controllers/job_seekers_controller.rb
```diff
@@ -33,26 +33,53 @@ def create
...
       redirect_to root_path
     else
-      @model_errors = @jobseeker.errors
       render 'new'
     end
   end
 
   def edit
     @jobseeker = JobSeeker.find(params[:id])
+    @current_resume = Resume.find_by_job_seeker_id(@jobseeker.id)
```
joaopapereira: The job seeker object format have access to this information?

joaopapereira:
Generally looks good, check my comment
Looking at the controller we are starting to have a lot of logic there maybe we will need to rethink the approach and move the creation and update of the resume to the save of the job seeker model. Not saying note but in the future as an update

app/controllers/job_seekers_controller.rb
patmbolger: Do you mean could I have done this instead?
```
@current_resume = @jobseeker.resumes[0]
```
then, yes, I could have.  Do you think that is a cleaner way to do it?  I am neutral on that.

patmbolger:
I agree that the controller is looking kind of fat.  

Perhaps this is an argument for a model-level AS::Concern.  In fact, we may be able to reuse much of the same logic when we work on sending a Job to the cruncher (which in my mind right now will use a JobCruncher model as intermediary between Job model and CruncherService).

Something to work out later.

app/controllers/job_seekers_controller.rb
joaopapereira: I think so because it hides the need of finding it under the hood of the JobSeeker

patmbolger: OK, I made that change.

joaopapereira:
Looks good going to merge

PR #251, opened 2016-04-20
Closes AgileVentures/MetPlus_tracker#286

joaopapereira:
Added a new javascript module that allow us to convert all the pagination using it to retrieve the using Ajax all the pages


app/assets/javascripts/jobs.js
```diff
@@ -0,0 +1,3 @@
+$(document).ready(function () {
+    PaginationManager("/jobs/list/", '.jobs-view', 'job-type');
+});
```
patmbolger: This could be renamed to 'pagination_setup.js', and we could add all pagination setup calls for each collection in this one file.

app/assets/javascripts/pagination.js
```diff
@@ -0,0 +1,83 @@
...
+                if( typeof successCallback == "function" )
+                    successCallback();
+            },
+            error: function (xhrObj) {
+                Notification.error_notification(xhrObj.responseJSON['message']);
+                if( typeof errorCallback == "function" )
+                    errorCallback();
+            }
+        });
+    };
```
patmbolger: This function is not filtering out clicks on 'Next' and 'Previous' links when the current page is the last page and the first page, respectively.  Even though bootstrap indicates with a hover icon that those links cannot be clicked, the click event still fires and this function is invoked.

The result is that an empty div is returned from the controller.

In ManageData.update_paginate_data I handled it this way:

```
// Check if this anchor element is disabled (e.g. 'Previous'
// link is disabled when on page 1 of pagination)
// (anchor is contained in 'li' element which will have 'disabled' class)

if ($(this).parent().hasClass('disabled')) { return false; };
```

app/controllers/jobs_controller.rb
```diff
@@ -47,6 +51,15 @@ def destroy
 		redirect_to jobs_url 
 	end
 
+	def list
+    @job_type = params[:job_type] || 'my-company-all'
+    raise 'Unsupported request' if not request.xhr?
```
patmbolger: Should this statement occur before any assignment statements?

patmbolger:
I completed an initial review - please see in-line comments.

Also, the visual behavior of the pagination is very "jumpy" and does not provide a smooth user experience.

Compare to the pagination for the agency admin pages.

To help with this, I added to seeds.rb: created 50 jobs for the same company, and created a 'known' company person for that company, so that we have some jobs to paginate through in the company person home page.  The person login credentials are:

email: hr@widgets.com

password: qwerty123

You can grab that seed file from my repo, branch:  joao-feature/286

joaopapereira:
Going to add a spinner to remove the jumpiness
Add the seeds file change
Change the instantiation to be more intuitive
Move the assignment

joaopapereira:
@patmbolger Done the tasks, ready for new review

app/assets/javascripts/pagination.js
```diff
@@ -0,0 +1,221 @@
...
+ * The list can be retrieve using AJAX from the url: /potatos/list/all-my-potatoes
+ *
+ * The implementation:
+ * Add the following HTML code to your page
+ * <div id='potato-1' class='pagination-div' data-url='/potatos/list/' data-potato-pagination-type='all-my-potatoes'>
+ * </div>
+ *
+ * Remark, the ID of the div should never be the same, to make sure everything works as expected.
+ *
+ * If you want to setup a specific call back of success in the previous example you should do:
```
patmbolger: " ..... callback for success or error ...."

app/assets/javascripts/pagination.js
```diff
@@ -0,0 +1,221 @@
...
+        this.start = function() {
+            target.addClass('opaque');
+            self._spinner.spin(target[0]);
+        };
+        this.stop = function() {
+            target.removeClass('opaque');
+            self._spinner.stop();
+        };
+        return this;
+    };
```
patmbolger: The spinner stuff would be very useful if available in a generic set of functions - they can be used, for instance, during job seeker registration, when 1) JS is validated and saved, 2) résumé is validated and saves, and 3), résumé file is uploaded to cruncher.

app/assets/javascripts/pagination.js
```diff
@@ -0,0 +1,221 @@
...
+ * The implementation:
+ * Add the following HTML code to your page
+ * <div id='potato-1' class='potato-view'>
+ * </div>
+ *
+ * Add one of the following options to your JS file
+ *   Select by id
+ *     PaginationHandler('/potatos/list/potato-pagination-type', '#potato-1')
+ *
+ *   Select by class if we have more then 1
```
patmbolger: Select by class if we have more than one pagination section on the page for the same collection
(e.g. a view showing "Job Seekers without a Job Developer" and "Job Seekers without a Case Manager")

patmbolger:
Done with second review.

PR #252, opened 2016-04-20
Closes AgileVentures/MetPlus_tracker#279

joaopapereira:
Apparently it was works before, maybe something in the environment caused the failure.
Reactivated the test


patmbolger:
Reviewed PR .... looks good (and its good to see that the test actually works without change)

PR #253, opened 2016-04-21

amba178:
added column in the jobs tables, address_id using migration generator


app/views/layouts/_footer.html.haml
```diff
@@ -7,12 +7,20 @@
...
 			-if pets_user.is_a?(CompanyPerson)
-				For employers:
-				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id))
+				
+				=link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id), company_person_id: current_user.try(:actable_id))
 				|
 				=link_to "Search Resumes", '#'
 				|
-				=link_to "Partner with Us", new_company_registration_path 
+			-elsif pets_user.try(:agency)  && pets_user.is_job_developer?(pets_user.try(:agency))
```
amba178: This is the best I can do without checking for agency_person. Let me know if you have any trick for using ONLY THIS PART "pets_user.is_job_developer?"

joaopapereira: Ok I now understand you problem.
We need to create a new method called: pets_agency that will return us the current agency instead of relying on the pets_user.
I am ok with this for now, I will create a story for this

joaopapereira:
Ok looks, good. Going to merge

PR #256, opened 2016-04-21
Closes Agency Person homepage (Case Manager #230)

sidhene:


sidhene:
The failing test is the one I got help on in PL today, but apparently I did not understand the answer.

spec/controllers/people_invitations_controller_spec.rb
```diff
@@ -60,7 +60,7 @@
       end
       it 'redirects as specified' do
         expect(response).
-            to render_template(@controller.after_invite_path_for(agency_admin))
+            to render_template(devise/mailer/invitation_instructions)
```
patmbolger: the argument to render_template should be a string (see line 205 in same file)

patmbolger:
You should rebase against upstream development and then push to your repo again .... your PR includes 140 files!

sidhene:
@patmbolger - I'm going to need to pair on the rebase: http://stackoverflow.com/questions/36638043/i-cant-believe-git-rebase-is-supposed-to-work-this-way

spec/controllers/people_invitations_controller_spec.rb
sidhene: Ok Thanks.

PR #259, opened 2016-04-22
Closes AgileVentures/MetPlus_tracker#230 - Case Manager Home Page

sidhene:

PR #261, opened 2016-04-22
Closes AgileVentures/MetPlus_tracker#231 - job seekers associated w/ CM

sidhene:
@joaopapereira  - since most of the hard part was done in JDhome#226, the challenge here was mainly parsing out the rspec tests to handle different agency person role scenarios. Producing lists of associated and "orphan" JS was trivial, so they are both in this story. 

The next hard chunk is going to be breaking out the lists in the home view into partials and calling them successfully. So #234 will be about this work. I've updated the two stories accordingly.


sidhene:
Failing Scenarios:
cucumber features/jobs.feature:70 # Scenario: Creating, Updating, and Deleting Job successfully and unsuccessfully
cucumber features/people_invite.feature:38 # Scenario: invite (and reinvite) new agency person
cucumber features/people_invite.feature:85 # Scenario: invite (and reinvite) new Company person

These also fail locally on cucumber but work manually. I am not sure what's wrong. I'm on Ruby 2.2.2 now and many of the problems from the latest build have disappeared.

app/views/agency_people/home.html.haml
```diff
@@ -1,7 +1,103 @@
-%h3 Your Job Seekers
-- if @your_jobseekers_jd.empty?
-  %i There are no job seekers assigned to you yet.
-- else
+%h2
+  = "Welcome back to PETS, #{pets_user.full_name}"
```
patmbolger: Recommend changing from
```
pets_user.full_name
```
to
```
pets_user.full_name(last_name_first: false)
```
So instead of seeing this:

"Welcome back to PETS, Smith, John"

the user would see:

"Welcome back to PETS, John Smith"

app/views/agency_people/home.html.haml
```diff
@@ -1,7 +1,103 @@
...
+                  = js.job_seeker_status.value
+              %td
+                - if js.last_sign_in_at.nil?
+                  %i never
+                - else
+                  = js.last_sign_in_at
+      = will_paginate @your_jobseekers_cm, param_name: 'your_jobseekers_cm_page'
+    .clearfix
+
+    %h3 Job Seekers Without a Case Manager
```
patmbolger: Maybe add logic to handle the case where there are *no* job seekers without a CM.

app/views/agency_people/home.html.haml
```diff
@@ -1,7 +1,103 @@
...
+                  = js.job_seeker_status.value
+              %td
+                - if js.last_sign_in_at.nil?
+                  %i never
+                - else
+                  = js.last_sign_in_at
+      = will_paginate @your_jobseekers_jd, param_name: 'your_jobseekers_jd_page'
+    .clearfix
+
+  %h3 Job Seekers Without a Job Developer
```
patmbolger: Maybe add logic to handle the case where there are *no* job seekers without a JD.

spec/models/job_seeker_spec.rb
```diff
@@ -57,6 +57,9 @@
       adam.assign_case_manager(cm_person, agency)
       dave.assign_job_developer(jd_person, agency)
     end
+    # describe '#self.js_without_jd' do
+    #   expect(JobSeeker.js_without_jd).to include(adam)
+    # end
```
patmbolger: Not sure why this is commented out?

In any case, I recommend adding tests for these methods:

js_without_jd
js_without_cm
your_jobseekers_jd
your_jobseekers_cm

Also, to be consistent with naming it would be reasonable to name the first two methods:

jobseekers_without_jd
jobseekers_without_cm

patmbolger:
Reviewed PR with some comments.

joaopapereira:
@sidhene Looking at the tests the problem is one that we already saw before, you need to create the Agency Roles of CM and JD to make sure the tests pass.
This issue is included by the in_role_of that assumes that the Roles are all created, and in the tests they are not

features/company_registration.feature
```diff
@@ -10,6 +10,8 @@ Background: seed data added to database
   Given the following agency roles exist:
   | role  |
   | AA    |
+  | JD    |
+  | CM |
```
joaopapereira: Is the spacing correct here?

joaopapereira:
This PR is overlapping the #259 so I suggest that you merge the Comments on both of them and see wish ones make sense to be implemented in each PR.
Ex: the unit tests that @patmbolger recommends some should be in the PR #259 because that is the place where you add the functions

joaopapereira:
I will review again this PR again when we merged the #259

app/views/agency_people/home.html.haml
sidhene: Done

app/views/agency_people/home.html.haml
sidhene: I think I have that in another story. Joao wanted me to deliver just the home page.

sidhene: done

app/views/agency_people/home.html.haml
sidhene: done

spec/models/job_seeker_spec.rb
sidhene: js_without_jd
/spec/controllers/agency_people_controller_spec.rb:44

js_without_cm
/spec/controllers/agency_people_controller_spec.rb:60

your_jobseekers_jd
/spec/controllers/agency_people_controller_spec.rb:40

your_jobseekers_cm
/spec/controllers/agency_people_controller_spec.rb:56

features/company_registration.feature
sidhene: done

sidhene:
Added "WIP" because I've found an edge case that I want to make a new test for before merging. However, I need to bounce it off semaphore first because of a couple of tests that only fail on my setup.

app/models/job_seeker.rb
```diff
@@ -67,10 +67,10 @@ def assign_agency_person(agency_person, role_key)
       # Is this role assigned already to an agency person?
 
       # If so, is this the same agency person? - then we're done
-      return if ap_relation.agency_person == agency_person
+      return if ap_relation == agency_person
```
joaopapereira: This was not working before?

spec/controllers/concerns/user_parameters_spec.rb
```diff
@@ -7,6 +7,7 @@ class TestClass
 RSpec.describe UserParameters do
 
   describe '#handle_user_form_parameters' do
+
```
joaopapereira: Is this change needs ?

app/controllers/agency_people_controller.rb
```diff
@@ -14,7 +14,9 @@ def home
     @agency_person = AgencyPerson.find(params[:id])
     @agency = @agency_person.agency
     @js_without_jd = JobSeeker.paginate(:page=> params[:js_without_jd_page], :per_page=>5).js_without_jd
-    @your_jobseekers_jd = JobSeeker.paginate(:page=> params[:your_jobseekers_jd], :per_page=> 5).your_jobseekers_jd(@agency_person)
+    @js_without_cm = JobSeeker.paginate(:page=> params[:js_without_cm_page], :per_page=>5).js_without_cm
+    @your_jobseekers_jd = JobSeeker.paginate(:page=> params[:your_jobseekers_jd_page], :per_page=> 5).your_jobseekers_jd(@agency_person)
+    @your_jobseekers_cm = JobSeeker.paginate(:page=> params[:your_jobseekers_cm_page], :per_page=> 5).your_jobseekers_cm(@agency_person)
```
joaopapereira: Really not a fan of these function names
JobSeeker.with_job_developer @agencyperson
Looks better
Were they created for this story?

joaopapereira:
Done WIP review

app/models/job_seeker.rb
```diff
@@ -67,10 +67,10 @@ def assign_agency_person(agency_person, role_key)
...
 
       # Otherwise, reassign case manager role for this job seeker
-      ap_relation.agency_person = agency_person
+      ap_relation = agency_person
```
patmbolger: The logic for "assign_agency_person" has a bug related to reassigning an AgencyPerson in the specific role (JD or CM).  (it seems to have been there awhile, before this PR).

The method *find_agency_person(role_key)* returns an AP instance, not an AgencyRelation (AR) instance.

However, lines 73 - 74 seem to want to treat the return value as an AR:

```
# Otherwise, reassign case manager role for this job seeker
ap_relation = agency_person
ap_relation.save
```
(if *ap_relation* actually *was* an AR, the correct first statement above would be ap_relation.agency_person = agency_person).

This code was associated with prior logic that was refactored.  The var *ap_relation* (in prior logic) was an AR that associated the job seeker with an AP in the specific role.  Thus, assigning another AP person to that relationship, and saving it, reassigned the JD or CM role, for the job seeker, to another AP.

In this version of the code, *ap_relation* is just being assigned to an AP, and the AP is then saved, with no effect on the existing JB <> AP relationship.

This can be tested by changes to job_seeker_spec.rb, specifically 1) add a 2nd case manager (cm_person2) to the test setup, and 2) adding this test:
```
describe 'reassign case manager' do
  it 'replaces current case manager' do
    adam.assign_case_manager(cm_person2, agency)
    expect(adam.case_manager).to eq(cm_person2)
  end
end
```
(*adam* had been assigned case manager *cm_person* in the test setup).

The reassign fails with this:
```
Failure/Error: expect(adam.case_manager).to eq(cm_person2)
     
       expected: #<AgencyPerson id: 2, agency_id: 1, created_at: "2016-04-27 10:29:47", updated_at: "2016-04-27 10:29:47", branch_id: 2, status: "Active">
            got: #<AgencyPerson id: 1, agency_id: 1, created_at: "2016-04-27 10:29:47", updated_at: "2016-04-27 10:29:47", branch_id: 1, status: "Active">
```

patmbolger:
Completed second review .... please note comment related to bug.

app/models/job_seeker.rb
sidhene: This is the edge case I mentioned. I'll explain in parking lot.

app/controllers/agency_people_controller.rb
sidhene: How badly do you want this? My guess at the cost is "a couple of days + a couple of help sessions (either by parking lot or individual)".

app/models/job_seeker.rb
sidhene: I've inserted the suggested test. It's failing. I'll get help either in parking lot or pairing to fix the JobSeeker#assign_agency_person method behind it.

sidhene: done. pushed.

sidhene:
I'm not sure what to do about the failing cucumber test. In practice, the 'post jobs' link is in the footer of every page, so capybara should have no difficulty finding it. This also fails locally, along with two other tests that do not fail on semaphore. We can talk about it today after standup.

sidhene:
Well...it looks like that worked...

app/models/job_seeker.rb
```diff
@@ -81,11 +81,18 @@ def assign_agency_person(agency_person, role_key)
   end
 
   def find_agency_person(role_key)
+    agency_relation = find_agency_person_relation(role_key)
```
joaopapereira: I would like to discuss this past of the change because now I see multiple functions that do the same thing, or are very similar.
The function to assign the agency person currently looks like to much code

app/views/agency_people/_job_seekers.html.haml
```diff
@@ -0,0 +1,25 @@
+#jobseekers_table
+%h3 This is a partial!
```
joaopapereira: ?

spec/models/job_seeker_spec.rb
```diff
@@ -94,5 +95,15 @@
         expect(adam.job_developer).to eq(nil)
       end
     end
+    describe '#reassign agency person' do
```
joaopapereira: The describe should be "Assign agency person"

joaopapereira:
Looking in retrospective the assignation process looks to cumbersome with to many functions that I am not sure if they bring more value. The assign could use the find_or_create_by function for rails that would give us always a radiation object to work with and we just would need to save it.

joaopapereira:
The PR looks good, just take a look at my comments

app/models/job_seeker.rb
sidhene: Ok, but rather than expanding the story, I'd prefer to have refactoring as a separate story.
I've found that the more things get changed/added, the more muddled everything becomes and the longer it takes.

sidhene: We could also address the variable names at this point. These two changes are a good size for a story, and since they are limited, they're more likely to be produced/tested/merged smoothly.

joaopapereira: Ok, then rollback on this file, create 2 stories 1 to solve this corner case and refactor the assign functions, and another one to rename the other functions

app/views/agency_people/_job_seekers.html.haml
sidhene: That's the beginning of #234, verifying that the partial loads correctly.

app/models/job_seeker.rb
sidhene: Sorry I don't understand. We can discuss in PL.

app/views/agency_people/_job_seekers.html.haml
joaopapereira: But way is this in the current PR?

joaopapereira:
This story is waiting on the creation of the 2 stories missing and understand why we have a partial that we are not using

app/views/agency_people/_job_seekers.html.haml
sidhene: So I wouldn't break anything in the process of removing it

sidhene:
@joaopapereira - one of the stories you requested was to handle an edge case that I mentioned which does not exist anymore (it was corrected as part of this review). I've created one story (#313) for the refactoring. Let me know if you want that broken into two stories.

sidhene:
@joaopapereira - The edge case for one of the stories you asked for no longer exists (it was fixed as part of this pull request). I've created #313 for the refactoring. Let me know if you need that split into 2 stories.

joaopapereira:
Just created the following tasks from this PR:
AgileVentures/MetPlus_tracker#317
AgileVentures/MetPlus_tracker#316
AgileVentures/MetPlus_tracker#315
AgileVentures/MetPlus_tracker#314

This task was also created, but I think we need to split it into parts
AgileVentures/MetPlus_tracker#313

joaopapereira:
Going to merge this story

PR #268, opened 2016-04-28
connects AgileVentures/MetPlus_tracker#213

amba178:
job<>job_category

### DIFF
--- a/app/views/jobs/_form.html.haml
+++ b/app/views/jobs/_form.html.haml
@@ -9,6 +9,10 @@
 	=f.label :title, class: "col-sm-3 control-label text-right"
 	%div{class: "col-sm-6"}
 		=f.text_field :title, class: "form-control" 
+.form-group
+	=f.label  :job_category_id, "Job Category", class: "col-sm-3 control-label text-right"
+	%div{class: "col-sm-6"}
+		=f.collection_select :job_category_id, JobCategory.all, :id, :name, {prompt: true}, {class: "form-control"}
 
 -if @cp_or_jd.is_a?(AgencyPerson) && @cp_or_jd.is_job_developer?(pets_user.agency)
 	.form-group
--- a/app/views/jobs/show.html.haml
+++ b/app/views/jobs/show.html.haml
@@ -30,8 +30,7 @@
 					
 			%tr
 				%td Job category:
-				/ %td=@job.job_category.name 
-				%td Walmart co. placeholder
+				%td=@job.job_category.name 
 			%tr 
 				%td Full time: 
 				%td=@job.fulltime ? "Yes" : "No"
--- a/features/step_definitions/jobs_steps.rb
+++ b/features/step_definitions/jobs_steps.rb
@@ -16,3 +16,7 @@ end
 Given(/^the Widgets, Inc\. company name with address exist in the record$/) do
 	FactoryGirl.create(:company)
 end
+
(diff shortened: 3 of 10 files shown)


app/views/jobs/_form.html.haml
```diff
@@ -9,6 +9,10 @@
 	=f.label :title, class: "col-sm-3 control-label text-right"
 	%div{class: "col-sm-6"}
 		=f.text_field :title, class: "form-control" 
+.form-group
+	=f.label  :job_category_id, "Job Category", class: "col-sm-3 control-label text-right"
+	%div{class: "col-sm-6"}
+		=f.collection_select :job_category_id, JobCategory.all, :id, :name, {prompt: true}, {class: "form-control"}
```
joaopapereira: Is this a multi select? because we need to be able to select multiple categories.
The idea I had for the categories was something inspired on the stackoverflow Tags, do you know what I am talking about? When you create a question you can Tags it
tags_stackoverflow

app/views/jobs/show.html.haml
```diff
@@ -30,8 +30,7 @@
 					
 			%tr
 				%td Job category:
-				/ %td=@job.job_category.name 
-				%td Walmart co. placeholder
+				%td=@job.job_category.name
```
joaopapereira: We will have multiple categories in one job

features/step_definitions/jobs_steps.rb
```diff
@@ -16,3 +16,7 @@
 Given(/^the Widgets, Inc\. company name with address exist in the record$/) do
 	FactoryGirl.create(:company)
 end
+
+Given(/^the Software Engineer\-RoR job category exist in the test db$/) do
```
joaopapereira: The category name should be variable to make it more flexible if we need

app/views/jobs/show.html.haml
patmbolger: We should discuss at scrum.  Our model has been job 'belongs_to' category for some time.

Do we need multiple categories for a job?

We need to come to a better understanding of the business model behind our assumptions that drive the model.

IMO, if we have job *categories*, then a job should belong to one category (since category is a classification of job type).

However, perhaps a better model of to have *skills categories*.  That way, a job has many categories, and the skills required for any particular job can be built up from those categories.

This latter model might be more useful.  For example, a specific job for a SW developer can be assembled to include categories (and associated skills for) 'Javascript Development', 'RoR Development', 'DB modeling', 'Scrum Master', etc.

The more I think about it, it seems that SkillsCategory might be a more useful model than JobCategory ....

spec/factories/jobs.rb
```diff
@@ -5,7 +5,7 @@
     company_id 1
     company_person_id 1
     address_id 1
-    # job_category_id 1
+    job_category_id 1
```
joaopapereira: Why dont you use the factory?

app/views/jobs/show.html.haml
joaopapereira: I agree with you. The categories just help on grouping things up.

app/views/jobs/_form.html.haml
amba178: We should change the association than...

patmbolger:
PR reviewed .... please see comments.

We have to decide 1) if a job can be associated with many categories, and 2) what *kind* of categories (job of skills)?

joaopapereira:
We need to decide what to do with the Skills. After that we will look into this PR

PR #269, opened 2016-04-28
Closes AgileVentures/MetPlus_tracker#267

patmbolger:


joaopapereira:
The change looks good.
Good job on the dragable part
Let's keep this PR waiting until we can talk what to do about the categories

PR #270, opened 2016-04-28
Closes AgileVentures/MetPlus_tracker#275

joaopapereira:


db/seeds.rb
```diff
@@ -13,18 +13,14 @@ def create_address(location = nil)
...
-                                              description: "A jobseeker with a job and not looking
-                        for a job for now.")
+JobSeekerStatus.all_valid.each do |status|
+  case status.key
+    when JobSeeker::STATUS[:unemployed_seeking]
+      @jss1 = status
+    when JobSeeker::STATUS[:unemployed_seeking]
+      @jss2 = status
+    when JobSeeker::STATUS[:unemployed_seeking]
+      @jss3 = status
```
patmbolger: I think you wanted jss2 and jss3 to be other status value than :unemployed_looking ...

app/models/job_seeker.rb
```diff
@@ -1,20 +1,39 @@
...
   validates  :year_of_birth, :year_of_birth => true
 
 
+  STATUS = {:unemployed_seeking   => 'UNEMPLOYEDLOOKING',
+            :employed_seeking     => 'EMPLOYEDLOOKING',
+            :employed_not_seeking => 'EMPLOYEDNOTLOOKING'}
+
+  validates_presence_of :job_seeker_status
+  validates :job_seeker_status, inclusion: { in: STATUS.values,
+              message: "%{value} is not a valid size" }
```
patmbolger: *message: "%{value} is not a valid status" }*  ??

patmbolger:
Reviewed PR with some comments.

Overall, though, I am not sure of the benefits of having valid status keys defined in a  constant in JobSeeker.  This means that we have to alter the code in order to add/delete/edit JS status values.  That seems to be overkill since our code logic paths do not depend on any specific status values.

I would think that having the status defined solely in a table would be sufficient.  That would also allow the user (eventually) to change status values without our having to change the code.  

What is the rationale for this?

joaopapereira:
This was what we talked about last week to move this into a enum.
The only caveat we have there is if we want to sad need status, in that case we would need to add them to the enum.
The rest of the information is all sorted in the database

patmbolger:
When  we discussed I was assuming an enum field in the DB.

I was thinking along the lines of this: http://guides.rubyonrails.org/4_1_release_notes.html#active-record-enums

However, I now realize that this would not work as we would still need to modify code to edit status values.

At one of the client meetings, Chet was asking about the ability to add status values.  He mentioned "Community Service" as an example.  That makes me think that wants to customize these values (e.g. "In Training").

patmbolger:
PR reviewed - looks good.

PR #271, opened 2016-05-02
Closes AgileVentures/MetPlus_tracker#287

patmbolger:


app/controllers/companies_controller.rb
```diff
@@ -30,6 +32,22 @@ def update
     end
   end
 
+  def list_people
```
joaopapereira: This should be here or in CompanyPeople controller?

patmbolger: I went back and forth on this (your wording in the story made me think you thought it should join the companies controller .... ).

Anyway, I agree that it is better to put this in the company people controller .... done.

joaopapereira:
Going to merge the PR

PR #274, opened 2016-05-07
Closes AgileVentures/MetPlus_tracker#141

patmbolger:
Here is what the revised home page looks like.  Also, I removed the 'New User' link from the menu bar.

screen shot 2016-05-07 at 5 53 39 pm


joaopapereira:
Looks nice, I will merge it and we will see how can we build up on it :D

PR #275, opened 2016-05-08
Closes AgileVentures/MetPlus_tracker#291

patmbolger:


app/views/devise/invitations/new.html.haml
```diff
@@ -1,6 +1,10 @@
 %h2= t "devise.invitations.new.header"
 = form_for resource, :as => resource_name, :url => invitation_path(resource_name), :html => {:method => :post} do |f|
-  = devise_error_messages!
+  -# = devise_error_messages!
```
joaopapereira: We should remove this if we are not going to use it

spec/controllers/agencies_controller_spec.rb
```diff
@@ -65,12 +65,9 @@
         @agency.valid?
         agency_admin = FactoryGirl.create(:agency_admin, agency: @agency)
         sign_in agency_admin
-        patch :update, agency: FactoryGirl.attributes_for(:agency, 
+        patch :update, agency: FactoryGirl.attributes_for(:agency,
                         phone: '', website: 'nodomain'), id: @agency
       end
-      it 'assigns @model_errors for error display in layout' do
```
joaopapereira: Maybe just removing this is not enough, are we checking the error messages are set? Or that we are rendering something?

joaopapereira:
The comment about the specs need to be revisited for all the specs

joaopapereira:

patmbolger:
Changes made in response to comments.

PR #257, opened 2016-04-22
Closes AgileVentures/MetPlus_tracker#276

joaopapereira:

PR #278, opened 2016-05-10
WIP - closes #257

sidhene:
Addition of 'my open tasks' partial to AP homepage w/ cucumber integration tests:
Tasks for 
JD
CM
can move tasks forward
can only see role-specific tasks
will see ‘you have no tasks’ if no tasks

AA
that AA can see all OPEN tasks
can assign tasks
that tasks disappear from AA view when assigned to another AP

Using PR to ping off semaphore


sidhene:
Deleting because placed against wrong branch

PR #279, opened 2016-05-11
Testing against Semaphore

sidhene:
This PR is to test my copy of the development branch against semaphore to see if it fails in the same way.


sidhene:
Does not match upstream/development. Closing PR and deleting branch

PR #281, opened 2016-05-11
Closes AgileVentures/MetPlus_tracker#257

sidhene:
Addition of 'my open tasks' partial to AP homepage w/ cucumber integration tests:
Tasks for 
JD
CM
can move tasks forward
can only see role-specific tasks
will see ‘you have no tasks’ if no tasks

AA
that AA can see all OPEN tasks
can assign tasks
that tasks disappear from AA view when assigned to another AP

Tests improvement:
devise_steps.rb - from hard-coded to object property
jobs.feature - from poltergeist to selenium 
company_registration.feature - corrected misspelling

### DIFF
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -13,11 +13,11 @@ class AgencyPeopleController < ApplicationController
   def home
     @agency_person = AgencyPerson.find(params[:id])
     @agency = @agency_person.agency
+    @task_type = 'mine-open'
     @js_without_jd = JobSeeker.paginate(:page=> params[:js_without_jd_page], :per_page=>5).js_without_jd
     @js_without_cm = JobSeeker.paginate(:page=> params[:js_without_cm_page], :per_page=>5).js_without_cm
     @your_jobseekers_jd = JobSeeker.paginate(:page=> params[:your_jobseekers_jd_page], :per_page=> 5).your_jobseekers_jd(@agency_person)
     @your_jobseekers_cm = JobSeeker.paginate(:page=> params[:your_jobseekers_cm_page], :per_page=> 5).your_jobseekers_cm(@agency_person)
-
   end
 
   def update
--- a/app/views/agency_people/home.html.haml
+++ b/app/views/agency_people/home.html.haml
@@ -5,8 +5,8 @@
     :id => 'edit-profile' do
     %i.fa.fa-pencil-square-o.fa-1
 
-%h3 Tasks that need your attention
-%i Tasks would go here
+%h3 Your Open Tasks
+= render partial: 'tasks/tasks_structure' , :locals => {task_type: @task_type}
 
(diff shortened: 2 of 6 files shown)


joaopapereira:
Please rebase this PR

app/controllers/agency_people_controller.rb
```diff
@@ -67,6 +67,7 @@ def update
 
         @agency_person.job_seekers = []
       end
+      @model_errors = @agency_person.errors
```
patmbolger: We are not using @model_errors anymore to display model validation errors.  (this will not display any errors, in fact, since the code for using @model_errors has been removed).

You need to use the partial app/views/shared/_error_messages.html.haml.

There are lots of examples for that.

app/controllers/agency_people_controller.rb
```diff
@@ -83,6 +84,7 @@ def update_profile
       flash[:notice] = "Your profile was updated successfully."
       redirect_to root_path
     else
+      @model_errors = @agency_person.errors
```
patmbolger: same comment as above

app/views/agency_people/home.html.haml
```diff
@@ -91,7 +92,7 @@
                 = js.full_name
               %td
                 - unless js.job_seeker_status.nil?
-                  = js.status.short_description
+                  = js.job_seeker_status.description
```
patmbolger: may be better to use 'short_description'

patmbolger:

app/controllers/agency_people_controller.rb
sidhene: That shouldn't be in there. It's probably because #update was changed since I started the story. I've replaced this file with a new copy that should only have one line in #home that is different.

app/views/agency_people/home.html.haml
sidhene: done. thanks

features/company_registration.feature
```diff
@@ -81,18 +81,19 @@ Scenario: company registration approval
   When "hughjobs@widgets.com" opens the email with subject "Confirmation instructions"
   Then they should see "You can confirm your account email through the link below:" in the email body
 
-@javascript
+@selenium
```
joaopapereira: Do we need selenium in these tests?

sidhene: Yes. This change converted 4 failing scenarios to passing both in Semaphore and locally. Apparently whenever a scenario includes a pop up window, it needs selenium instead.

sidhene:
Ok, we're done troubleshooting the tests. If there aren't any more change requests I'll do the rebase and last push.

joaopapereira:
I believe you can rebase

sidhene:
rebased and pushed

patmbolger:
Reviewed PR - looks good.

joaopapereira:
Looks good going to merge

PR #282, opened 2016-05-11
Test semaphore

sidhene:
last pull request is failing in a way it should not. This is to verify whether the fail comes from that branch or not.


sidhene:
This is very weird. About a dozen tests failing even though this is identical to the main development branch except 3 characters on a html page.

patmbolger:
This PR is against master.  Try again against development.

PR #284, opened 2016-05-11
Jobscreation cruncher

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#288

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -39,6 +39,7 @@ class Job < ActiveRecord::Base
 
   def last_application_by_job_seeker(job_seeker)
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
+  end
   
   def save!
    super
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -41,8 +41,15 @@ class Job < ActiveRecord::Base
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
   
-  def save!
-   super
-   JobCruncher.create_job(id, title, description)
+  def save
+      return false if not super
+      begin
+        return true if JobCruncher.create_job(id, title, description)
+      rescue 
+        errors.add(:job, 'could not be created -please try again')
+        destroy
+        raise
+      end
+      false
   end
 end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -41,12 +41,13 @@ class Job < ActiveRecord::Base
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
   
-  def save!
+  def self_create()
+      after_create
       return false if not super
       begin
         return true if JobCruncher.create_job(id, title, description)
       rescue 
-        errors.add(:job, 'could not be created -please try again')
+        errors.add(:job, 'could not be created, please try again')
         destroy
         raise
       end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -1,4 +1,5 @@
 class Job < ActiveRecord::Base
+  after_save :save_job_to_cruncher
   belongs_to :company
   belongs_to :company_person
   belongs_to :address
@@ -41,16 +42,17 @@ class Job < ActiveRecord::Base
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
   
-  def self_create()
-      after_create
-      return false if not super
+  
+  def save_job_to_cruncher()
       begin
-        return true if JobCruncher.create_job(id, title, description)
+        if self.id_changed?
+          return true if JobCruncher.create_job(id, title, description)
+        end
+        
       rescue 
         errors.add(:job, 'could not be created, please try again')
-        destroy
-        raise
+        raise ActiveRecord::RecordInvalid.new(self)
       end
-      false
+      
   end
 end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -53,6 +53,5 @@ class Job < ActiveRecord::Base
         errors.add(:job, 'could not be created, please try again')
         raise ActiveRecord::RecordInvalid.new(self)
       end
-      
   end
 end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -39,5 +39,9 @@ class Job < ActiveRecord::Base
 
   def last_application_by_job_seeker(job_seeker)
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
+  
+  def save!
+   super
+   JobCruncher.create_job(id, title, description)
   end
 end
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -41,7 +41,7 @@ class Job < ActiveRecord::Base
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
   
-  def save
+  def save!
       return false if not super
       begin
         return true if JobCruncher.create_job(id, title, description)
--- a/db/migrate/20160501153517_create_job_crunchers.rb
+++ b/db/migrate/20160501153517_create_job_crunchers.rb
@@ -1,0 +1,8 @@
+class CreateJobCrunchers < ActiveRecord::Migration
+  def change
+    create_table :job_crunchers do |t|
+
+      t.timestamps null: false
+    end
+  end
+end
--- a/db/migrate/20160501153517_create_job_crunchers.rb
+++ b/db/migrate/20160501153517_create_job_crunchers.rb
@@ -1,8 +1,0 @@
-class CreateJobCrunchers < ActiveRecord::Migration
-  def change
-    create_table :job_crunchers do |t|
-
-      t.timestamps null: false
-    end
-  end
-end
--- a/spec/factories/job_crunchers.rb
+++ b/spec/factories/job_crunchers.rb
@@ -1,6 +1,0 @@
-FactoryGirl.define do
-  factory :job_cruncher do
-  end
-
-end
-
--- a/spec/factories/job_crunchers.rb
+++ b/spec/factories/job_crunchers.rb
@@ -1,9 +1,5 @@
 FactoryGirl.define do
   factory :job_cruncher do
-   jobId 10
-   title 'Software Engineer'
-   description 'description of the job'
-   
   end
 
 end
--- a/spec/factories/job_crunchers.rb
+++ b/spec/factories/job_crunchers.rb
@@ -1,0 +1,13 @@
+FactoryGirl.define do
+  factory :job_cruncher do
+   jobId 10
+   title 'Software Engineer'
+   description 'description of the job'
+   
+  end
+
+end
+
+
+
+
--- a/spec/factories/job_crunchers.rb
+++ b/spec/factories/job_crunchers.rb
@@ -4,6 +4,3 @@ FactoryGirl.define do
 
 end
 
-
-
-
--- a/spec/models/job_cruncher_spec.rb
+++ b/spec/models/job_cruncher_spec.rb
@@ -1,5 +1,0 @@
-require 'rails_helper'
-
-RSpec.describe JobCruncher, type: :model do
-  pending "add some examples to (or delete) #{__FILE__}"
-end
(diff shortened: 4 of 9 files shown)


vijayakarumudi:
PR for review

app/models/job.rb
```diff
@@ -40,4 +40,9 @@ def apply job_seeker
   def last_application_by_job_seeker(job_seeker)
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
+  
+  def save!
+   super
+   JobCruncher.create_job(id, title, description)
+  end
```
patmbolger: This will raise an (unhanded) exception even for simple model validation errors, which is not what we want.

Perhaps something like this would work better:
```
def save
    return false if not super
    begin             # successful model save, now try to create job in Cruncher
      return true if JobCruncher.create_job(id, title, description)
    rescue
      errors.add(:job, 'could not be created - please try again')
      destroy         # destroy the saved job instance in DB
      raise           # raise the exception for handling further up the stack
    end
    false
  end
```

db/migrate/20160501153517_create_job_crunchers.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJobCrunchers < ActiveRecord::Migration
+  def change
+    create_table :job_crunchers do |t|
+
+      t.timestamps null: false
+    end
+  end
+end
```
patmbolger: The JobCruncher class is not an ActiveRecord subclass, and is not persisted.

This migration file is not needed.

spec/factories/job_crunchers.rb
```diff
@@ -0,0 +1,6 @@
+FactoryGirl.define do
+  factory :job_cruncher do
+  end
+
+end
+
```
patmbolger: The JobCruncher class is not an ActiveRecord subclass, and is not persisted.

This factory is not needed.

spec/models/job_cruncher_spec.rb
```diff
@@ -0,0 +1,5 @@
+require 'rails_helper'
+
+RSpec.describe JobCruncher, type: :model do
+  pending "add some examples to (or delete) #{__FILE__}"
+end
```
patmbolger: You can delete this file as the job_cruncher_spec.rb is located under the 'services' directory.

db/migrate/20160501153517_create_job_crunchers.rb
vijayakarumudi: I will remove this file.

spec/factories/job_crunchers.rb
vijayakarumudi: will remove this one also.

patmbolger:
Reviewed PR .... please see comments.

spec/models/job_cruncher_spec.rb
vijayakarumudi: I deleted this file.

spec/factories/job_crunchers.rb
vijayakarumudi: I removed this file.

db/migrate/20160501153517_create_job_crunchers.rb
vijayakarumudi: I removed this one.

app/models/job.rb
vijayakarumudi: I updated save method

vijayakarumudi:
I updated all the files based on the comments.

app/models/job.rb
```diff
@@ -40,4 +40,16 @@ def apply job_seeker
   def last_application_by_job_seeker(job_seeker)
     job_applications.where(job_seeker: job_seeker).order(:created_at).last
   end
+  
+  def save!
+      return false if not super
+      begin
+        return true if JobCruncher.create_job(id, title, description)
+      rescue 
+        errors.add(:job, 'could not be created -please try again')
```
joaopapereira: Instead of 
could not be created -please try again

Should be
could not be created, please try again

joaopapereira:
Good job,
Missing implementation of tests on the save! of the model.
Are we using the save! in the controller? if not maybe we need to override the save function instead.
Missing stubbing in the job controller, to make sure when we create a job or edit it we do not have errors.

I think this is all that we are missing now.

app/models/job.rb
```diff
@@ -40,4 +40,16 @@ def apply job_seeker
...
+        destroy
+        raise
+      end
+      false
+  end
```
patmbolger: I believe you want to override the #save action, not #saved!

The latter will raise an exception for model validation errors and we don't want that.

app/models/job.rb
vijayakarumudi: Done.

app/models/job.rb
vijayakarumudi: I changed override save action some rspec tests that are related to job controller failed. I will dis cuss this is in the Scrum.

vijayakarumudi:
Updated PR for review.

app/models/job.rb
patmbolger: Calling *after_create* with no argument has no effect.

To make the tests work (see added tests below), I changed the model as follows:

1) Added:
```
after_save :save_job_to_cruncher
```

2) Added this method:
```
def save_job_to_cruncher
    begin
      return true if JobCruncher.create_job(id, title, description)
    rescue
      errors.add(:job, 'could not be created, please try again')
      destroy
      raise ActiveRecord::RecordInvalid.new(self)
    end
  end
```

3) Removed the method *self_create*

Note that I am using *after_save* callback rather than *after_create*.  This is so it will work with both 'create' and 'update" actions.

Also, please see this for an explanation of the added method logic: http://tech.taskrabbit.com/blog/2013/05/23/rollback-after-save/

spec/models/job_spec.rb
```diff
@@ -97,4 +97,16 @@
...
+
+      stub_request(:post, CruncherService.service_url + '/authenticate').
+          to_return(body: "{\"token\": \"12345\"}", status: 200,
+          :headers => {'Content-Type'=> 'application/json'})
+      stub_request(:post, CruncherService.service_url + '/job/create').
+          to_return(body: "{\"resultCode\":\"SUCCESS\"}", status: 200,
+          :headers => {'Content-Type'=> 'application/json'})
+      expect(JobCruncher.create_job(10,'SoftwareEngineer','description of the job')).to be true
+    end
+  end
```
patmbolger: This test is testing a JobCruncher method, *not* a Job method (I should have noticed this before ...).

I pulled your branch and changed this whole 'describe' block to this:
```
describe 'Create Job method (AR model and CruncherService)' do

    before(:each) do
      stub_request(:post, CruncherService.service_url + '/authenticate').
          to_return(body: "{\"token\": \"12345\"}", status: 200,
          :headers => {'Content-Type'=> 'application/json'})
    end

    it 'is succeeds with valid parameters' do

      stub_request(:post, CruncherService.service_url + '/job/create').
          to_return(body: "{\"resultCode\":\"SUCCESS\"}", status: 200,
          :headers => {'Content-Type'=> 'application/json'})

      job = FactoryGirl.build(:job)

      expect(job.save).to be true
      expect(Job.count).to eq 1
    end

    it 'fails with invalid model parameters' do

      stub_request(:post, CruncherService.service_url + '/job/create').
          to_return(body: "{\"resultCode\":\"SUCCESS\"}", status: 200,
          :headers => {'Content-Type'=> 'application/json'})

      job = FactoryGirl.build(:job, title: nil)

      expect(job.save).to be false
      expect(job.errors.full_messages).
              to include("Title can't be blank")
      expect(Job.count).to eq 0
    end

    it 'fails with valid model but cruncher create failure' do

      stub_request(:post, CruncherService.service_url + '/job/create').
          to_raise(RuntimeError)

      job = FactoryGirl.build(:job)

      expect(job.save).to be false
      expect(Job.count).to eq 0
    end
  end
end
```

Note that the last test tests the behavior of the *after_save* callback.  

With the changes to the Job model these tests all pass.

patmbolger:
Please note that adding the JobCruncher *create_job* call to the Job create action, many Rspec tests now fail - tests that have nothing to do with this particular story.

To see this, comment out this WebMock setting in spec/spec_helper.rb (line 31):
```
#  This allows non-mocked external service calls to proceed:
WebMock.allow_net_connect!
```
and run Rspec tests.  

I had opened a story for this #319 (https://waffle.io/AgileVentures/MetPlus_tracker/cards/5731feddabf66d0101b4706c)

db/migrate/20160501153517_create_job_crunchers.rb
joaopapereira: File still present, and it shouldnt

spec/factories/job_crunchers.rb
joaopapereira: File still present, and it shouldnt

spec/models/job_cruncher_spec.rb
joaopapereira: File still present, and it shouldnt

app/models/job.rb
joaopapereira: I think we can remove this function for now because we are not using it, and we will have anoither story for it

joaopapereira:
As Patrick said in the previous comment, this change will impact other stories, like the Job Controller specs were we call the new function, so we need to stub out that controller as well

spec/models/job_cruncher_spec.rb
vijayakarumudi: I deleted this from current branch and i did not update to github.
I removed this file from github.

db/migrate/20160501153517_create_job_crunchers.rb
vijayakarumudi: I removed

vijayakarumudi:
Updated PR for review

spec/services/job_cruncher_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    it 'returns success (true) for create a new job' do
+
+       stub_request(:post, CruncherService.service_url + '/job/create').
+           to_return(body: "{\"resultCode\":\"SUCCESS\"}" , status: 200,
+           :headers => {'Content-Type' => 'application/json'})
+       expect(JobCruncher.create_job(10,'Software Engineer',
+              'description of the job')).to be true
+    end
+
+    it 'returns fail(false) for create a new job' do
```
patmbolger: suggest:
```
it 'returns fail (false) if cruncher create fails' do
```

spec/services/job_cruncher_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+    end
+
+    it 'returns success (true) for job already exists' do
+
+       stub_request(:post, CruncherService.service_url + '/job/create').
+           to_return(body: "{\"resultCode\":\"SUCCESS\"}" , status: 200,
+           :headers => {'Content-Type' => 'application/json'})
+       expect(JobCruncher.create_job(10,'Software Engineer',
+              'description of the job')).to be true
+    end
```
patmbolger: The actual behavior of the cruncher service, when attempting to create a job that already exists, it to return a failure status.  This is the return value from the call to RestClient.post (non-stubbed):

screen shot 2016-05-19 at 7 30 11 am

So, this test should be changed to test a failure (false) return value when attempting to create a duplicate job.  The stub for '/job/create/' should return resultCode of 'JOB_ID_EXISTS' - that should cause a false return from JobCruncher.create_job  (which is the behavior we are testing in this test).

patmbolger:

spec/services/job_cruncher_spec.rb
vijayakarumudi: I updated based on your suggestion.

spec/services/job_cruncher_spec.rb
vijayakarumudi: I changed this test to the actual  behavior of cruncher services.

vijayakarumudi:
PR for review.

patmbolger:
Updated PR looks good.

joaopapereira:
Good job @vijayakarumudi the PR looks good going to merge

PR #285, opened 2016-05-15
Closes AgileVentures/MetPlus_tracker#320

patmbolger:


app/views/agency_people/show.html.haml
```diff
@@ -1,3 +1,109 @@
-%h3 Agency Person
+.row
+  .col-sm-2
+    %h3
+      = @agency_person.full_name(last_name_first: false)
 
-= render @agency_person
```
joaopapereira: Any special reason for not using the partial anymore?

patmbolger: The partial was not being used by anything other than the #show view.

As it is, the #show view contains controls used by the agency admin (e.g. 'Edit Person').   If we create an agency person view for job seekers (which I think we will) we should factor some of the #show view into a partial for that purpose.

joaopapereira: Ok then

joaopapereira:
Looks good going to merge

PR #290, opened 2016-05-25
WIP- Refactor302

sidhene:
No need to review - changing the devise steps is part of this story and I expect that will affect several other features. Using Semaphore to find out which.


joaopapereira:
When this is ready to review, please change the name to the correct version

joaopapereira:
Going to close the PR, since as not activity since May and code changed

PR #291, opened 2016-05-28
Closes AgileVentures/MetPlus_tracker#268

patmbolger:
Associating job skills with a job.

### DIFF
--- a/app/models/skill.rb
+++ b/app/models/skill.rb
@@ -2,10 +2,10 @@ class Skill < ActiveRecord::Base
 
   validates_presence_of   :name
   validates_uniqueness_of :name, case_sensitive: false
-  
+
   validates_presence_of :description
 
-  has_and_belongs_to_many :job_categories
-  has_many :job_skills
+  has_many :job_skills, dependent: :destroy
+  
   has_many :jobs, through: :job_skills
 end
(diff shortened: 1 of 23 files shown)


app/models/skill.rb
```diff
@@ -2,10 +2,10 @@ class Skill < ActiveRecord::Base
 
   validates_presence_of   :name
   validates_uniqueness_of :name, case_sensitive: false
-  
+
   validates_presence_of :description
 
-  has_and_belongs_to_many :job_categories
-  has_many :job_skills
+  has_many :job_skills, dependent: :destroy
```
joaopapereira: Not sure if this is a good approach.
Despite that I do not believe we will ever remove a Skill, the removal of it in the system should only be allowed if no jobs depend on it, or if the Job depends it should be Inactive.

joaopapereira:
Looks good, please review my comment.
Another thing that we might need to get through is the Tabs vs Spaces in the editors, because you have a bunch of changes in your PR that are spaces that were removed and things like that

app/models/skill.rb
patmbolger: We've been able to delete a skill for some time now.  This is just to make sure that if a skill is deleted the associated JobSkill is not orphaned.

I was going to create a story to decide whether and when we should be able to delete a skill.  I am OK with removing the ability to delete a skill.  We can add back in later if we decide that is what we want.

joaopapereira:
Add the story #337 to solve the previous comment

joaopapereira:
Going to merge the story

PR #292, opened 2016-05-30
Closes AgileVentures/MetPlus_tracker#223

patmbolger:
List jobs for job seeker, paginated view in:

1. JS home page
2. JS show page

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -1,6 +1,7 @@
 class JobSeekersController < ApplicationController
 
   include UserParameters
+  include JobSeekersViewer
 
   def new
     @jobseeker = JobSeeker.new
@@ -89,6 +90,7 @@ class JobSeekersController < ApplicationController
     @newjobs = Job.new_jobs(@jobseeker.last_sign_in_at).paginate(:page => params[:page], :per_page => 5)
     @past_week = Job.new_jobs(Time.now - 3.weeks).paginate(:page => params[:page], :per_page => 5)
 
+    @application_type = params[:application_type] || 'my-applied'
   end
 
   def index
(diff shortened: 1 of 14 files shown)


app/controllers/job_seekers_controller.rb
```diff
@@ -89,6 +90,7 @@ def home
     @newjobs = Job.new_jobs(@jobseeker.last_sign_in_at).paginate(:page => params[:page], :per_page => 5)
     @past_week = Job.new_jobs(Time.now - 3.weeks).paginate(:page => params[:page], :per_page => 5)
 
+    @application_type = params[:application_type] || 'my-applied'
```
joaopapereira: Are we planning on having this sent by parameters? If not we could just set the values directly in the view

app/controllers/job_seekers_controller.rb
```diff
@@ -97,8 +99,22 @@ def index
 
   def show
     @jobseeker = JobSeeker.find(params[:id])
+    @application_type = params[:application_type] || 'js-applied'
```
joaopapereira: Are we planning on having this sent by parameters? If not we could just set the values directly in the view

app/controllers/job_seekers_controller.rb
```diff
@@ -97,8 +99,22 @@ def index
...
   end
 
+  def applied_jobs
```
joaopapereira: Not 100% sure if this should be here or should be in the Jobs controller, but as you added the view in the job maybe we should move the function there

joaopapereira:
Please review my comments

app/controllers/job_seekers_controller.rb
patmbolger: Yes, the pagination SW will send this via parameter to the controller.

app/controllers/job_seekers_controller.rb
patmbolger: See comment above.

app/controllers/job_seekers_controller.rb
patmbolger: Lets discuss.  This is the action to show applied jobs for a specific job seeker.  The JS ID is passed in via params.  I put the view under jobs as most of the actual fields we display are Job attributes - so it felt more natural to put the view there.

joaopapereira:
After discussion in the Standup we decided the PR implementation was good to go

PR #297, opened 2016-06-08
Closes AgileVentures/MetPlus_tracker#341

patmbolger:
Job developer assign self to job seeker.

### DIFF
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -63,6 +63,45 @@ class AgencyPeopleController < ApplicationController
     end
   end
 
+  def assign_job_seeker
+    # Assign agency person, in specified role, to the job seeker
+
+    @agency_person = AgencyPerson.find(params[:id])
+    @job_seeker    = JobSeeker.find(params[:job_seeker_id])
+
+    role_key = params[:agency_role].to_sym
+
+    # confirm that agency person has this role and assign person to job seeker
+    case role_key
+    when :JD
+      return render(json: {:message => 'Agency Person is not a job developer'},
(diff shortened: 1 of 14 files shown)


app/controllers/agency_people_controller.rb
```diff
@@ -63,6 +63,45 @@ def update
     end
   end
 
+  def assign_job_seeker
```
joaopapereira: We need to talk about what should we do with code climate evaluation.

app/controllers/agency_people_controller.rb
```diff
@@ -63,6 +63,45 @@ def update
...
+                    status: 400)
+    end
+
+    if request.xhr?
+      render :partial => 'job_seekers/assigned_agency_person',
+             :locals => {job_seeker: @job_seeker,
+                         agency_person: @agency_person,
+                         agency_role: params[:agency_role]}
+    else
+      redirect_to job_seeker_path(@job_seeker)
```
joaopapereira: Are we going to allow none xhr requests on this endpoint? If not this should be moved to the beginning of the function

joaopapereira:
The pr looks good read my comments and we also need to talk about the code climate

app/controllers/agency_people_controller.rb
patmbolger: I am allowing non-xor requests for this action since there is no reason to not allow that and the logic is the same regardless.  Right now, there is no non-xor request, but we have that capability if it makes sense to use that in the future.

joaopapereira:
After talking in the standup we decided that we will merge this PR we will revisit later this update function

PR #296, opened 2016-06-05
Closes AgileVentures/MetPlus_tracker#338

patmbolger:
Notify job developer when assigned a job seeker.

PR #299, opened 2016-06-10
Closes AgileVentures/MetPlus_tracker#344

patmbolger:
Notify case manager of job seeker assignment (by agency admin).

This PR contains changes that are also included in PR #296 (waffle #338).  That PR should be merged first, and then this PR should be rebased.

### DIFF
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -56,7 +56,7 @@ class AgencyPeopleController < ApplicationController
         obj = Struct.new(:job_seeker, :job_developer)
         Event.create(:JS_ASSIGN_JD, obj.new(JobSeeker.find(js_id),
                                             @agency_person))
-      end
+      end if new_jd_job_seeker_ids
 
       flash[:notice] = "Agency person was successfully updated."
       redirect_to agency_person_path(@agency_person)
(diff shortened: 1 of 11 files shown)


joaopapereira:
Please rebase this pr

patmbolger:
PR rebased.

app/controllers/agency_people_controller.rb
```diff
@@ -132,5 +136,20 @@ def new_job_seeker_ids agency_person, job_seeker_ids, role_key
     return new_job_seeker_ids
   end
 
+  def notify_ap_new_js_assignments agency_person, job_seeker_ids, role_key
```
joaopapereira: Not sure if this should be in the controller, maybe the notification function should be inside the event. or even if this notification should be sent by the Model when updating the JS.

app/models/event.rb
```diff
@@ -98,20 +99,35 @@ def self.create(evt_type, evt_obj)
...
+    when :JS_ASSIGN_JD       # evt_obj = struct(:job_seeker, :agency_person)
       # Business rules:
       #    Notify job developer (email and popup)
 
       Pusher.trigger('pusher_control',
                      EVT_TYPE[:JS_ASSIGN_JD],
                      {js_id:   evt_obj.job_seeker.id,
                       js_name: evt_obj.job_seeker.full_name(last_name_first: false),
-                      jd_user_id: evt_obj.job_developer.user.id})
+                      jd_user_id: evt_obj.agency_person.user.id})
```
joaopapereira: Looking at this change, I am not sure the evt_obj is a good approach, maybe we need to revisit this later and create functions per event type instead of this huge Case that brings a lot of coupling to the code

joaopapereira:
Please review my comments, we need to talk about the code climate as well

joaopapereira:
Looks good, going to merge

PR #301, opened 2016-06-16
Closes AgileVentures/MetPlus_tracker#339

patmbolger:
job form:
  - limit list of addresses to those associated with chosen (or given) company
	
company person job create/update
  - allow change of company address (job location), including null

agency person job create/update
  - allow selection of company for the job
  - allow change of company address (job location), including null
  - address list (for selected company) is updated via XHR action when company selection changes


joaopapereira:
Looks good, going to merge

PR #303, opened 2016-06-18
Closes AgileVentures/MetPlus_tracker#356

patmbolger:
In Event.create there are calls to NotifyEmailJob that include the argument '@@delay.seconds'. The class var '@@delay' should not be called directly, but rather through the class method Event.delay_seconds.


joaopapereira:
Looks good, going to merge

PR #317, opened 2016-07-11
Closes AgileVentures/MetPlus_tracker#368

sandhyakoti:
The color needs to be the same as the header color
    The footer is too tall - it needs to be as short as (or shorter than) the header.
    There is a visible white margin around the footer (unlike the header). This needs to be removed.

Files changed:
	app/views/layouts/_footer.html.haml
	1. Panel removed - to eliminate the white line
	2. div class footer added

	app/assets/stylesheets/application.css
	1. Footer height increased
	2. Background colour changed
	3. Footer link styling

	.footer {
    	position: absolute;
	    bottom: 0px;
    	width: 100%;
	    height: 45px;
    	background-color: #0C7BA1;
	}
	.footer > a {
    	color: #fff;
	    text-decoration: underline;
	}

### DIFF
--- a/app/assets/stylesheets/application.css
+++ b/app/assets/stylesheets/application.css
@@ -98,8 +98,13 @@ body {
     position: absolute;
     bottom: 0px;
     width: 100%;
-    height: 20px;
-    background-color: #f5f5f5;
+    height: 45px;
+    background-color: #0C7BA1;
+}
+
+.footer > a {
+    color: #fff;
+    text-decoration: underline;     
 }
 
 .navbar-default .navbar-text > a {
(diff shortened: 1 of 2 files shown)


patmbolger:
The color looks good and the size is better as well.

The cucumber test failure is due to that fact that the footer is occluding (covering up) the bottom portion of views - at least those with buttons.  For instance, this is the page were the cucumber test step is failing:

screen shot 2016-04-11 at 10 01 38 am

However, when I reduce the height of the window you can see that the footer occludes the buttons:

screen shot 2016-07-11 at 4 28 31 pm

The button is covered up and thus cucumber cannot "see" the delete button, and the test fails.

joaopapereira:
Looks good, going to merge

PR #319, opened 2016-07-12
Closes AgileVentures/MetPlus_tracker#369

patmbolger:
**Story**: add states to Job model

The states are defined in the first table here: http://www.agileventures.org/projects/metplus/documents/system-behavior

**Implementation**:

- Add to the model, e.g.
    STATUS = {ACTIVE: 'active',
          FILLED: 'filled',
          REVOKED: 'revoked'}
(note that job 'state' is called 'status' here since a job has_one address, and an address has a 'state' field - thus to avoid ambiguous SQL statements we use 'status'.

- Create a migration file to add field to DB

- Add tests to job_spec.rb

- Add 'status' field to Job factory

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -29,9 +29,14 @@ class Job < ActiveRecord::Base
   validates_length_of   :description, maximum: 10000
   validates_presence_of :company_id
   validates_presence_of :company_person_id, allow_nil: true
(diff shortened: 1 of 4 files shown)


joaopapereira:
Looks good, going to merge

PR #320, opened 2016-07-13
Closes AgileVentures/MetPlus_tracker#308

dsmegha:
Fixed the regex in app/validators/phone_validator.rb to reflect the necessary changes.

Since the validators for phone and fax essentially do the same thing, removed the fax_validator.rb and made sure both the phone and fax fields in the agency, company and user models are using the phone_validator.rb

Added placeholders in the forms to help users input valid phone numbers.

Made changes in the test files so that phone numbers are validated correctly.


joaopapereira:
Looks good, going to merge

PR #321, opened 2016-07-13
Agency person js assigned jd#316

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#316
Pagination in Agency people homepage : When Jobseeker assigned Job developer


vijayakarumudi:
PR for review

app/views/agency_people/_assigned_job_seekers_jd.html.haml
```diff
@@ -0,0 +1,33 @@
+%table.table.table-hover
```
joaopapereira: Do weer so use this file?

joaopapereira:
Looks good, please review my comment. 
We'd well need to refractor the functions to increase the code climate site. But I belive that can wait for now

app/views/agency_people/_assigned_job_seekers_jd.html.haml
vijayakarumudi: No We will delete this file

vijayakarumudi: I deleted that file.

vijayakarumudi:
I updated my PR.

app/controllers/agency_people_controller.rb
```diff
@@ -15,7 +15,8 @@ def home
     @agency_person = AgencyPerson.find(params[:id])
     @agency = @agency_person.agency
     @task_type = 'mine-open'
-    @people_type = 'jobseeker-cm'
+    @people_type_cm = 'jobseeker-cm'
+    @people_type_jd = 'jobseeker-jd'
     @js_without_jd = JobSeeker.paginate(:page=> params[:js_without_jd_page], :per_page=>5).js_without_jd
     @js_without_cm = JobSeeker.paginate(:page=> params[:js_without_cm_page], :per_page=>5).js_without_cm
     @your_jobseekers_jd = JobSeeker.paginate(:page=> params[:your_jobseekers_jd_page], :per_page=> 5).your_jobseekers_jd(@agency_person)
```
patmbolger: This variable (of class JobSeeker::ActiveRecordRelation) is not actually used to load records for the view.  It is used sole in the view to determine if there are any job seekers assigned to this person (in the person's role of job developer).  The actual loading of records is done via the XHR call to action #list_js_jd.

So, calling *paginate* on it is not serving any purpose.

Recommend changing this line to:
```
@your_jobseekers_jd = JobSeeker.your_jobseekers_jd(@agency_person)
```

app/views/agency_people/home.html.haml
```diff
@@ -13,16 +13,16 @@
   - if @your_jobseekers_jd.empty?
     %i There are no job seekers assigned to you yet.
   - else
-    .col-sm-9
-    = render partial: 'job_seekers' , :locals => {jobseekers: @your_jobseekers_jd}
-
+    .pagination-div{id: "agency_person-#{@people_type}",
```
patmbolger: This is using the same ID as line 24, and needs to be unique on the page.

We have a instance var set up in the controller for this, so this line should be:
```
.pagination-div{id: "agency_person-#{@people_type_jd}",
```

app/views/agency_people/home.html.haml
```diff
@@ -13,16 +13,16 @@
...
+     data: {url: list_js_jd_agency_people_path(@agency_person.id,@people_type_jd)}}
+    
 - if @agency_person.is_case_manager?(@agency)
   %h3 Your Job Seekers (as case manager)
   - if @your_jobseekers_cm.empty?
     %i There are no job seekers assigned to you yet.
   - else
     .pagination-div{id: "agency_person-#{@people_type}",
```
patmbolger: See comment above.   This line should be:
```
.pagination-div{id: "agency_person-#{@people_type_cm}",
```

patmbolger:

I recommend also adding a simple scenario to agency_people.feature which just loads the AP home page and confirms that the page contains names of 1) job seekers assigned to person as job developer, and 2) job seekers assigned to person as case manager.

There is a fixture step available to create these assignments in the Background setup of the test:
```
Given(/^the following agency relations exist:$/)
```
See example usage in job_seeker.feature.

With this test, we would have caught the fact that 1) only job seekers assigned to person as job developer are showing up on the home page, and 2) those job seekers are actually assigned to the agency person as a *case manager*.  (this is a result of the duplicate ID's in the view, as explained in comments).

app/views/agency_people/home.html.haml
vijayakarumudi: I updated this one.

app/views/agency_people/home.html.haml
vijayakarumudi: This is done.

app/controllers/agency_people_controller.rb
vijayakarumudi: we discussed this pagination at scrum. I updated those lines.

features/agency_person.feature
```diff
@@ -128,3 +128,21 @@ Feature: Agency Person
...
+    And I wait 1 second
+    And I should see "Your Job Seekers (as job developer)"
+  
+  @javascript
+  Scenario: Job seekers assigned to person as Case manager
+    Given I am on the home page
+    And I login as "jane@metplus.org" with password "qwerty123"
+    And I should be on the Agency Person 'jane@metplus.org' Home page
+    And I wait 1 second
+    And I should see "Your Job Seekers (as case manager)"
```
patmbolger: Both of these scenarios (this and the preceding one) don't test whether the actual job seekers are loaded in the view.  I suggest:

1) Use just one scenario rather than these two
2) In that scenario, login as an agency person who has both a JD and a CM role
3) The home page for that person should show assigned job seekers for both roles

Steps 2 and 3 will require additional data to be set up in the Background section.  You can use this step to help with that:
```
Given(/^the following company roles exist:$/) do |table|
  table.hashes.each do |hash|
    hash['role'] = CompanyRole::ROLE[hash['role'].to_sym]
    CompanyRole.create!(hash)
  end
end
```
This lets you creates relationships (JD or CM) between a job seeker and an agency person.

To look for job seekers for the JD role, you want to make sure that they occur in the "your job seekers as job developer" table.  You can use this step to help with that:
```
Then(/^I should see "([^"]*)" between "([^"]*)" and "([^"]*)"$/) do |toSearch, first, last|
  regex = /#{first}.+#{toSearch}.+#{last}/
  search_text regex
end
```
So, for instance, you could have a step like:
```
And I should see "Seeker, Tom" between "Your Job Seekers (as job developer)" and "Your Job Seekers (as case manager)"
```
(since the JD assignments are shown on the screen first.  Note here that "Tom Seeker" is a job seeker assigned to the person in their JD role).

For the job seekers assigned to the person in their CM role, you could use a similar step:
```
And I should see "Jones, Mary" after "Your Job Seekers (as case manager)"
```
(here, "Mary Jones" is a job seeker assigned to the person in their CM role).

patmbolger:
Reviewed PR - see comments re an improved cucumber test.

patmbolger:
Reviewed PR .... looks good.  Thanks for adding the suggested cucumber scenario.

joaopapereira:
Looks good, going to merge

PR #322, opened 2016-07-14
Cleared the select index.

sandhyakoti:


app/assets/javascripts/tasks.js
```diff
@@ -109,8 +109,12 @@ var TaskModal = {
                       Notification.error_notification(xhrObj.responseJSON['message']);
                   }
               });
+              $("#task_assign_select")[0].selectedIndex = -1;    
               return false;
           });
+          $('#cancelTaskModal_button').off('click').click( function() {
+              $("#task_assign_select")[0].selectedIndex = -1;
+          });
```
patmbolger: These 3 lines are within the handler for the task modal "assign" button click handler.  In general, the preceding code will result in a successful assignment, and line 113 will cause a return - hence these lines will not be run.

So, the only way for the cached field value to be cleared is if an assignment is made by the user.  If the modal is cancelled by the user, the cache is not cleared.  (unless I am looking at this incorrectly - I do have this feature branch set up in my environment is this is the behavior I am seeing).

patmbolger:
Reviewed the PR and the fix is not working for me.

Also, consider a simpler approach ... clear the cached field value whenever the modal is shown to the user.  This way the field is cleared regardless of whether the previous user action in the modal was to click "cancel" or "assign".

This can be done by adding the second line shown here to tasks.js (it would be line 6 in that file):
```
this.load_assign_modal = function (event) {
        $('#task_assign_select').val('');
```

PR #327, opened 2016-07-16
Closes AgileVentures/MetPlus_tracker#375

patmbolger:
**Story Description**:

When a company registers (actually, requests registration) in PETS a popup notification is shown to all logged in agency people.  The name of the company in the message is a link to the company 'show' view.  Rails is indicating that there is no such routing, for example:
```
No route matches [GET] "/company_admin/companies/202"
```

**Fixed via**:

1. Corrected URL in pusher notification message

2. Added steps to event_management.feature scenario to confirm that URL in popup works and creates another window showing the company information.

3. Refactored (and simplified) company *show* view (and associated partials) for agency admin, company admin, and non-admin user.


app/controllers/companies_controller.rb
```diff
@@ -6,6 +6,14 @@ def show
     @company        = Company.find(params[:id])
     @company_admins = Company.company_admins(@company)
     @people_type    = 'company-all'
+    @admin_type     = params[:admin_type]
+    if @admin_type == 'NONE'
+      if pets_user.is_company_admin? @company
+        @admin_type = 'CA'
```
joaopapereira: I do not like this solution of having these string passing around. Isnt enough to check the current user to see if we have a AA or a CA?

patmbolger:
Refactored to:

1. remove 'admin_type' from params for companies_controller action,

2. negate passing string to views for determining specific items to show specific user types (non-admin, company admin, agency admin).

joaopapereira:
Looks good, going to merge

PR #330, opened 2016-07-18
Closes AgileVentures/MetPlus_tracker#354 (Add address to company person profile )

jeanwei:
1. Selection of company addresses is included to company_people/edit_profile.html.haml
2. :address_id is permitted in company_people_controller
3. change #update_profile redirect_to #show if success 
    (note: url for #show is /company_admin/company_people/:id)
4. change on rspec & cucumber files accordingly
5. lengthen the waiting time in 2 cucumber features to allow passing

### DIFF
--- a/app/views/company_people/edit_profile.html.haml
+++ b/app/views/company_people/edit_profile.html.haml
@@ -13,6 +13,11 @@
 	- @user_form = cp
 	= render partial: 'users/subform'
 	.form-group
+		= cp.label :address, 'Address', class: 'control-label col-sm-2'
+		.col-sm-6
+			= cp.select(:address_id, @company_addresses.collect { |a| [a.full_address, a.id] }, {include_blank: true}, class: 'form-control')
+
+	.form-group
 		.col-sm-2.col-sm-offset-2
 			=@user_form.submit class: 'btn btn-primary'
 		.col-sm-2.col-sm-offset-1
--- a/app/views/company_people/edit_profile.html.haml
+++ b/app/views/company_people/edit_profile.html.haml
@@ -13,7 +13,7 @@
 	- @user_form = cp
 	= render partial: 'users/subform'
 	.form-group
-		= cp.label :address, 'Address', class: 'control-label col-sm-2'
+		= cp.label :address_id, 'Address', class: 'control-label col-sm-2'
 		.col-sm-6
 			= cp.select(:address_id, @company_addresses.collect { |a| [a.full_address, a.id] }, {include_blank: true}, class: 'form-control')
 
@@ -22,5 +22,4 @@
 			=@user_form.submit class: 'btn btn-primary'
 		.col-sm-2.col-sm-offset-1
 			=link_to 'Cancel', root_path, class: 'btn btn-danger'
-%i (leave password blank, if you don't want to change it)
 %br
--- a/features/company_person.feature
+++ b/features/company_person.feature
@@ -30,6 +30,13 @@ Feature: Company Person
       | MetPlus | Widgets Inc. | widgets.com | 555-222-3333 | corp@widgets.com | 12-3456789 | Active |
       | MetPlus | Feature Inc. | feature.com | 555-222-3333 | corp@feature.com | 12-3456788 | Active |
 
+    Given the following company addresses exist:
+      | company       | street           | city    | zipcode | state      |
+      | Widgets Inc.  | 12 Main Street   | Detroit | 02034   | Michigan   |
+      | Widgets Inc.  | 14 Main Street   | Detroit | 02034   | Michigan   |
+      | Feature Inc.  | 100 River Valley | Utah    | 12334   | New Jersey |
+      | Feature Inc.  | 111 River Valley | Utah    | 12334   | New Jersey |
+
     Given the following company people exist:
       | company      | role  | first_name | last_name | email            | password  | phone        |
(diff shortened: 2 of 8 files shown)


patmbolger:
I was thinking that we would have the company person add their own address explicitly, rather than choose from a list of company addresses.  This would allow, for instance, for a remote worker who works as home to enter their home address.

However, we don't know if the above is useful or not, so we should hold off until and if the client asks for that.

So, selecting a company address is OK, I think.

In addition to my other comments, could you please remove this line (25) from *edit_profile.html.haml*:
```
%i (leave password blank, if you don't want to change it)
```
(we have the placeholders in the password/password_comfirmation fields to we don't need this in the form)

app/views/company_people/edit_profile.html.haml
```diff
@@ -13,6 +13,11 @@
 	- @user_form = cp
 	= render partial: 'users/subform'
 	.form-group
+		= cp.label :address, 'Address', class: 'control-label col-sm-2'
```
patmbolger: suggest changing this line to:
```
= cp.label :address_id, 'Address', class: 'control-label col-sm-2'
```
That way the label will be linked with the selection list (this will allow improvement to cucumber scenarios - see below).

features/company_person.feature
```diff
@@ -156,6 +163,30 @@ Feature: Company Person
...
     And I should see "Tom"
 
+  Scenario: company admin can update address
+    Given I am on the home page
+    And I login as "ca@widgets.com" with password "qwerty123"
+    Then I press "edit-profile"
+    And I do not have an address
+    And I should see selections of "Widgets Inc." addresses 
+    And I should not see selections of "Feature Inc." addresses 
+    And I select "12 Main Street Detroit, Michigan 02034" in select list "company_person_address_id"
```
patmbolger: It is best to have scenario steps reflect what the user can see in the view.  So, suggest changing this line to:
```
And I select "12 Main Street Detroit, Michigan 02034" in select list "Address"
```

features/company_person.feature
```diff
@@ -156,6 +163,30 @@ Feature: Company Person
...
+    And I should see "12 Main Street Detroit, Michigan 02034"
+
+  Scenario: company contact can update address
+    Given I am on the home page
+    And I login as "jane@widgets.com" with password "qwerty123"
+    Then I press "edit-profile"
+    And I do not have an address
+    And I should see selections of "Widgets Inc." addresses 
+    And I should not see selections of "Feature Inc." addresses 
+    And I select "12 Main Street Detroit, Michigan 02034" in select list "company_person_address_id"
```
patmbolger: Suggest this is changed to:
```
And I select "12 Main Street Detroit, Michigan 02034" in select list "Address"
```

patmbolger:
PR reviewed ..... looks great, but please see comments.

jeanwei:
Thanks for the feedback. I have implemented them. Please review.

As for the address input, I have thought of text field for updating the address. To my opinion, care needs to be taken on user input to avoid repeated company address created & company addresses should not be created or updated by company person except admin. select options seems to be the most straightforward way of implementation without user input check. However, when home address is considered, yea.. we need to discuss a specific way to implement (input text for add address & include the select address list, etc.)

patmbolger:
Reviewed PR .... looks good.

app/views/company_people/edit_profile.html.haml
```diff
@@ -13,9 +13,13 @@
...
+		= cp.label :address_id, 'Address', class: 'control-label col-sm-2'
+		.col-sm-6
+			= cp.select(:address_id, @company_addresses.collect { |a| [a.full_address, a.id] }, {include_blank: true}, class: 'form-control')
+
+	.form-group
 		.col-sm-2.col-sm-offset-2
 			=@user_form.submit class: 'btn btn-primary'
 		.col-sm-2.col-sm-offset-1
 			=link_to 'Cancel', root_path, class: 'btn btn-danger'
-%i (leave password blank, if you don't want to change it)
```
joaopapereira: Did we add this information somewhere else? Or are we assuming that the users know this?

joaopapereira:
PR looks good, please review my comment

app/views/company_people/edit_profile.html.haml
jeanwei: We have added this information in the password placeholder

jeanwei: Just read through the users/_subform.haml.html. Do you mind if I ask what is the if statement doing here with new_form? what is new_form for?
- if defined? new_form and new_form
		- pw_placeholder = ''
	- else
		- pw_placeholder = "leave blank if you don't want to change password"

joaopapereira:
I am ok with the comments, going to merge the PR

app/views/company_people/edit_profile.html.haml
patmbolger: The subform is called as a partial from other forms.  If the local var 'new_form' is defined in the rendering of the subform, the placeholder text is not shown.  That is because for a new object we *need* a password (and confirmation) to be entered, and the placeholder text in that case would be misleading.

Here is an example render from job_seekers/new.html.haml:
```
= render partial: 'form', locals: {new_form: true}
```
(I think 'form' then calls subform and passes on the new_form value)

I wonder if Rails has a cleaner way to do this?

PR #335, opened 2016-07-27
Js without assigned to jd 317

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#317
Pagination in Agency people homepage : JobSeeker's without Jobdeveloper

### DIFF
--- a/app/controllers/agency_people_controller.rb
+++ b/app/controllers/agency_people_controller.rb
@@ -196,20 +196,19 @@ class AgencyPeopleController < ApplicationController
 
     raise 'Unsupported request' if not request.xhr?
        
-    @agency_person= AgencyPerson.find(params[:id])
+    agency_person= AgencyPerson.find(params[:id])
 
-    @people_type_without_jd = params[:people_type] || 'jobseeker-without-jd'
+    people_type_without_jd = params[:people_type] || 'jobseeker-without-jd'
 
-    @people = []
-    @people = display_job_seekers @people_type_without_jd, @agency_person
+    people = []
+    people = display_job_seekers people_type_without_jd, agency_person
     
     
-
      render :partial => 'agency_people/assigned_job_seekers', 
-                       locals: {jobseekers: @people,
+                       locals: {jobseekers: people,
                                 controller_action:'list_js_without_jd',
-                                people_type: @people_type_without_jd,
-                                agency_person:@agency_person}
+                                people_type: people_type_without_jd,
+                                agency_person: agency_person}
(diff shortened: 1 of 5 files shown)


vijayakarumudi:
PR for review

app/controllers/agency_people_controller.rb
```diff
@@ -190,6 +192,29 @@ def list_js_jd
...
+  def list_js_without_jd
+
+    raise 'Unsupported request' if not request.xhr?
+       
+    @agency_person= AgencyPerson.find(params[:id])
+
+    @people_type_without_jd = params[:people_type] || 'jobseeker-without-jd'
+
+    @people = []
+    @people = display_job_seekers @people_type_without_jd, @agency_person
```
patmbolger: The variables @agency_person, @people_type_without_jd, @people are passed directly into the partials via the 'locals' hash, so these vars don't need to be instance vars and could be simply local vars for this method.  I suspect that as a result there is a little more overhead occurring than we need (Rails detects controller instance vars and makes those vars available to the view - even though the controller and view are *not* related via class structure - this requires work that can be avoided with local vars).

patmbolger:
Reviewed PR .... looks good.  

I do have one suggested cleanup related to controller instance vars.

app/controllers/agency_people_controller.rb
joaopapereira: Good point here

config/routes.rb
```diff
@@ -44,6 +44,13 @@
              'agency_people#list_js_jd', as: :list_js_jd_agency_people
 
 
+
+  get 'agency_people/:id/list_js_without_jd/:people_type' => 
+              'agency_people#list_js_without_jd',as: :list_js_without_jd_agency_people
+
```
joaopapereira: I think we can remove these extra spaces as well

joaopapereira:
Just finished my review, and the PR looks good, please just revisit my comments and Patrick's and I believe we are done

config/routes.rb
vijayakarumudi: I removed the space.

app/controllers/agency_people_controller.rb
vijayakarumudi: I updated variables.

vijayakarumudi:
updated PR.

patmbolger:
Reviewed PR again .... looks good.

joaopapereira:
Looks good, going to merge

PR #336, opened 2016-07-30
Closes AgileVentures/MetPlus_tracker#337

sandhyakoti:
Remove ability to delete a skill when it is associated with a job.
The same template file was being used for both Job Specialities and Job Skills. Hence a new template file has been created for rendering Job Skills.

### DIFF
--- a/app/views/agency_admin/_job_skills.html.haml
+++ b/app/views/agency_admin/_job_skills.html.haml
@@ -1,0 +1,26 @@
+%div(id="#{partial_id}")
+  %table.table.table-hover
+    %thead
+      %tr
+        %th
+          %strong Name
+        %th
+          %strong Description
+        %th
+          %strong Delete
+    %tbody
+      - job_skills.each do |job_prop|
+        %tr
+          %td
+            = link_to "#{job_prop.name}", send(show_property_path, job_prop), method: :edit
+          %td
+            = job_prop.description
+          %td
+            - if job_prop.has_job?
+              %span.glyphicon.glyphicon-remove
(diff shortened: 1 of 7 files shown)


app/views/agency_admin/_job_skills.html.haml
```diff
@@ -0,0 +1,26 @@
...
+    %tbody
+      - job_skills.each do |job_prop|
+        %tr
+          %td
+            = link_to "#{job_prop.name}", send(show_property_path, job_prop), method: :edit
+          %td
+            = job_prop.description
+          %td
+            - if job_prop.has_job?
+              %span.glyphicon.glyphicon-remove
```
patmbolger: I think it could be confusing for the user to have to discern between a plain black "X" and a red "X", and understand that the former indicates that the skill cannot be deleted because it is associated with one or more jobs  (and, a color-blind person would have a real problem).

I suggest showing  this instead:
```
(in use)
```
That would be clearer to the user.  And, we could eventually make that a link to go to a view showing all jobs that use that skill.  (this would be very useful in the case where a skill really *should* be deleted, but the jobs need to be edited first to remove that skill).  (Note: not asking for this now ..... but we should add a story for this and assigned the label "Release 2.0")

patmbolger:
Review PR.  Please see comment re skills delete control indicator.

Also, could you please add a couple of steps for this feature to agency_admin.features?  

There are some scenarios for 'job specialties' in that file, but none for job skills.  At the least, we should test that a skill assigned to a job cannot be deleted in the UI.

PR #337, opened 2016-07-30
Closes AgileVentures/MetPlus_tracker#371

jeanwei:
Implemented tasks
- revoke job by updating job.status from 'active' to 'revoked'
- only 'active' job can be 'revoked'
- 'revoked' job can not be applied. But still in the Jobs list.
- notification and email events to job developers after a single job is revoked
- confirmation on revoke action 
- change of view & test accordingly

not implemented
- authorization for revoke
- should the email & notification events happen for a collection of job instead?

### DIFF
--- a/app/assets/javascripts/jobs.js
+++ b/app/assets/javascripts/jobs.js
@@ -14,5 +14,20 @@ $(function () {
               $('#address_select').html(data);
             }
           });
-    });
   });
+
+  
+  $('button[data-action="revoke"], #revoke_link').on('click', function() {
+    var id = $(this).attr('data-job-id');
+    var title = $(this).attr('data-job-title');
+    var companyJobId = $(this).attr('data-job-companyJobId');
+    
+    $('#revokeModal').find('#title').html('job title: ' + title);
+    $('#revokeModal').find('#company_job_id').html('company job id: ' + companyJobId);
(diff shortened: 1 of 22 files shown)


patmbolger:
Reviewed PR ..... great work.  My comments:

1. Modal title is the name of the user - this should be the title of the job instead (this was the situation before and should have been caught earlier).

modal

2. On the jobs listing page, provide a little whitespace between the "edit" and "revoke" buttons:

joblist

3. Add job status to 'show' view.  (this is especially important for when a job developer clicks the job show link in the "job revoked" notification popup - they should see a confirmation that the job is, in fact, revoked.

4. In the file features/step_definitions/jobs_steps.rb, it would be better to have descriptive steps that are understandable to the user.  So, this step:

steps

might be replace by step(s) that could be expressed as:
```
		I should see "Revoked" next to "hr assistant"
		I should see "revoke" button for "hr assistant"
```

patmbolger:
I reviewed the PR after recent updates - it all looks good.

Thanks for making the changes.

jeanwei:
Thanks for the review. It fails both codeclimate and semaphoreci.

joaopapereira:
Looks good, going to merge

joaopapereira:
The codeclimate unless is a major problem we can ignore, but we will need to review that in due time

PR #344, opened 2016-08-06
Closes AgileVentures/MetPlus_tracker#332

arun1595:
- Add a class method `match_jobs` in CruncherService
- Invoke the `match_jobs` method from JobCruncher
- Add specs for both the methods

### DIFF
--- a/app/models/job_cruncher.rb
+++ b/app/models/job_cruncher.rb
@@ -7,5 +7,21 @@ class JobCruncher
 
      CruncherService.create_job(jobId, title, description)
   end
-end
 
+  def self.match_jobs(resumeId)
+    # If match exists, returns a hash of job matches where
+    # key is the matcher id and value is the array of job ids
+    # Otherwise returns nil(in case of resume not found)
+
+    # Example return value from Cruncher Service:
+    # { "matcher2": [3, 2], "matcher1": [1, 2] }
+    # Note that the keys returned are 'matcher1' and 'matcher2'
+    # In the current version of API, we have a single matcher - Word Count
+    # In future releases, additional matchers will be implemented and
+    # the keys might also be modified to reflect the matcher used likely
+    # { "expressionCruncher": [3, 2], "naiveBayes": [1, 2] }
+    # The caller can make use of the keys if required
+
+    CruncherService.match_jobs(resumeId)
+  end
+end
--- a/app/services/cruncher_service.rb
+++ b/app/services/cruncher_service.rb
@@ -102,6 +102,44 @@ class CruncherService
 
   end
 
+  def self.match_jobs(resume_id)
+
+    # Define retry_search to true outside the begin block
+    # otherwise retry_search would be set to true everytime
+    # we retry the block. Can also set it inside the begin block
+    # like retry_search ||= true so that its set only the first
+    # time
+
+    retry_search = true
+
+    begin
+      result = RestClient.get(service_url + '/job/match/' + resume_id.to_s,
+                              { 'X-Auth-Token': auth_token })
+
+      result_hash = JSON.parse(result)
+
+      # Set matching jobs to nil if resume couldn't be found
+      matching_jobs = nil
+
+      if result_hash['resultCode'] == 'SUCCESS'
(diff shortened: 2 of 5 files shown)


arun1595:
This can't be merged. Blocked by AgileVentures/MetPlus_tracker#393

app/models/job_cruncher.rb
```diff
@@ -7,5 +7,10 @@ def self.create_job(jobId, title, description)
 
      CruncherService.create_job(jobId, title, description)
   end
-end
 
+  def self.match_jobs(resumeId)
+    # Returns a hash of matches if match exists
+    # Otherwise returns nil(in case of resume not found)
+    CruncherService.match_jobs(resumeId)
```
patmbolger: The cruncher API state that the return body for a successful call will look like this:
```
{"resultCode":"SUCCESS","message":"Success","jobs":{"matcher2":["job3","job2"],"matcher1":["job1","job2"]}}
```
We should expand on your comments here with more explanation as to what a "hash of matches" consists of, and, in particular, what is meant by the "matcher1", "matcher2" keys.

(I assume these might be "Word Frequency", "Bayesian", etc - but that is just a guess).  

This is good information and we may want to display matching jobs by job matcher type.

app/services/cruncher_service.rb
```diff
@@ -67,6 +67,40 @@ def self.create_job(jobId, title, description)
...
+                              { 'X-Auth-Token': auth_token })
+
+      result_hash = JSON.parse(result)
+
+      # Set matching jobs to nil if resume couldn't be found
+      matching_jobs = nil
+
+      if result_hash['resultCode'] == 'SUCCESS'
+        # May or may not contain jobs matching the resume
+        matching_jobs = result_hash['jobs']
```
patmbolger: We should convert the job ID's to integers.

patmbolger:
Reviewed PR - overall looks good but please see comments.

Appreciate the useful documentation in the code.

app/models/job_cruncher.rb
arun1595: I'll add more useful comments like you suggested and update the PR.

app/models/job_cruncher.rb
```diff
@@ -7,5 +7,16 @@ def self.create_job(jobId, title, description)
...
-end
 
+  def self.match_jobs(resumeId)
+    # If match exists, returns a hash of job matches where
+    # key is the matcher id and value is the array of job ids
+    # Otherwise returns nil(in case of resume not found)
+
+    # Example return value from Cruncher Service:
+    # { "matcher2": [3, 2], "matcher1": [1, 2] }
+    # Matchers used are ExpressionCruncher(uses word frequency) and NaiveBayes
```
patmbolger: Instead of "matcher2", "matcher1" will it actually be "ExpressionCruncher", "NaiveBayes"?

If so, please update comment to reflect that.

patmbolger:
Reviewed PR again - looks better.  See one comment re clarification of a comment.

app/models/job_cruncher.rb
arun1595: The comment says that the returned value from `CruncherService.match_jobs` will be of the form     
`{ "matcher2": [3, 2], "matcher1": [1, 2] }`. I have added the additional comment to notify what types of matchers are used to do the match. It has nothing to do with the format of the returned value.

patmbolger: Yes, it's not format that I am asking about.  I am wondering what the hash key values will be(?).

I would hope that they will be useful strings, like "Word Frequency Match", or the like - rather than generic terms like "matcher1".  If the key makes sense to a user, it could be displayed alongside the results of that matcher, for example:
```
Matching jobs using "Word Frequency Match":
  Job 1
  Job 2
  etc...
```

joaopapereira:
Looks good, is this ok to be merged?

patmbolger:
Since https://github.com/AgileVentures/MetPlus_tracker/issues/393 has been closed this PR can be merged.

joaopapereira:
Going to merge

PR #345, opened 2016-08-07
Closes AgileVentures/MetPlus_tracker#385

patmbolger:
Story:
```
As a developer
I need to provide a capability to download a résumé for a specific job seeker
For various purposes
For instance, so that the résumé can be used for a job application
Or, the résumé can be viewed online
```

Implementation:

1. Add class method download_file(file_id) to CruncherService
2. Add class method download_resume(file_id) to ResumeCruncher
3. Add rspec tests for both methods above


joaopapereira:
Looks good, going to merge

PR #346, opened 2016-08-08
Closes AgileVentures/MetPlus_tracker#391

arun1595:
- Add a class method in CruncherService to fetch jobs
- Invoke the method in ResumeCruncher
- Add specs for both CruncherService and ResumeCruncher

### DIFF
--- a/app/models/resume_cruncher.rb
+++ b/app/models/resume_cruncher.rb
@@ -21,4 +21,21 @@ class ResumeCruncher
     CruncherService.download_file(file_id)
   end
 
+  def self.match_resumes(job_id)
+    # If a match exists, returns a hash of resume matches where
+    # key is the matcher id and value is the array of resume ids
+    # If no match exists, returns as an empty hash
+
+    # Example return value from CruncherService:
+    # { "matcher2": [3, 2], "matcher1": [1, 2] }
+    # Note that the keys returned are 'matcher1' and 'matcher2'
+    # In the current version of API, we have a single matcher - Word Count
+    # In future releases, additional matchers will be implemented and
+    # the keys might also be modified to reflect the matcher used like
+    # { "expressionCruncher": [3, 2], "naiveBayes": [1, 2] }
+    # The caller can make use of the keys if required
+
+    CruncherService.match_resumes(job_id)
+  end
+
 end
--- a/app/services/cruncher_service.rb
+++ b/app/services/cruncher_service.rb
@@ -41,6 +41,7 @@ class CruncherService
     end
   end
 
+
   def self.download_file(file_id)
     # Download the contents of the file which was stored in the Cruncher
     # using the file_id (e.g. resume.id).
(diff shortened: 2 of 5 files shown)


app/models/resume_cruncher.rb
```diff
@@ -14,4 +14,10 @@ def self.upload_resume(file, file_name, file_id)
     CruncherService.upload_file(file, file_name, file_id)
   end
 
+  def self.match_resumes(job_title, job_description)
+
+    #Returns an object containing the resume_ids matched by each cruncher
```
patmbolger: What kind of an "object"?

app/services/cruncher_service.rb
```diff
@@ -67,6 +67,31 @@ def self.create_job(jobId, title, description)
...
+
+    retry_match = true
+
+    begin
+      result = RestClient.post(service_url + '/curriculum/match',
+                { 'title': job_title,
+                  'description': job_description },
+                { 'Accept': 'application/json',
+                  'X-Auth-Token': auth_token })
+      matching_resumes = JSON.parse(result)['resumes']
```
patmbolger: I opened a story (https://waffle.io/AgileVentures/MetPlus_tracker/cards/57a8d01b485aeb7a00d1b9a4) to change this cruncher API to use job ID, instead of job title and description.

spec/services/cruncher_service_spec.rb
```diff
@@ -162,6 +188,19 @@
...
+    describe 'match resumes' do
+
+      it 'returns success' do
+        stub_cruncher_match_resumes
+
+        expect(CruncherService.match_resumes('Web developer',
+                    'A person experienced in developing scalable web applications'
+        )).not_to be nil
+      end
+
```
patmbolger: How about a test for "no matching résumés found" case?

patmbolger:
Reviewed PR - looks good but please review comments.

spec/services/cruncher_service_spec.rb
arun1595: If no resumes in the cruncher match with the title and description, we would get the following response.

```
{
"resultCode": "SUCCESS"
"message": "Success"
"resumes": {}
}
```

The current test makes sure that returned object from `CruncherService.match_resumes` is not nil . I think this suffices because in both the cases(match and mis-match), in case of HTTP success, a object would be returned(empty or with resume_ids). What do you think?

app/services/cruncher_service.rb
arun1595: When AgileVentures/MetPlus_tracker#396 is fixed, let me update the PR to send a GET with the job ID rather than POST which is currently used.

patmbolger:
Story https://github.com/AgileVentures/MetPlus_tracker/issues/396 ha been delivered (see: http://joaopapereira.github.io/MetPlus_resumeCruncher/#match-resumes-with-job-identifier) this PR can be adjusted to use that new API endpoint (job ID rather than job title & description).

PR #347, opened 2016-08-09
Closes AgileVentures/MetPlus_tracker#342

dsmegha:


patmbolger:
Reviewed PR and the code looks good.

There is an anomaly that we may want to address.  As the admin, when I assign an unassigned task to myself (such as 'review company registration') then that task appears in both "Agency's Remaining Tasks" and "Your Open Tasks".

Then, I can move the task to *in progress* in either table.  This results in a view like this:

screen shot 2016-08-09 at 4 59 25 pm

Perhaps we need to *not* update "Agency's Remaining Tasks" if the task is being assigned to him or herself.   On then other hand, if we don't place the task there than a second agency admin will lose visibility to the task.

So, instead, maybe we place the assigned task in "Agency's Remaining Tasks", and *not* in "Your Open Tasks".

I guess a third option is to place the task in both locations, and disable the state-change controls for the task in "Agency's Remaining Tasks".   That way, the task can be moved along in "Your Open Tasks", and all agency admins retain visibility to the task in "Agency's Remaining Tasks".

I think I like this last option best - but we should discuss at the next standup.

dsmegha:
I now understand the problem it is creating. However I am thinking of this solution. We can have tabs for each of the set of tasks. Everybody who logs in can see only his open tasks once he logs in, but has tabs above which he can click to view his closed tasks. If he is an admin, he will have tabs to view all agency's tasks, agency's unassigned tasks as well as agency's closed tasks. This will some UI change as well, but I think we look much cleaner especially when the number of tasks are high. He will not have to scroll down to reach different set of tasks. 

Something similar to this : https://jqueryui.com/tabs/#ajax

Let me know what you think.

patmbolger:
I like the tab idea, as it will definitely look cleaner.

So, for an admin, I assume we'd see four tabs (?):

1. Unassigned agency tasks
2. Assigned (open) agency tasks (assigned to myself or others)
3. My open tasks (where I can change the state to 'in progress', 'complete', etc.)
4. Closed agency tasks (including my own)

I tab 2, I think we need to make sure that the admin cannot change the state of his own tasks.   That capability is present in tab 3.

Is this consistent with how you are thinking about this?

dsmegha:
Yes, this is consistent with what I am thinking. However do we have to have editing property for an admin only in tab3? He can maybe change the status in both the tabs.

patmbolger:
I think having the ability to change the state in both tabs is OK, as long as the state of the task is consistent in each tab.

dsmegha:
Perfect. I will make the necessary changes

PR #348, opened 2016-08-09
Closes AgileVentures/MetPlus_tracker#372

jeanwei:
Company person Accept Job Application
- add 'accept' link
- add confirmation for accept
- notification event for job developer after accept 

Question:
- In Job applications list, there is a column for 'applied' which refer to 'updated_at'. This info might not be accurate because updating the job application status to 'accepted' will change the 'updated_at'. 
- Notice that in email preview, error happens (looks broken) when a logged-out user click the link to view job application show page. Is it a real bug or can be ignore?
- what is this error refer to in cucumber test? It will disappear by re-running
screen shot 2016-08-09 at 17 19 03

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -6,6 +6,10 @@ class ApplicationController < ActionController::Base
   end
 
   protect_from_forgery with: :exception
+  def redirect_back_or_default(default = root_path)
+    redirect_to (request.referer.present? ? :back : default)
+  end
+  
   helper_method :pets_user, :current_agency, :determine_if_admin
 
   include Pundit
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,6 +30,9 @@ class ApplicationController < ActionController::Base
     stored_location_for(resource) || request.referer || root_path
   end
 
+  def redirect_back_or_default(default = root_path)
+    redirect_to (request.referer.present? ? :back : default)
+  end
 
   protected
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -30,10 +30,6 @@ class ApplicationController < ActionController::Base
     stored_location_for(resource) || request.referer || root_path
   end
 
-  def redirect_back_or_default(default = root_path)
-    redirect_to (request.referer.present? ? :back : default)
-  end
-
   protected
 
     def action_description
--- a/app/controllers/job_applications_controller.rb
+++ b/app/controllers/job_applications_controller.rb
@@ -25,9 +25,5 @@ class JobApplicationsController < ApplicationController
 			redirect_back_or_default
 		end
 	end
-
-	def redirect_back_or_default(default = root_path)
-    redirect_to (request.referer.present? ? :back : default)
-  end
   
 end
--- a/app/controllers/job_applications_controller.rb
+++ b/app/controllers/job_applications_controller.rb
@@ -1,0 +1,28 @@
+class JobApplicationsController < ApplicationController
+	include JobApplicationsViewer
+
+	before_action :find_application
+
+	def accept
+		unless @job_application.active? 
+			flash[:alert] = "Invalid action on inactive job application."
+		else
+			begin
+				@job_application.accept
+				Event.create(:APP_ACCEPTED, @job_application)
+				flash[:info] = "Job application accepted."
+			rescue Exception => e
+				flash[:alert] = "Unable to accept at this moment, please try again."
+			end
+		end
+		redirect_to applications_job_url(@job_application.job)
+	end
+
+	def show
+	end
+
+	def find_application
+		@job_application = JobApplication.find(params[:id])
+		flash[:alert] = "Job Application Entry not found." unless @job_application
+	end
+end
--- a/app/controllers/job_applications_controller.rb
+++ b/app/controllers/job_applications_controller.rb
@@ -7,13 +7,9 @@ class JobApplicationsController < ApplicationController
 		unless @job_application.active? 
 			flash[:alert] = "Invalid action on inactive job application."
 		else
-			begin
-				@job_application.accept
-				Event.create(:APP_ACCEPTED, @job_application)
-				flash[:info] = "Job application accepted."
-			rescue Exception => e
-				flash[:alert] = "Unable to accept at this moment, please try again."
-			end
+			@job_application.accept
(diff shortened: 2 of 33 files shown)


jeanwei:
I split the company_person.feature into 2 files (company_person_jobs_view.feature) because I failed the create job setup for the earlier scenario which do not need job setup. In fact that is only used for 1 scenario

patmbolger:
Re the questions:
1) In Job applications list, there is a column for 'applied' which refer to 'updated_at'. This info might not be accurate because updating the job application status to 'accepted' will change the 'updated_at'.
  - Good point ... we'll probably need to add fields to the DB to timestamp specific application state change(s).

2) Notice that in email preview, error happens (looks broken) when a logged-out user click the link to view job application show page. Is it a real bug or can be ignore?
  - We can ignore this. The email previews are meant to allow you to view how the email works.   Links and other elements in the email are not expected to necessarily work.

3) what is this error refer to in cucumber test?
  - I think that is due to the attempt to create jobs (as specified in the "Background" section of the file.  The create method tries to send the job to the backend "Cruncher".   That app is hosted in a free account in Red Hat's cloud.  Since it is free, it is put to sleep after a period of inactivity.  Bringing it back up probably causes a timeout on the Rails side.  Trying a second time, soon after, succeeds because now the Cruncher is up and running.

app/controllers/job_applications_controller.rb
```diff
@@ -0,0 +1,33 @@
...
+		begin
+			@job_application = JobApplication.find(params[:id])
+		rescue
+			flash[:alert] = "Job Application Entry not found."
+			redirect_back_or_default
+		end
+	end
+
+	def redirect_back_or_default(default = root_path)
+    redirect_to (request.referer.present? ? :back : default)
```
patmbolger: This method is a great idea.

How about adding it to application_controller.rb and making it available to other controllers by adding it to line 9 in that controller?

features/step_definitions/accept_job_applications_steps.rb
```diff
@@ -0,0 +1,93 @@
...
+ 		         		 applicant's name: #{@job_app.job_seeker.full_name(last_name_first: false)}
+                 job title:  #{@job_app.job.title}
+                 company job id: #{@job_app.job.company_job_id}")
+end
+
+And(/^I click the "([^"]*)" confirmation$/) do |action|
+	find("a[href='/job_applications/#{@job_app.id}/#{action.downcase}']").click
+end 
+
+And(/^I should see "3" active applications for the job$/) do
```
patmbolger: Suggest making the application count a paramour rather than a hard-wired number ("3"), and change logic to confirm that this many applications are shown on the page.

features/step_definitions/accept_job_applications_steps.rb
```diff
@@ -0,0 +1,93 @@
...
+	@app_accepted = JobApplication.find_by(job_seeker: job_seeker)
+	within("#applications-#{@app_accepted.id}") { expect(page).to have_content("accepted") }
+end 
+
+And(/^other applications change to not accepted$/) do 
+	@app_ids.each do |app|
+		expect(page.find("#applications-#{app.id}")).
+		to have_content("not_accepted") unless app.id == @app_accepted.id
+	end
+end
```
patmbolger: Suggest moving this immediately below the step defined at line 35, and adding a comment saying that 1) this step can only be used in a scenario that includes the step above this one, and, 2) the step above this one has to occur *before* this step.

(either that or reload @app_ids in this step)

spec/models/event_spec.rb
```diff
@@ -20,10 +20,19 @@
...
   end
+  
+  let(:job_application) do
+    js = FactoryGirl.create(:job_seeker)
+    jd = FactoryGirl.create(:job_developer)
+    jd.agency_relations << FactoryGirl.create(:agency_relation, 
+                           job_seeker: js,
+                           agency_role_id: 1)
+    FactoryGirl.create(:job_application, job_seeker: js, job: job) 
+  end
```
patmbolger: We could simplify to use just one application (var `application`) for all tests by these steps:
1) Remove this statement creating job_application
2) Move let!(:job_developer) before let(:job_seeker) and change let(:job_seeker), so that lines 7-12 look like:
```
let!(:job_developer)  { FactoryGirl.create(:job_developer, agency: agency) }
  let(:job_seeker) do
    js = FactoryGirl.create(:job_seeker)
    js.assign_job_developer(job_developer, agency)
    js
  end
```
3) Change `let(:case_manager)` to `let!(:case_manager)`.  
4) Change test references from var `job_application` to `application`.

patmbolger:
Reviewed PR - looks good and very comprehensive.

Please see comments.

app/controllers/job_applications_controller.rb
jeanwei: why specifically line 9?

spec/models/event_spec.rb
jeanwei: After making the suggested change, made a slight change to line:110 to quiet down this error.
screen shot 2016-08-11 at 12 09 49

jeanwei:
Thanks for the review. Changes updated.

app/controllers/application_controller.rb
```diff
@@ -6,6 +6,10 @@ class ApplicationController::AuthorizationException < Exception
   end
 
   protect_from_forgery with: :exception
+  def redirect_back_or_default(default = root_path)
+    redirect_to (request.referer.present? ? :back : default)
+  end
+  
   helper_method :pets_user, :current_agency, :determine_if_admin
```
patmbolger: Sorry about the confusing reference to line 9 .... what I meant is to add the method name to this line, ala:
```
helper_method :pets_user, :current_agency, :determine_if_admin, redirect_back_or_default
```
That makes the method available to controllers (otherwise, a helper method is only available to views).

patmbolger:
Reviewed edited PR - looks good.

Please see one remaining suggestion.

app/controllers/application_controller.rb
jeanwei: shouln't it be moving the method to application_controller makes the method available to the rest of controllers? Because the rest of the controllers inherit from application_controller. while adding to helper_method makes the method available to views instead?

patmbolger: Yes, you're correct - I wrote that comment too quickly.

No need to change anything then.

joaopapereira:
Going to merge

PR #350, opened 2016-08-12
Closes AgileVentures/MetPlus_tracker#292

sandhyakoti:


app/assets/javascripts/application.js
```diff
@@ -15,6 +15,8 @@
 //= require jquery_ujs
 //= require turbolinks
 //= require js.cookie
+//= require jquery-readyselector
+//= require moment-with-locales.min
```
patmbolger: Can we fetch this via a CDN rather than from our local 'vendor' directory?  A CDN would almost certainly be able to deliver the library to the user's browser faster than our hosting server.

Perhaps one or more of these links could be added to views/layouts/application.html.haml to pull in the correct file(s):  https://cdnjs.com/libraries/moment.js

app/assets/javascripts/utc_to_local.js
```diff
@@ -0,0 +1,3 @@
+$(".job_seekers.edit").ready(function() {
+    $('#utc_to_local_time').text(moment($('#utc_to_local_time')[0].innerHTML).format('LLLL'));
+})
```
patmbolger: 1) Perhaps we could use a class rather than an ID to access the elements that need to be converted.  That way, we don't have to worry about having unique ID's for multiple timestamps on the same page.

For instance, perhaps something like this for the DOM-ready function:

$.each($('.utc_to_local_time'), function(index, value) {
	*convert UTC to local time .....*
});

app/views/job_seekers/_form.html.haml
```diff
@@ -29,10 +29,8 @@
     .clearfix
     .col-sm-8.col-sm-offset-2
       - if @current_resume
-        = "(Your resume on file is \"#{@current_resume.file_name}\",       |
-                 and was uploaded on #{@current_resume.updated_at.         |
-                in_time_zone('Eastern Time (US & Canada)').to_s(:long)})"  |
-
+        = "Your resume on file is \"#{@current_resume.file_name}\", and was uploaded on"
+        %span#utc_to_local_time #{@current_resume.updated_at}
```
patmbolger: I am trying to test if this works and cannot show that it does work(?).

I am testing with the seeded job seeker, "Tom Seeker" (email: tom@gmail.com):
```
js = JobSeeker.where(email: 'tom@gmail.com')[0]

js.resumes[0].updated_at
  Resume Load (0.7ms)  SELECT "resumes".* FROM "resumes" WHERE "resumes"."job_seeker_id" = ?  [["job_seeker_id", 201]]
 => Fri, 12 Aug 2016 20:14:26 UTC +00:00 
```
This is what I see in the form view:
screen shot 2016-08-14 at 4 56 44 pm

There is no conversion to my local timezone (UTC-4 (4:14 PM)).

patmbolger:

Also, please add a follow-on story to add Jasmine test(s) for this functionality (those tests don't have to be part of this story). (Sandy does not necessarily have to implement the test story .... that is open to anyone who wants to take it (after voting, of course)).

joaopapereira:
Do we have any follow-up in this story?

patmbolger:
I got an 'Invalid time' error when I tested this locally.  It seems that the moment() function did not like a time string that included ' UTC' at the end.

 To get it to work I made these changes:

1) In `layouts/application.html.haml` - added this:
```
    = javascript_include_tag 'https://cdnjs.cloudflare.com/ajax/libs/moment-timezone/0.5.7/moment-timezone-with-data.min.js'
```

2) Changed `utc_to_local.js` to this:
```
$(document).ready(function () {
    $.each($('.utc_to_local_time'), function (index, value) {
        utc_time = moment.tz(value.innerHTML.replace(/ UTC/, ''), 'UTC')
        $(value).text(moment(utc_time).tz(moment.tz.guess()).format('LLL'));
    })
});
```
(note that I changed the format to not include the day of the week .... not sure we need detail)

joaopapereira:
Closing this one that no longer make sense

PR #353, opened 2016-08-16
Closes AgileVentures/MetPlus_tracker#400

patmbolger:
# Story:
Currently we are showing timestamps for entity states based solely on the 'updated_at' timestamp for the entity. 

For instance, this code segment (from views/jobs/_job_applications.html.haml) uses the 'updated_at' field to display the time when a job application was submitted:

```
  %table.table.table-hover
    %thead
      %tr
        - application_fields(@application_type).each do |field|
          - case field
            - when :name
              %th.strong Name
            - when :js_status
              %th.strong Job Seeker Status
            - when :updated_at
              %th.strong Applied
            - when :action
              %th.strong Action
```
If the application status changes to 'accepted', or 'not_accepted', this view will show incorrect information.

This story is to design and implement a general way to track status changes for any entity with a simple status change structure (that is, when we only need to track 1) the state transitioned to (integer) and 2) the time when that transition occurred).  We get the second one for free from Rails (created_at).

# Implementation

- [ ] Design a model that will provide the structure and appropriate methods for managing state change, for example:
```
class StatusChange < ActiveRecord::Base
  belongs_to :entity, polymorphic: true

  validates_presence_of status_change_to
  validates_presence_of status_change_from, allow_nil: true
  .....
```
- [ ] Migration to add to DB.  **NOTE** - you can create the model file and the DB migration with this one command:
```
rails g model StatusChange status_change_to:integer   status_change_from:integer   entity:references{polymorphic}
```
- [ ] Rspec tests

NOTE: An entity for which this functionality is used must use an *enum* field to track status.  For instance, from the JobApplication model:
```
enum status: [:active, :accepted, :not_accepted]
```
This allows some tolerance for change in how status values are displayed to the user.  For instance, if we decide to change the first status for JobApplication from :active to :Active, this will not affect the integrity of the information maintained by StateChange.

- [ ] Add has_many status_changes association to JobApplication
  - for testing purposes
  - this model will need to be augmented to use StatusChange in order to address flaw(s) in view(s) (like the one described above)

### DIFF
--- a/app/models/status_change.rb
+++ b/app/models/status_change.rb
@@ -19,16 +19,20 @@ class StatusChange < ActiveRecord::Base
                      status_change_to: entity.class.statuses[to_status])
   end
 
-  def self.status_change_time(entity, to_status)
-    # Returns time when entity transitioned to specific status.
-    # Returns nil if to_status is not found.
-    # NOTE: this assumes that an entity assume the specified status
-    #       only once in its life cycle.  If this changes, this code
-    #       must be changed as well.
+  def self.status_change_time(entity, to_status, which = :latest)
+    # Returns time(s) when entity transitioned to specific status.
+    # If 'which' == :latest, returns latest time that entity status == to_status
+    # If 'which' == :all, returns an array of times that entity status
+    #             == to_status (in ascending order)
+    # Returns nil if to_status is not found
+    # Raises exception if 'which' is invalid
 
-    change = entity.status_changes.
-        where(status_change_to: entity.class.statuses[to_status])[0]
-    return change.created_at if change
-    nil
+    change_times = entity.status_changes.
+        where(status_change_to: entity.class.statuses[to_status]).
+              order(:created_at).pluck(:created_at)
+
+    return change_times.last if which == :latest
+    return change_times      if which == :all
+    raise ArgumentError.new("Invalid 'which' argument")
   end
 end
--- a/app/models/status_change.rb
+++ b/app/models/status_change.rb
@@ -1,0 +1,34 @@
+class StatusChange < ActiveRecord::Base
+  belongs_to :entity, polymorphic: true
+
+  validates_presence_of :status_change_to
+  validates_numericality_of :status_change_to, only_integer: true,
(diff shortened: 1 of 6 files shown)


app/models/status_change.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  def self.status_change_time(entity, to_status)
+    # Returns time when entity transitioned to specific status.
+    # Returns nil if to_status is not found.
+    # NOTE: this assumes that an entity assume the specified status
+    #       only once in its life cycle.  If this changes, this code
+    #       must be changed as well.
+
+    change = entity.status_changes.
+        where(status_change_to: entity.class.statuses[to_status])[0]
```
joaopapereira: You are missing the order here, to make sure that you are getting the latest change right?

spec/models/status_change_spec.rb
```diff
@@ -0,0 +1,45 @@
...
+  describe 'Class methods' do
+    let(:ja) { FactoryGirl.create(:job_application) }
+
+    it 'adds a status change record for an entity' do
+      expect{ StatusChange.update_status_history(ja, nil, :active) }.
+            to change(StatusChange, :count).by 1
+      expect(ja.status_changes.count).to eq 1
+    end
+
+    it 'returns status change time for an entity' do
```
joaopapereira: Add another test to make sure that you return the latest change to that state

joaopapereira:

app/models/status_change.rb
patmbolger: This code is based on the assumption that any specific status for the entity occurs just once.  If this changes we'll have to alter this method to accommodate that.

The timestamp for the specified status change does not need to be the timestamp of the most recent status change.

joaopapereira: There is a story AgileVentures/Metplus_track#399 to be able to re-apply for a job. So we need to look into this, because if we are retrieving the first date that we find in the table, doesn't mean it is the correct one.

patmbolger:
Adjusted PR per review comments.

joaopapereira:
Looks good going to merge

PR #354, opened 2016-08-18
Closes AgileVentures/MetPlus_tracker#403

patmbolger:
Improve exception message for job create failure when cause if failure to create a job in the Cruncher (current exception message is not clear as to whether 1) Job model is invalid or 2) Cruncher job create is the cause of the Job save exception).


joaopapereira:
Looks good, going to merge

PR #356, opened 2016-08-24
Closes AgileVentures/MetPlus_tracker#401

patmbolger:
# Story:

As a developer
I want to track status changes for a job application
So that I can accurately display status change times in the UI

# Implementation:

This story includes:

- [ ] Implement tracking of job application status changes via the StatusChange model.

- [ ] Check each view where job application(s) is shown, and correct the display of status change timestamp as needed (see example described in #400).

**NOTE**: Also added a wait before an unrelated cuke test (event mgmt) that causes intermittent and frequent failure in CI testing.


joaopapereira:
Looks good, going to merge

PR #358, opened 2016-08-26
Closes AgileVentures/MetPlus_tracker#402

patmbolger:
# Story

As a developer
I want Job status attributes to be enums
So that I have a common mechanism with other entities
And can leverage common helper code

# Implementation

This story includes:

- [ ] Convert Job instance status structure from a hash to enum

- [ ] Add status change updates, via StatusChange, at all places in the code where such status changes occur

- [ ] Check all views where job(s) attributes are displayed and determine if any of these need to be changed to use the status change tracking mechanism (see example of a view that needs to be updated for change tracking (for job application in that case))

Note: Also changed *StatusChange.update_status_history(entity, from_status to_status)* to *StatusChange.update_status_history(entity, to_status)*.

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -176,7 +176,7 @@ class JobsController < ApplicationController
 	end
 
 	def revoke
-		if @job.status == 'active' && @job.update(status: 'revoked')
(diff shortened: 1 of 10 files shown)


joaopapereira:
Looks good, going to merge it

PR #360, opened 2016-08-27
Closes AgileVentures/MetPlus_tracker#40

jeanwei:
Feature: Job Developer submit job application for his job seeker
- Link to application form is created at views/jobs/show
- Dropdown select with his list of job seekers is created in view/job_applications/_jd_apply_job_modal
- Preview for the application is created in view/job_applications/_preview_modal
- Controllers, routes and spec files are updated accordingly

Changes indirectly related to feature developed:
- job_seeker_email_job.rb, app/mailers/job_seeker_mailer.rb are updated for more general use. Please review if there will be any problem?
- job.rb are updated so that #apply method return the job_application because notice that always need to call for the job application after job is applied.

### DIFF
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -51,7 +51,7 @@ var PusherControl = {
       // Process event if I am logged in and I am in notify list
       if (parseInt(Cookies('user_id')) === data.js_user_id) {
         Notification.info_notification('Your job developer ' +
-            " has applied <a href='/jobs/" + data.job_id +
+            " has applied to <a href='/jobs/" + data.job_id +
             "' target='_blank'>this job</a>" + ' for you.');
       }
     });
--- a/app/assets/javascripts/pusher.js.erb
+++ b/app/assets/javascripts/pusher.js.erb
@@ -47,6 +47,15 @@ var PusherControl = {
       }
     });
 
+    channel.bind('job_applied_by_job_developer', function(data) {
+      // Process event if I am logged in and I am in notify list
+      if (parseInt(Cookies('user_id')) === data.js_user_id) {
+        Notification.info_notification('Your job developer ' +
+            " has applied <a href='/jobs/" + data.job_id +
+            "' target='_blank'>this job</a>" + ' for you.');
+      }
+    });
+
     channel.bind('job_application_accepted', function(data) {
       if (parseInt(Cookies('user_id')) === data.jd_user_id) {
         Notification.info_notification("Job Application: " +
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -105,7 +105,8 @@ class JobSeekersController < ApplicationController
   def preview_info
     raise 'Unsupported request' if not request.xhr?
     @jobseeker = JobSeeker.find(params[:id])
-    render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker }
+    render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker,
+                                                   preview_mode: true }
   end
 
   def applied_jobs
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -102,6 +102,12 @@ class JobSeekersController < ApplicationController
     @application_type = params[:application_type] || 'js-applied'
   end
 
+  def preview_info
+    raise 'Unsupported request' if not request.xhr?
+    @jobseeker = JobSeeker.find(params[:id])
+    render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker }
+  end
+
   def applied_jobs
     raise 'Unsupported request' if not request.xhr?
 
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -166,13 +166,25 @@ class JobsController < ApplicationController
 			redirect_to job_path(@job)
 			return
 		end
-		begin
-			@job.apply @job_seeker
-			Event.create(:JS_APPLY, @job.last_application_by_job_seeker(@job_seeker))
-		rescue Exception => e
-			flash[:alert] = "Unable to apply at this moment, please try again."
+
+		if pets_user == @job_seeker
+			begin
+				@job.apply @job_seeker
+				Event.create(:JS_APPLY, @job.last_application_by_job_seeker(@job_seeker))
+			rescue Exception => e
+				flash[:alert] = "Unable to apply at this moment, please try again."
+				redirect_to job_path(@job)
+			end
+		elsif pets_user == @job_seeker.job_developer
+			job_app = @job.apply @job_seeker
+			Event.create(:JD_APPLY, job_app)
+			flash[:info] = "Job is successfully applied for #{@job_seeker.full_name}"
+			redirect_to job_path(@job)
+		else
+			flash[:alert] = "Invalid application for not your job seekers"
 			redirect_to job_path(@job)
 		end
+	
 	end
 
 	def revoke
@@ -185,7 +197,6 @@ class JobsController < ApplicationController
 		end
 		redirect_to jobs_path
 	end
-
 
 	private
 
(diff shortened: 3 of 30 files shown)


app/assets/javascripts/pusher.js.erb
```diff
@@ -47,6 +47,15 @@ var PusherControl = {
       }
     });
 
+    channel.bind('job_applied_by_job_developer', function(data) {
+      // Process event if I am logged in and I am in notify list
+      if (parseInt(Cookies('user_id')) === data.js_user_id) {
+        Notification.info_notification('Your job developer ' +
+            " has applied <a href='/jobs/" + data.job_id +
```
patmbolger:
```
" has applied to <a href='/jobs/" + data.job_id +
```

app/controllers/job_seekers_controller.rb
```diff
@@ -102,6 +102,12 @@ def show
     @application_type = params[:application_type] || 'js-applied'
   end
 
+  def preview_info
+    raise 'Unsupported request' if not request.xhr?
+    @jobseeker = JobSeeker.find(params[:id])
+    render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker }
```
patmbolger: This partial will include an "Assign Myself" button if 1) the job seeker does not have a case manager assigned, and 2) the user is (also) a case manager:

screen shot 2016-08-27 at 2 46 51 pm

That button should not be there as it results in an XHR call to the server, which will not work since this is a modal dialog.

app/controllers/jobs_controller.rb
```diff
@@ -166,13 +166,25 @@ def apply
...
+				flash[:alert] = "Unable to apply at this moment, please try again."
+				redirect_to job_path(@job)
+			end
+		elsif pets_user == @job_seeker.job_developer
+			job_app = @job.apply @job_seeker
+			Event.create(:JD_APPLY, job_app)
+			flash[:info] = "Job is successfully applied for #{@job_seeker.full_name}"
+			redirect_to job_path(@job)
+		else
+			flash[:alert] = "Invalid application for not your job seekers"
```
patmbolger: Suggested wording:
```
flash[:alert] = "Invalid application: You are not the Job Developer for this job seeker"
```

app/mailers/job_seeker_mailer.rb
```diff
@@ -7,20 +7,27 @@ class JobSeekerMailer < ApplicationMailer
...
 
   private
 
-  def send_job_seeker_mail(job_seeker, agency_person, person_type)
-    @job_seeker    = job_seeker
-    @agency_person = agency_person
-    @person_type   = person_type
-    @agency        = agency_person.agency
-    mail to: job_seeker.email, template_name: 'agency_person_assigned'
+  def send_job_seeker_mail(options = {})
```
patmbolger: I like the change to use an options hash!

app/mailers/job_seeker_mailer.rb
```diff
@@ -7,20 +7,27 @@ class JobSeekerMailer < ApplicationMailer
...
+    # user template name as stated or the previous method name
+    template = options.delete(:template) || caller_locations(1,1)[0].label
```
patmbolger: This obscures how the template name was arrived at (if not an explicit option) and is prone to error since the developer has to confirm (and maintain) that the method name is the same as the mailer template name.  Looking into the call stack might be slower as well.  I suggest always passing in a template name to the method.

(I like the ingenuity of the idea, however - just don't think its appropriate for this application)

app/models/event.rb
```diff
@@ -142,6 +145,25 @@ def self.evt_js_apply(evt_obj)  # evt_obj = job application
     Task.new_review_job_application_task(evt_obj.job, evt_obj.job.company)
   end
 
+  def self.evt_jd_apply(evt_obj)  # evt_obj = job_app
+    # Job is applied by job developer for his job seeker
+    # Business rules:
+    #    Notify job seeker
+    #    Create task for 'job application' (application to be reviewed)
```
patmbolger: When a JS applies directly, we notify the company contact associated with the job (if there is one).

I think we should do the same here as well.

app/models/job.rb
```diff
@@ -44,6 +44,7 @@ def number_applicants
   def apply job_seeker
     job_seekers << job_seeker
     save!
+    last_application_by_job_seeker(job_seeker)
   end
```
patmbolger: This is probably a good idea - it makes sense to return the job application that was just submitted for the job.

Please change existing references to reflect this change.  In particular:
- agency_mailer_spec.rb
- agency_mailer_preview.rb
- event_spec.rb
- job_spec.rb

app/views/job_seeker_mailer/job_applied_by_job_developer.html.haml
```diff
@@ -0,0 +1,22 @@
...
+  This is to let you know that
+  %strong
+    your job developer,
+    = mail_to(@job_developer.email,
+      "#{@job_developer.full_name(last_name_first: false)}",
+      subject: "My Job Application")
+    has applied for you the job: 
+    = link_to @job.title, job_url(id: @job.id)
+    , in company: #{@job.company.name}
+%p
```
patmbolger: In order to improve readability I suggest this for lines 9-17:
```
	This is to let you know that
	  %strong
	    your job developer, &nbsp
	    = mail_to(@job_developer.email,
	      "#{@job_developer.full_name(last_name_first: false)}",
	      subject: "My Job Application")
	    , &nbsp has submitted an application on your behalf to the job: &nbsp
	    = link_to @job.title, job_url(id: @job.id)
	    , &nbsp in company: #{@job.company.name}
```

features/job_application_by_jd.feature
```diff
@@ -0,0 +1,99 @@
+Feature: Job developer submit application for his job seeker
+
+As a job developer
+I want to apply a job for my job seeker
```
patmbolger: I want to apply **to** a job for my job seeker

spec/controllers/jobs_controller_spec.rb
```diff
@@ -814,6 +814,45 @@
...
+      before :each do
+        agency = FactoryGirl.create(:agency)
+        job_developer = FactoryGirl.create(:job_developer, agency: agency)
+        job_seeker.assign_job_developer(job_developer, agency)
+        allow(Pusher).to receive(:trigger)
+        allow(Event).to receive(:create).and_call_original
+        allow(controller).to receive(:current_user).and_return(job_developer)
+        get :apply, :job_id => @job.id, :user_id => job_seeker.id
+      end
+      it 'job applicantion is created' do
```
patmbolger: 'job application is created'

patmbolger:
Reviewed PR - great work, and very comprehensive.

Please see my code comments.

jeanwei:
Thanks for the review. Codes are updated. Please review.

features/job_application_by_jd.feature
```diff
@@ -0,0 +1,105 @@
...
+    When "john.seeker@places.com" opens the email
+    Then they should see "your job developer," in the email body
+    Then they should see "Jane Jones" in the email body
+    Then they should see "has submitted an application on your behalf to the job:" in the email body
+    Then they should see "software developer" in the email body
+    Then they should see "in company: Widgets Inc." in the email body
+
+    Then I am in Company Admin's browser
+    And I should see "Job Seeker: John Seeker has applied to this job"
+    And I am on the Company Person 'ca@widgets.com' Home page
```
jeanwei: Currently this line indicates a page refresh to show updated data. Do we want to implement an auto refresh on paginate data? this would be sending xhr request on an interval basis i guess or any other way?

patmbolger: This is a good point.  

I don't think we want to periodically poll for updated pagination data.  

One way we might be able to do this is forcing a refresh of all paginated pages (currently open) via something like this:
```
$('.pagination-div li.active a').click()
```
This will simulate a click on the current pagination link(s), which will then force a reload of the data on that page.

This could be called from within the `channel.bind` handlers in pusher.js.erb, when specific events require a data reload.  

There may be other, better ways to do this as well.

However, we don't need this for MVP.  I recommend adding a story to the backlog for this idea.

patmbolger:
Reviewed PR again - changes look great and I think this is ready to be merged.

joaopapereira:
Looks good, Rebuilding now to see if the tests pass if so I will merge

PR #367, opened 2016-09-10
Closes AgileVentures/MetPlus_tracker#342

dsmegha:
I have changed the home page of an agency person to include tabs for the jobs he can see.

Am admin have four tabs, Unassigned Agency tasks, His open tasks, agency's open tasks and agency's closed tasks.

A case manager/job developer can only his open tasks and his closed tasks.

I have changed the views and also some scopes. Also made sure when a change is made to a task, it is reflected across all the tabs.

I have changed the tests accordingly and removed two tests that are checking for incorrect behaviour.


app/models/task.rb
```diff
@@ -45,28 +47,46 @@ def task_owner=(user: nil, agency: {agency: nil, role: nil}, company: {company:
...
 
   def self.find_by_owner_user_closed user
     closed_tasks.user_tasks(user)
   end
 
+  def self.find_by_agency user
+    today_tasks.agency_tasks(user)
+  end
+
+  def self.find_by_agency_new user
```
joaopapereira: If is this simple I believe we do not need to have a function for it as we are using the scope only, but this is not a blocker just a tip

joaopapereira:
Looks good, keep in mind my comment for further changes. Lets see if @patmbolger is ok with the change

app/controllers/agency_people_controller.rb
```diff
@@ -15,12 +15,15 @@ def home
...
+    @agency_new = 'agency-new'
+    @agency_closed = 'agency-closed'
     @people_type_without_jd = 'jobseeker-without-jd'
     @people_type_without_cm = 'jobseeker-without-cm'
-    @js_without_jd = JobSeeker.js_without_jd
-    @js_without_cm = JobSeeker.js_without_cm
+    @people_type_cm = 'jobseeker-cm'
+    @people_type_jd = 'jobseeker-jd'
+    @js_without_jd = JobSeeker.paginate(:page=> params[:js_without_jd_page], :per_page=>5).js_without_jd
+    @js_without_cm = JobSeeker.paginate(:page=> params[:js_without_cm_page], :per_page=>5).js_without_cm
```
patmbolger: Have you rebased against upstream development since creating this PR?

  The two lines above look suspect - we dropped the `.paginate` method call in these statements some time ago.

features/step_definitions/tasks_steps.rb
```diff
@@ -15,6 +15,19 @@
   end
 end
 
+Then(/^The (?:task|tasks) ((\d+,?)+) (?:is|are)( not)? hidden/) do |tasks, _, not_see|
```
patmbolger: I suggest changing `not_see` to `should_see`.  If 'not' is present, then the operative phrase is 'not hidden', which translates to "I *should see* this element".

features/step_definitions/tasks_steps.rb
```diff
@@ -15,6 +15,19 @@
...
+  tasks.split(/,/).each do |task_id|
+    error = false
+    begin
+      find("#task-#{task_id}", visible:false)
+      error = true if not_see
+    rescue Exception => e
+      raise "Task with id #{task_id} could not be found" unless not_see
```
patmbolger: I think this logic might be wrong.

Line 22 will raise an exception if an invisible element with this ID is not found.

However, this does not necessarily mean that a *visible* element with this ID is actually on the page.  This logic, however, assumes that.

That is, lets assume we expect that element 'A' is on the page, and is visible.  However, lets also assume the 'A' is not actually on the page.   In this case, line 22 raises an exception.  But that line 25 negates that exception simply because 'A' is *supposed* to be visible.

app/views/agency_people/home.html.haml
```diff
@@ -5,8 +5,35 @@
...
+    %li.active(role="presentation")
+      %a(href="#tab2" aria-controls="tab2" role="tab" data-toggle="tab")
+        Your Open Tasks
+    - if @agency_person.is_agency_admin?(@agency)
+      %li(role="presentation")
+        %a(href="#tab3" aria-controls="tab3" role="tab" data-toggle="tab")
+          All Agency Open Tasks
+    %li(role="presentation")
+      %a(href="#tab4" aria-controls="tab4" role="tab" data-toggle="tab")
+        Closed Tasks
```
patmbolger: I know the code might be messy, but I think it would good for this tab to be name:

`All Agency Closed Tasks` if user == agency admin

`Your Closed Tasks` otherwise

patmbolger:
Reviewed PR ... please see comments.

In general, this looks great for agency people (admin and other roles).

For company people, however, we will need to enhance what we currently have. 

The current company person home page (for a company admin) has this:

screen shot 2016-09-11 at 6 49 03 pm

Also, a company person should be able to assign a 'review job application' task to anyone else in the company.  However, this is what appears when attempting to do so:

screen shot 2016-09-11 at 6 51 23 pm

So, we need to:

1. Convert to the tabbed format.  For company admin there would be 4 tabs (ala agency admin) and for company contact there would be two tabs (ala job developer).

2. Allow a company admin to assign tasks to other company people (e.g., review job application) 

3. Confirm that tasks show up in the appropriate tabs as they undergo state changes, for both company admin and (assigned) company contact.

I think that we should add a separate story for the company person task features above, and merge this PR (when ready) as-is.  (I will create that separate story).

patmbolger:
Forgot to add .... this is great work on a very complex story.

dsmegha:
I have made the changes in the test case as Patrick mentioned. I realised we only need a step to verify we have hidden elements as the other case is already taken care.

patmbolger:
Reviewed revised PR - looks good to merge.

I think the tabbed interface is a great improvement in the UI.  We may want to extend the use of that in other areas, to help address Sam's observation that the user has to sometimes scroll down too much to find things.

PR #368, opened 2016-09-10
Phone number prefix#413

dsmegha:
Changed the validator to accept a phone number with a prefix "1" or "1 " or "1-"


app/validators/phone_validator.rb
```diff
@@ -1,7 +1,7 @@
 class PhoneValidator < ActiveModel::EachValidator
   def validate_each(object, attribute, value)
     return if value == nil
-    reg = /^(1-)?((\(\d{3}\))|\d{3})( |-)?\d{3}( |-)?\d{4}( ?x\d+)?$/
+    reg = /^(1-|1( )*)?((\(\d{3}\))|\d{3})( |-)?\d{3}( |-)?\d{4}( ?x\d+)?$/
```
joaopapereira: With this regex we accept the number:
1               222 222 2222
maybe instead of 
(1-|1( )*)?
we should have
(1-|1( )?)?

spec/models/user_spec.rb
```diff
@@ -51,10 +51,11 @@
 
      describe 'Phone number format check' do
        subject {FactoryGirl.build(:user)}
-       it { should_not allow_value('asd', '123456', '123 123 12345', '123 1231 1234', '1123 123 1234', ' 123 123 1234').for(:phone)}
-
-       it { should allow_value('123 123 1234', '(123) 123 1234', '(123)-123 1234', '1231231234', '(123) 1231234').for(:phone)}
-
+       it { should_not allow_value('+1 123 123 1234', 'asd', '123456', '123 1231  1234',                
+               ' 123 123 1234', '(234 1234 1234', '786) 1243 3578').for(:phone)}
+    it { should allow_value('123 123 1234', '(123) 123 1234', '1231231234',
```
joaopapereira: Check the indentation of the line

joaopapereira:

dsmegha:
I have made the change that Joao has mentioned. It will be great if the reviewer can double check the valid and invalid phone/fax numbers once since I am not very sure of the American phone number system. We had "+1 123 123 1234" as an invalid number, I added it to the valid set and changed the regex to accept that.

patmbolger:
This looks good to me.  We can always adjust as needed if we run into other use patterns once in production, but I would go with this for now.

joaopapereira:
Looks good going to merge

PR #370, opened 2016-09-11
Closes AgileVentures/MetPlus_tracker#398

arun1595:
Display the status of a job and the time of the status change in `company_person` home page

Also, modified the date format of last update field as a part of this PR since the change was trivial.

### DIFF
--- a/app/models/job.rb
+++ b/app/models/job.rb
@@ -34,8 +34,12 @@ class Job < ActiveRecord::Base
   enum status: [:active, :filled, :revoked]
   has_many :status_changes, as: :entity, dependent: :destroy
 
+  # `self.status.to_sym` is necessary because sometimes we would need to create
+  #  a job with a status other than `active`(in tests) and we don't want
+  #  to have a mismatch in statuses across `jobs` and `status_changes` tables.
+
   after_create do
-    StatusChange.update_status_history(self, :active)
+    StatusChange.update_status_history(self, self.status.to_sym || :active)
   end
 
   def number_applicants
--- a/app/views/jobs/_list_all.html.haml
+++ b/app/views/jobs/_list_all.html.haml
@@ -5,6 +5,8 @@
         - case field
           - when :title
             %th.strong Title
+          - when :status
+            %th Status
           - when :updated_at
             %th.strong Last update
           - when :poster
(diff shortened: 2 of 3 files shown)


joaopapereira:
Looks good, lets see if @patmbolger as anything to add.
If now I will merge it

app/views/jobs/_list_all.html.haml
```diff
@@ -19,15 +21,22 @@
           - when :title
             %td.col-md-4
               = link_to(job.title, job_path(job))
+          - when :status
+            %td.col-md-2
+              = job.status.capitalize
+              &nbsp;
+              %span.glyphicon.glyphicon-time{ 'aria-hidden': true }
+              %span.text-muted.small
+                = job.created_at.strftime('%d %b, %Y')
```
patmbolger: What we want here is the time of change to the latest status for the job, *not* the time when the job itself was created.  (for an `:active` job, this *might* be the same as the job creation time, but not necessarily - in the future we might allow a job to go from `:revoked` back to `:active`, for example).

We get this time from:
```
StatusChange.status_change_time(job, job.status)
```

patmbolger:
Reviewed PR .... please see comment.

app/views/jobs/_list_all.html.haml
patmbolger: Also, I really like the timestamp with the clock icon!

arun1595: Hmm. My bad. Actually, I remember using `StatusChange`'s class method. Maybe I was careless. I'll update the PR with the fix.

arun1595:
I have updated `after_create` in `job.rb` to create a `status_change` record with the
right status.
Previously we were storing a `status_change_to` of `active` in the `status_changes` table regardless of the job status when its created.
(see background steps for creating jobs in `manage_job_status.feature`)

joaopapereira:
I am ok with that last change

app/models/job.rb
```diff
@@ -35,7 +35,7 @@ class Job < ActiveRecord::Base
   has_many :status_changes, as: :entity, dependent: :destroy
 
   after_create do
-    StatusChange.update_status_history(self, :active)
+    StatusChange.update_status_history(self, self.status.to_sym || :active)
```
patmbolger: What is the rationale for this?  We are *creating* the job instance, so by definition there is no prior state (or, to be more precise, the previous state is nil, which is what StatusChange sets it to).

arun1595: Yeah, I agree but in `manage_job_status.feature`, we are creating a job with a `status` of `revoked`.
```
| title | company_job_id | shift | fulltime | description | company | creator | status  |
| hr assistant | KRK01K  | Day | true | internship  | Widgets Inc. | ca@widgets.com | revoked |
```
and we have a `after_callback` in `job.rb` which creates a `StatusChange` record with `status_change_to` to `active` eventhough the created job has a status of `revoked`. 

```ruby
after_create do
    StatusChange.update_status_history(self, :active)
  end
```

So, we have a job with a status of `revoked` in jobs table but a `status_change_to` of `active` in `status_changes` table.
Because of this, I am unable to display the `job`'s status with the change_time since there's no such record in the `status_changes` table. This line is to fix that issue.

patmbolger: I see.  We could add logic to the fixture step to first create a job and then advance the status if initial status != :active, but that would be cumbersome.  

Thanks for the explanation.

You might want to put a comment above the revised code explaining that - so someone looking at the code in 6 months doesn't say "this is not needed, I will fix ......".

joaopapereira:
Looks good, going to merge

PR #371, opened 2016-09-11
Closes AgileVentures/MetPlus_tracker#416

patmbolger:
Also, removed old test code that included `admin_type` params key - which is no longer used.


joaopapereira:
Looks good, going to merge

PR #373, opened 2016-09-13
Closes AgileVentures/MetPlus_tracker#399

jeanwei:
Recreate PR for #399 due to merge conflict. Please review


joaopapereira:
Are you using tabs to indent? if so please change it to 2 spaces

joaopapereira:
Generally looks good, do you think we should have a Cucumber test to ensure that everything is working?
Like a user trying to apply twice to the same job?

jeanwei:
Thanks for the review. code has updated as suggested. Please review if there is anything missing.

patmbolger:
This is good work and worth keeping - but it does not fully meet the intent of the story - which is to detect an attempt to reapply to a job and popup a dialog box explaining why this is not possible (and not even useful, actually).  

The story says:

- [ ] Determine if job seeker has already applied to this job (after the current check to confirm that the job is 'active', and the job seeker is found in the DB)

- [ ] If so, and that job application is 'active', show a dialog box to the user, explaining that:
  - "You already have an application submitted for this job."
  - "Your job application is still active."
  - "If you want the employer to see a revised version of your résumé, simply edit your profile by clicking on this icon: (*show edit icon here*) on your home page, and upload the revised résumé."

We're not helping the job seeker by making them review the apply confirmation and hit 'Apply", and then immediately tell them that they had already applied.  ("this app should have known that before it let me try to reapply").

The dialog box text is important in 1) explaining to the job seeker why they should not worry about reapplying since 1) their application is still 'active', and 2) if they want to make sure a revised résumé is seen by the employer, they just need to upload the revised résumé in their profile.

patmbolger:
Also, in checking for whether we allow reapply - we don't allow if the job seeker's application is 'active'.  If it is of any other status (such as 'not_accepted') we should allow the job seeker to submit a new application for the same job.

jeanwei:
Is it going to change the company person's decision with multiple application from the same job seeker? Company person do feel irritated with overwhelming applications such that they will post message "Do not apply more than 1 job from the company. We will arrange it accordingly if there is other job that fit your resume" to job seekers. In addition, we do not know what we are going to do with the rejected job application yet- which might be cleaned up. Then the dialog might not be relevant. Usually, if a job is still vacant after a while, the company person will repost the job so it appear new. In this scenario, the job_id will be unique. the job seeker can re-apply to the same job w/o problem. As for the update resume, we can add it in the flash message if the applicant is a job seeker? Just my thoughts, feel free to comment. =)

jeanwei:
a panel to inform job seeker on applied job is added in app/views/jobs/show.html.haml. 
feature file has updated accordingly.

PR #374, opened 2016-09-13
Closes AgileVentures/MetPlus_tracker#419

patmbolger:
# Story:

The cruncher API endpoint has been changed to accommodate matching of a résumé to jobs, and provide a matching score (1 to 5) for each matched job.

The cruncher API endpoint has been changed to accommodate matching of a job to résumés, and provide a matching score (1 to 5) for each matched résumé.

# Implementation:

- [ ] In `cruncher_service.rb` update these methods:
  - [ ] `self.match_jobs(resume_id)`
  - [ ] `self.match_resumes(job_id)`

- [ ] In `job_cruncher.rb`:
  - [ ] Change comments to explain to the developer the return value
  - [ ] Call utility method to process match results

- [ ] In `resume_cruncher.rb`:
  - [ ] Change comments to explain to the developer the return value
  - [ ] Call utility method to process match results

- [ ] Add `models/concerns/cruncher_utility.rb`
  - [ ] Add class method to process match results and return array of ordered matches

- [ ] Add/update rspec tests as needed

**Note** - also updated `resume_spec.rb` with stub methods, replacing in-place stubs.

### DIFF
--- a/spec/models/concerns/cruncher_utility_spec.rb
+++ b/spec/models/concerns/cruncher_utility_spec.rb
@@ -20,14 +20,14 @@ RSpec.describe CruncherUtility do
                                        {"resumeId"=>"7", "stars"=>1.7}]} }
 
   it 'processes job match results' do
-    processed_results = UtilityTester.process_match_results(job_results)
+    processed_results = UtilityTester.process_match_results(job_results, 'jobId')
     expect(processed_results.length).to be 4
     expect(processed_results).
       to eq [ [2, 5.5], [4, 4.4], [3, 3.8], [1, 1.0] ]
   end
 
   it 'processes resume match results' do
-    processed_results = UtilityTester.process_match_results(resume_results)
+    processed_results = UtilityTester.process_match_results(resume_results, 'resumeId')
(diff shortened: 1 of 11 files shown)


spec/models/concerns/cruncher_utility_spec.rb
```diff
@@ -0,0 +1,36 @@
...
+end
+
+RSpec.describe CruncherUtility do
+
+  let(:job_results) { {"matcher1"=>[{"jobId"=>"3", "stars"=>3.3},
+                                    {"jobId"=>"2", "stars"=>2.2},
+                                    {"jobId"=>"1", "stars"=>1.0}],
+                       "matcher2"=>[{"jobId"=>"4", "stars"=>4.4},
+                                    {"jobId"=>"3", "stars"=>3.8},
+                                    {"jobId"=>"2", "stars"=>5.5}]} }
```
joaopapereira: For accuracy sake we should use only numbers from 0 - 5

patmbolger: Good point .... changed accordingly.

PR #375, opened 2016-09-14
Closes AgileVentures/MetPlus_tracker#418

jeanwei:
When Job developer apply to a job on  behalf of his job seekers
 - Job seeker without resume is greyed out in the select option
 - His job seekers are listed in alphabetical order
 - "The Job Seekers not selectable do not have a resume on file." is shown under the select box

RSpec and cucumber test are updated accordingly

job_application.js file is updated because previous one return url: "/agency_people/undefined/my_js_as_jd" instead of  "/agency_people/2/my_js_as_jd"

### DIFF
--- a/app/assets/javascripts/job_applications.js
+++ b/app/assets/javascripts/job_applications.js
@@ -10,13 +10,13 @@ var dataHandler = {
     return this;
   },
 
-  load_form: function () {
+  load_form: function (jd_id) {
     $('.alert_msg').html('');
     $('#jd_apply_job_select').val('');
     $("#jd_apply_job_select").select2({
       placeholder: "Select your Job Seekers",
       ajax: { 
-        url: "/agency_people/"+this.jd_id+"/my_js_as_jd",
+        url: "/agency_people/"+jd_id+"/my_js_as_jd",
         dataType: 'json',
         type: "GET",
         delay: 250,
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/job_applications.js
```diff
@@ -10,13 +10,13 @@ var dataHandler = {
     return this;
   },
 
-  load_form: function () {
+  load_form: function (jd_id) {
```
joaopapereira: Why did we change this function? isn't the effect the same?

patmbolger:
Reviewed PR .... looks good.

app/assets/javascripts/job_applications.js
jeanwei: previously, 'this.jd_id' is intended to retrieve the value of handler object. But just notice that it lost its bind to handler when passed to 'select2' and returns 'undefined'. So now, pass in the argument to ensure that proper return of the value.

jeanwei: jd_id value is used to define the url in ajax call. strictly speaking, it should fail with the previous code. But since the value in param is not used in the controller, it does not post any error.

joaopapereira: OK I understand that

PR #377, opened 2016-09-20
Closes AgileVentures/MetPlus_tracker#410

jeanwei:
- job seekers permit job developer to apply job for them by default
- job seekers are able to update permission on job developer applying job for them from their edit profile page
- logics is added in jobs_controller #apply
- rspec and cucumber tests are updated accordingly

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -167,28 +167,50 @@ class JobsController < ApplicationController
 			return
 		end
 
-    if pets_user == @job_seeker || ( @job_seeker.consent && pets_user == @job_seeker.job_developer ) # to be removed once authorize is set properly
-      begin
-        job_app = @job.apply @job_seeker
-    	# ActiveRecord::RecordInvalid is raised when validation at model level fails
-    	# ActiveRecord::RecordNotUnique is raised when unique index constraint on the database is violated
-      rescue ActiveRecord::RecordInvalid, ActiveRecord::RecordNotUnique  => e
-        flash[:alert] = "#{@job_seeker.full_name(last_name_first: false)} has already applied to this job."
+    # if pets_user == @job_seeker || ( @job_seeker.consent && pets_user == @job_seeker.job_developer ) # to be removed once authorize is set properly
+    #   begin
+    #     job_app = @job.apply @job_seeker
+    # 	# ActiveRecord::RecordInvalid is raised when validation at model level fails
+    # 	# ActiveRecord::RecordNotUnique is raised when unique index constraint on the database is violated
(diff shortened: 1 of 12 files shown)


app/views/job_seekers/_form.html.haml
```diff
@@ -36,6 +36,13 @@
 
   - @address_form = js
   = render partial: 'address/addressform', locals: {fields_for_name: 'address'}
+
+  .form-group
+    .col-sm-2.control-label
+      =js.check_box :consent, {}, "true", "false"
+    .col-sm-6
+      =js.label :consent, 'By checking the box, you are giving consent to your job developer to submit a job application to a job on your behalf.'
```
patmbolger: Suggested wording:

> By checking the box to the right, you are giving consent to your job developer to submit applications to jobs on your behalf.

spec/controllers/jobs_controller_spec.rb
```diff
@@ -838,6 +838,23 @@
       end
     end
 
+    describe "invalid application as job developer: without job seeker's consent"do
```
patmbolger: please add space between describe string and `do`

app/controllers/jobs_controller.rb
```diff
@@ -167,7 +167,7 @@ def apply
 			return
 		end
 
-    if pets_user == @job_seeker || pets_user == @job_seeker.job_developer  # to be removed once authorize is set properly
+    if pets_user == @job_seeker || ( @job_seeker.consent && pets_user == @job_seeker.job_developer ) # to be removed once authorize is set properly
```
patmbolger: Suggest breaking this logic into separate statements so that we know whether we cannot apply to a job (as JD) because 1) the JS is not assigned to the user, or, 2) the JS has not given consent.

Then, change line 188 to issue error message depending on which condition applies.

jeanwei:
Thanks for the review. Code has updated as suggested. Please review to see if there is anything missed out.

app/views/job_seekers/_form.html.haml
```diff
@@ -36,6 +36,13 @@
 
   - @address_form = js
   = render partial: 'address/addressform', locals: {fields_for_name: 'address'}
+
+  .form-group
+    .col-sm-2.control-label
+      =js.check_box :consent, {}, "true", "false"
+    .col-sm-6
+      =js.label :consent, 'By checking the box to the right, you are giving consent to your job developer to submit applications to jobs on your behalf.'
```
patmbolger: OK, I'm officially losing it!  This should read:
```
By checking the box to the left, .....
```
Because, of course, the box *is* actually to the left of the text.  

Very sorry for that ....

PR #378, opened 2016-09-20
Closes AgileVentures/MetPlus_tracker#432

patmbolger:
Added rspec tests for Agency.

Added auth to agency_admin_controller

### DIFF
--- a/app/controllers/agency_admin_controller.rb
+++ b/app/controllers/agency_admin_controller.rb
@@ -1,11 +1,15 @@
 class AgencyAdminController < ApplicationController
-  def home
-    # Cancancan before_filter here .....
 
+  before_action :user_logged!
+
+  def home
     @agency = Agency.includes([ :agency_people,
                               companies: [:addresses],
                                branches: [:address] ]).
                         find(Agency.this_agency(current_user).id)
+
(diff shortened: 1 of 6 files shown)


app/controllers/agency_admin_controller.rb
```diff
@@ -1,12 +1,15 @@
...
 
+  before_action :user_logged!
+
+  def home
     @agency = Agency.includes([ :agency_people,
                               companies: [:addresses],
                                branches: [:address] ]).
                         find(Agency.this_agency(current_user).id)
 
+    check_authorization(@agency)
```
joaopapereira: Do you think we should have also fail tests due to authentication in the controller? or are they redundant?

patmbolger: Yes, controller tests make sense.   I added those.

I wonder if, with controller tests, we can dispense with acceptance tests?  Most (or all) of what we can do with those is to confirm the flash message, and maybe the page we are redirected to. We can do all of that in controller tests.  What do you think?

joaopapereira: Hmm Thinking what would be best, to do these UT for the controller or Acceptance, maybe acceptance make more sense

PR #381, opened 2016-09-23
Closes AgileVentures/MetPlus_tracker#426

dsmegha:
Changed the business logic to provide all company contacts for the assignment of tasks.

Added test case as well.


app/models/concerns/task_manager/business_logic.rb
```diff
@@ -91,6 +91,8 @@ def assignable_list
       info = ASSIGNABLE_LIST[task_type.to_sym]
       if info[:type] == :agency
         return Agency.send info[:function], task_owner[0].agency
+      elsif info[:type] == :company
+        return task_owner[0].company.people_on_role("Company Contact")
```
joaopapereira: We should create a function
Company.company_contact(company)
And call it like the agency
Company.send info[:function], task_owner[0].company

Change the ASSIGNABLE_LIST[:job_application] = {:type => :company, :function => :company_contact}

dsmegha:
Hi, 

We had used the send info since we needed to dynamically determine the function in case of agency, depending on the task type. But since for a company it's always the same function, I thought this will be a much readable code. We can add that code when we decide as another task type for the company.
Let me know if I should go ahead and change it.

app/models/concerns/task_manager/business_logic.rb
patmbolger: If we go with this, the method should be pluralized, that is:
```
Company.company_contacts(company)
```

patmbolger: Recommend using the key for the ROLE hash to find the role value string, that is:
```
return task_owner[0].company.people_on_role(CompanyRole::ROLE[:CC])
```
I don't know if we'll ever want to change the value but we should retain flexibility to do so.

patmbolger: 3rd comment on this .... as the story notes:

> The Company Contact might be himself [company admin] too.

This means the company contact needs to be able to assign the task to himself.  (in fact, it could be that for small companies there will be only one user from the company on PETS - who, by definition, is the company admin).

So, the assignable_list method has to include all people in the company.

patmbolger: Perhaps a way to address this is to ensure that all company admins also have the :CC role.

patmbolger: Re this 3rd comment .... lets address that in a separate story.

joaopapereira:
After talking on the standup I think we should change this PR as suggestion by me. @dsmegha if you can please do the changes suggestion

joaopapereira:
This funcitonality was added in another PR

PR #382, opened 2016-09-23
Closes AgileVentures/MetPlus_tracker#422

jeanwei:
Case manager is able to update profile of his job seekers
 - 'edit-by-cm.html.haml' is created to include the attributes allowed to be edited. Not all are included
 - spec and feature files are updated accordingly to allow them to pass user authentication. quite a big portion of spec file is updated. 

Question
1. how is the user authentication supposed to be implemented? currently, there is no authentication. so, :current_user is added for certain action to quiet the pets_user = nil situation due to warden lazy loading way.
2. will probably need to confirm on which parameters are allowed to be edited by case manager and whitelist it accordingly in pundit.  
3. is agency_admin allowed to edit or delete job seeker? Current implementation has button to that


app/views/job_seekers/edit_by_cm.html.haml
```diff
@@ -0,0 +1,48 @@
+%h2
```
joaopapereira: Can't we use the same form as we already have the edit?
I believe we can use pundit to select the fields that we need to edit

app/views/job_seekers/show.html.haml
```diff
@@ -1,9 +1,14 @@
 .col-sm-3
   %h3
     Job Seeker
+    - if pets_user == @jobseeker.case_manager
```
joaopapereira: Is this a different button? Can't we use the same as the job seeker?

app/views/job_seekers/_assigned_agency_person.html.haml
```diff
@@ -2,7 +2,7 @@
 - when 'JD'
   - if job_seeker.job_developer.nil?
     %i None assigned &nbsp &nbsp
-    - if !preview_mode && pets_user.is_job_developer?(current_agency)
+    - if !preview_mode && person.is_job_developer?(current_agency)
```
joaopapereira: Why this change?

app/controllers/job_seekers_controller.rb
```diff
@@ -1,4 +1,5 @@
 class JobSeekersController < ApplicationController
+  before_action :current_user, except: [:new, :create]
```
patmbolger: We have defined another method for this:
```
before_action :user_logged!, except: [:new, :create]
```

app/views/job_seekers/edit_by_cm.html.haml
```diff
@@ -0,0 +1,48 @@
...
+      - if @current_resume
+        = "(Your resume on file is \"#{@current_resume.file_name}\",       |
+                 and was uploaded on #{@current_resume.updated_at.         |
+                in_time_zone('Eastern Time (US & Canada)').to_s(:long)})"  |
+
+  .form-group
+    .col-sm-2.col-sm-offset-2
+      =js_form.submit class: 'btn btn-primary'
+    .col-sm-2.col-sm-offset-1
+      =link_to 'Cancel', job_seeker_path(@jobseeker), class: 'btn btn-danger'
```
patmbolger: I think it is important for the CM to be able to update these fields as well:

1. résumé
2. status

patmbolger:
Re the 3rd question: "is agency_admin allowed to edit or delete job seeker?"  For now, yes. 

We need some way to deactivate or remove a job seeker in the system.  Right now, this is only via deleting the user.  Later we might change to setting a status, but no need to change right now.

app/views/job_seekers/edit_by_cm.html.haml
jeanwei: ok. i have removed the redundant edit-by-cm file. will let pundit does the job.

app/views/job_seekers/show.html.haml
jeanwei: job seeker does not have edit button in job_seekers/show view. that is in home view. right now i let agency_admin & cm share the same button to edit the job seeker

app/views/job_seekers/_assigned_agency_person.html.haml
jeanwei: my miss. have reverted it back.

app/views/job_seekers/edit_by_cm.html.haml
jeanwei: Ok. But I am removing the edit-by-cm file and let pundit do the job on allowing specified fields. will be cleaner then

jeanwei:
Thanks for the review. Code has been updated as suggested. Please review again to see if there is something missing

joaopapereira:
Going to merge

PR #385, opened 2016-09-26
add notification to case manager

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#423
CM Informed when js hired.
In this story I modified the following files.
pusher.js.erb.
event.rb.
event_spec.rb.


vijayakarumudi:
PR for review.

PR #388, opened 2016-09-29
Close AgileVentures/MetPlus_tracker#442

joaopapereira:


app/policies/skill_policy.rb
```diff
@@ -0,0 +1,14 @@
...
+  def show?
+    not user.nil? and User.is_agency_admin? user
+  end
+  def update?
+    not user.nil? and User.is_agency_admin? user
+  end
+  def destroy?
+    not user.nil? and User.is_agency_admin? user
+  end
+end
```
patmbolger: If the skills controller had this:
```
before_action :user_logged!
```
we wouldn't have to check `user.nil?` every time.  (or should we anyway?)

joaopapereira: Better safe then sorry, but not 100% necessary

PR #389, opened 2016-09-29
Closes AgileVentures/MetPlus_tracker#439

joaopapereira:


app/policies/job_category_policy.rb
```diff
@@ -0,0 +1,14 @@
...
+  def show?
+    not user.nil? and User.is_agency_admin? user
+  end
+  def update?
+    not user.nil? and User.is_agency_admin? user
+  end
+  def destroy?
+    not user.nil? and User.is_agency_admin? user
+  end
+end
```
patmbolger: Same question as for skills ... should we add `before_action :user_logged!` to controller and not check `user.nil?` in the policies?

PR #393, opened 2016-09-30
Pundit auth to branches controller 435

vijayakarumudi:
Closes AgileVentures/MetPlus_tracker#435.


vijayakarumudi:
PR for review.

spec/controllers/branches_controller_spec.rb
```diff
@@ -3,13 +3,32 @@
...
+    end
+    
     before(:each) do
-      get :show, id: branch
+      sign_in jd
+      get :show, id: branch.id
     end
+ 
+    before(:each) do
+      sign_in cm
```
joaopapereira: I am confused you have 3 different before each, why?

spec/policies/branch_policy_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+
+    it 'allows access if user is an agency admin' do    
+      expect(BranchPolicy).to permit(admin, branch)
+    end
+
+  end
+  
+  permissions :show?  do
+
+    it 'allows access if user is an agency person' do
```
joaopapereira: Missing tests:
* Not agency people not permitted
* Agency people but from another agency not permitted

spec/policies/branch_policy_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  let(:cm)     {FactoryGirl.create(:case_manager, agency: agency)}
+  let(:admin)  {FactoryGirl.create(:agency_admin, agency: agency)}
+  
+  
+
+  permissions :new?, :create?, :edit?, :update?, :destroy? do
+
+    
+    it 'denies access if user is job developer' do
+      expect(BranchPolicy).not_to permit(jd, branch)
```
joaopapereira: Missing tests:
* Not agency people not permitted
* Agency people but from another agency not permitted

spec/controllers/branches_controller_spec.rb
vijayakarumudi: I tested all the agency persons. make sure it is working across the all the users.

joaopapereira: OK you have 3 it how many times do they run? 3 times each or 1 time each?

spec/policies/branch_policy_spec.rb
vijayakarumudi: I created new tests for another agency and not agency people

vijayakarumudi:
PR for review

spec/controllers/branches_controller_spec.rb
patmbolger: I think you're trying to check authorization for this action.  I suggest not doing that here but add a separate section to the file that contains auth tests for _all_ actions.

For instance, here is an example test for the `#new` action:
```
describe 'action authorization' do
  context '#new' do
    it 'authorizes agency admin' do
      expect(subject).to_not receive(:user_not_authorized)
      allow(controller).to receive(:current_user).and_return(admin)
      get :new, agency_id: agency
    end
    it 'does not authorize non-admin agency person' do
      allow(controller).to receive(:current_user).and_return(jd)
      get :new, agency_id: agency
      expect(flash[:alert]).to eq "You are not authorized to create a branch."
    end
  end
end
```
See more examples of controller auth tests in `agency_people_controller_spec.rb`, starting at line 556.

spec/controllers/branches_controller_spec.rb
```diff
@@ -3,13 +3,32 @@
 RSpec.describe BranchesController, type: :controller do
```
patmbolger: If you want to clean up this spec file a little, you could move all the object create statements up here (with exception), and remove those from the individual test suites (describe blocks).  For instance:
```
let(:agency)  { FactoryGirl.create(:agency) }
let(:branch)  { FactoryGirl.create(:branch, agency: agency) }
let(:jd)      { FactoryGirl.create(:job_developer, agency: agency) }
let(:cm)      { FactoryGirl.create(:case_manager, agency: agency) }
let(:admin)   { FactoryGirl.create(:agency_admin, agency: agency) }
```
Note that I would use `branch`, rather than `branch1` in the tests below that call for `branch1`.  There is no need to create `branch1` and `branch` can be used instead.

Also, the one object create that I would retain below is `branch2` - because in one place it is created, and in the other it is built - so, we need to keep that local to each of the two tests.

app/controllers/branches_controller.rb
```diff
@@ -1,15 +1,25 @@
...
   def create
     @agency = Agency.find(params[:agency_id])
-    @branch = Branch.new
+    @branch = Branch.new(agency: @agency)
+
+    self.action_description ="create a branch"
+    authorize @branch
+
     @branch.assign_attributes(branch_params)
-    @agency.branches << @branch
```
joaopapereira: Good catch here

spec/controllers/branches_controller_spec.rb
joaopapereira: ypu still have 3 before each condition. should have only one

vijayakarumudi: I updated my file.

spec/controllers/branches_controller_spec.rb
vijayakarumudi: I cleaned up the spec file.

spec/controllers/branches_controller_spec.rb
vijayakarumudi: I added all authorizations for this controller

vijayakarumudi:
updated PR for review

spec/controllers/branches_controller_spec.rb
```diff
@@ -170,4 +178,91 @@
...
+  
+    context '#create' do
+      it 'authorizes agency admin' do
+        expect(subject).to_not receive(:user_not_authorized)
+        allow(controller).to receive(:current_user).and_return(admin)
+        post :create, agency_id: agency, branch: FactoryGirl.attributes_for(:branch)
+      end
+      it 'does not authorize non-admin agency person' do
+        allow(controller).to receive(:current_user).and_return(jd)
+        get :create, agency_id: agency
```
patmbolger: HTTP verb should be `post`, not `get`.

spec/controllers/branches_controller_spec.rb
```diff
@@ -170,4 +178,91 @@
...
+        expect(flash[:alert]).to eq "You are not authorized to create a branch."
+      end
+    end
+
+    context '#update' do
+      
+      it 'authorizes agency admin' do
+        expect(subject).to_not receive(:user_not_authorized)
+        allow(controller).to receive(:current_user).and_return(admin)
+        get :update, id: branch.id, branch: FactoryGirl.attributes_for(:branch)
```
patmbolger: `patch` instead of `get`.  Also on line 217.

spec/controllers/branches_controller_spec.rb
```diff
@@ -170,4 +178,91 @@
...
+        expect(flash[:alert]).
+          to eq "You are not authorized to edit the branch."
+      end
+    end
+
+    context '#destroy' do
+      it 'authorizes agency admin' do
+        expect(subject).to_not receive(:user_not_authorized)
+        allow(controller).to receive(:current_user).and_return(admin)
+        patch :destroy, id: branch.id
```
patmbolger: `delete` instead of `patch`.  Also on line 245.

vijayakarumudi: I updated all HTTP verbs.

vijayakarumudi:
Updated the PR.

spec/controllers/branches_controller_spec.rb
```diff
@@ -170,4 +178,91 @@
     end
   end
 
+  describe 'action authorization' do
+    context '#new' do
+      it 'authorizes agency admin' do
+        expect(subject).to_not receive(:user_not_authorized)
```
joaopapereira: I believe this line should go to the end of the test.
In this example and all the next ones

spec/controllers/branches_controller_spec.rb
```diff
@@ -170,4 +178,91 @@
...
+        allow(controller).to receive(:current_user).and_return(admin)
+        get :new, agency_id: agency
+      end
+      it 'does not authorize non-admin agency person' do
+        allow(controller).to receive(:current_user).and_return(jd)
+        get :new, agency_id: agency
+        expect(flash[:alert]).to eq "You are not authorized to create a branch."
+      end
```
joaopapereira: Still missing the tests for all the other use type and also not authenticated user. 
Here and in the next ones

spec/controllers/branches_controller_spec.rb
patmbolger: João - please check that .... I believe this expectation is checked at the _completion_ of the test, so it does not matter where it is in the test.

"Unfulfilled
message expectations trigger failures when the example completes." - https://relishapp.com/rspec/rspec-mocks/docs/basics/expecting-messages

vijayakarumudi:
Updated PR for review

vijayakarumudi:
Updated PR for review

vijayakarumudi:
Updated PR for review

joaopapereira:
I will merge it tomorrow

PR #395, opened 2016-10-03
Closes AgileVentures/MetPlus_tracker#452

patmbolger:
## Story
As a developer
I want entity status structure to be consistent

## Implementation

This story consists of refactoring Company model status attribute from a string value to an enum, similar to the refactoring that was done for Job status.

Note that the migration file has to accommodate the fact that we are using PostgreSQL in production (see migration file: 20160825220149_change_job_status_to_enum.db)

This includes adding status change times to the Company model (has_many :status_changes, as: :entity, dependent: :destroy), and adding tracking of status change times as appropriate. (see StatusChange model).

This story includes changes to:

* Migration file (add)
* Company model
* companies_controller
* company_registrations_controller
* Views which reference company status
* db/seeds.rb
* Company model unit tests
* company_registrations_controller_spec
* spec/factories/companies
* Cucumber tests

### DIFF
--- a/db/migrate/20161003201209_change_company_status_to_enum.rb
+++ b/db/migrate/20161003201209_change_company_status_to_enum.rb
@@ -1,0 +1,14 @@
+class ChangeCompanyStatusToEnum < ActiveRecord::Migration
+  def change
+    begin
+      if ActiveRecord::Base.connection.instance_of? ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
+        execute "ALTER TABLE companies ALTER status DROP DEFAULT;"
+        change_column :comp, :status, 'integer USING CAST(status AS integer)', default: 0, null: false
+      else
+        change_column :companies, :status, :integer, default: 0, null: false
+      end
+    rescue NameError
+      change_column :companies, :status, :integer, default: 0, null: false
+    end
+  end
+end
--- a/db/migrate/20161003201209_change_company_status_to_enum.rb
+++ b/db/migrate/20161003201209_change_company_status_to_enum.rb
@@ -2,8 +2,12 @@ class ChangeCompanyStatusToEnum < ActiveRecord::Migration
   def change
     begin
       if ActiveRecord::Base.connection.instance_of? ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
-        execute "ALTER TABLE companies ALTER status DROP DEFAULT;"
-        change_column :comp, :status, 'integer USING CAST(status AS integer)', default: 0, null: false
+        execute "UPDATE TABLE companies SET status='0' where status='Pending Registration'"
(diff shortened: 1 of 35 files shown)


db/migrate/20161003201209_change_company_status_to_enum.rb
```diff
@@ -0,0 +1,14 @@
+class ChangeCompanyStatusToEnum < ActiveRecord::Migration
+  def change
+    begin
+      if ActiveRecord::Base.connection.instance_of? ActiveRecord::ConnectionAdapters::PostgreSQLAdapter
+        execute "ALTER TABLE companies ALTER status DROP DEFAULT;"
+        change_column :comp, :status, 'integer USING CAST(status AS integer)', default: 0, null: false
```
joaopapereira: This is not enough for the migration, In the one I did was enough because I went to Heroku and update the table by hand.
What we need here is to convert the string into the correction number
execute "UPDATE TABLE companies SET status='0' where status='Pending Registration'"
And so on for the other status

features/company_registration.feature
```diff
@@ -71,7 +71,7 @@ Scenario: company registration approval
   And I click the "Admin" link
   And I click the "Agency and Partner Companies" link
   Then I click the "Widgets, Inc." link
-  Then I should see "Pending Registration"
+  Then I should see "pending_registration"
```
joaopapereira: Should we see Pending Registration? In terms of visual output for the user I believe it is best to keep it in text format instead of using the symbol

db/migrate/20161003201209_change_company_status_to_enum.rb
patmbolger: Do you think we need to worry about recovering existing status values in dev?  Or just staging/prod?

patmbolger:
PR adjusted as recommended.

spec/javascripts/company_registrations_spec.js
```diff
@@ -33,7 +33,7 @@ describe('Company Registration', function () {
 
       $('#send_button').trigger('click');
       expect($.ajax).toHaveBeenCalled();
-      expect($('#company_status').html()).toContain('Registration Denied');
+      expect($('#company_status').html()).toContain('registration_denied');
```
joaopapereira: This no longer is the case, Registration Denied is the correct text again

spec/javascripts/fixtures/company_registration/registration_denied.html
```diff
@@ -1 +1 @@
-Registration Denied
+registration_denied
```
joaopapereira: This no longer is the case, Registration Denied is the correct text again

patmbolger: Corrected Jasmine test and fixture references.

PR #399, opened 2016-10-05
Closes AgileVentures/MetPlus_tracker#453

patmbolger:
Change CompanyPerson status from hash to enum.

### DIFF
--- a/app/models/company_person.rb
+++ b/app/models/company_person.rb
@@ -9,6 +9,8 @@ class CompanyPerson < ActiveRecord::Base
   enum status: [:company_pending, :invited, :active, :inactive, :company_denied]
   has_many :status_changes, as: :entity, dependent: :destroy
 
+  validate :not_removing_sole_company_admin, on: :update
+
   def company_pending
     company_pending!
     StatusChange.update_status_history(self, :company_pending)
@@ -33,8 +35,6 @@ class CompanyPerson < ActiveRecord::Base
     company_denied!
     StatusChange.update_status_history(self, :company_denied)
   end
-
-  validate :not_removing_sole_company_admin, on: :update
 
   scope :all_company_people, ->(company) {
     where(company_id: company.id).joins(:user).order('users.last_name')
--- a/app/models/company_person.rb
+++ b/app/models/company_person.rb
@@ -6,13 +6,33 @@ class CompanyPerson < ActiveRecord::Base
                           join_table: 'company_people_roles',
                           autosave: false
 
-  STATUS = { PND:   'Pending', # Company has registered but not yet approved
-             IVT:   'Invited', # Company approved, invite sent to confirm account
-             ACT:   'Active',
-             INACT: 'Inactive',
-             DENY:  'registration_denied'} # Company registration denied
+  enum status: [:company_pending, :invited, :active, :inactive, :company_denied]
+  has_many :status_changes, as: :entity, dependent: :destroy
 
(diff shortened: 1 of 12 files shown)


app/models/company_person.rb
```diff
@@ -6,13 +6,33 @@ class CompanyPerson < ActiveRecord::Base
...
-  STATUS = { PND:   'Pending', # Company has registered but not yet approved
-             IVT:   'Invited', # Company approved, invite sent to confirm account
-             ACT:   'Active',
-             INACT: 'Inactive',
-             DENY:  'registration_denied'} # Company registration denied
+  enum status: [:company_pending, :invited, :active, :inactive, :company_denied]
+  has_many :status_changes, as: :entity, dependent: :destroy
 
-  validates :status, inclusion: STATUS.values
+  def company_pending
```
joaopapereira: I would move this part of added code down to maintain validations on to our the file to be more readable

app/models/user.rb
```diff
@@ -96,9 +96,9 @@ def is_company_person? company
   end
 
   def inactive_message
-    if !approved? && actable.status == CompanyPerson::STATUS[:PND]
+    if !approved? && actable.company_pending?
       :signed_up_but_not_approved
```
joaopapereira: This function only exists in company people, do we call inactive_message on any other type of user? If so maybe we need to have a default behavior for the rest of the users or check if function is callable

app/views/company_people/show.html.haml
```diff
@@ -38,9 +38,9 @@
           - if User.is_agency_admin?(pets_user) || User.is_company_admin?(pets_user)
             = link_to '(Invite Again)',
               user_invitation_path(user: {email: @company_person.email,
-                                   first_name: @company_person.first_name,
-                                   last_name: @company_person.last_name} ),
-              method: :post if @company_person.status == CompanyPerson::STATUS[:IVT]
+                                     first_name: @company_person.first_name,
+                                      last_name: @company_person.last_name} ),
+                                         method: :post if @company_person.invited?
```
joaopapereira: The indentation looks funny

db/migrate/20151224155645_add_status_to_company_people.rb
```diff
@@ -1,6 +1,5 @@
 class AddStatusToCompanyPeople < ActiveRecord::Migration
   def change
-    add_column :company_people, :status, :string,
-                      default: CompanyPerson::STATUS[:ACT]
+    add_column :company_people, :status, :string, default: 'Active'
```
joaopapereira: Just for the sake of consistency I would keep the status all lower case as you did in the previous PRs

app/models/company_person.rb
patmbolger: OK, I moved the `validate` statement to near the top of the file (assuming this was what was meant?).

app/models/user.rb
patmbolger: We're only calling `inactive_message` on a company person.  However, I will use `try` to prevent errors if we end up calling the method on other people types.

app/views/company_people/show.html.haml
patmbolger: Adjusted.

db/migrate/20151224155645_add_status_to_company_people.rb
joaopapereira: You are right

PR #401, opened 2016-10-06
Closes AgileVentures/MetPlus_tracker#455

patmbolger:
The JobApplication model contained this method for returning a string representation of the application status, suitable for display:
```
def status_name
    status.to_s.camelcase
end
```
This method was deleted and instead the common helper method status_desc, defined in application_helper.rb, is used instead.

Also, added method tests to application_helper_spec.rb


joaopapereira:
cannot add comments on the files...

patmbolger:
Removed extraneous files.

PR #403, opened 2016-10-10
Closes AgileVentures/MetPlus_tracker#445 download resume

lollypop27:
Download a resume as an employer

https://github.com/AgileVentures/MetPlus_tracker/issues/445


features/job_application.feature
```diff
@@ -125,3 +128,16 @@ Given the following agency relations exist:
...
+    Given I am on the home page
+    And I am logged in as "ca@widgets.com" with password "qwerty123"
+    And I should see "software developer"
+    When I click the "1" link
+    And I wait 1 second
+    And I should see "Applications for this Job"
+    Then I click the "Seeker, John" link
+    Then I should see button "Download Resume"
+    And I click the "Download Resume" button
+    Then I should get a download with the filename "Janitor-Resume.doc"
```
patmbolger: When we removed the job application from the test setup we broke this scenario.  These changes will fix that:

1) Add this line following `Given the following jobseeker exist:`
```
| Jane      | Seeker   | jane.seeker@places.com    | 345-890-7890| password   |password             | 1990          |Unemployed Seeking |
```
(that will create a 2nd job seeker)

2) Add this line following `Given the following resumes exist:`
```
| Janitor-Resume.doc | jane.seeker@places.com |
```
(this will give JS Jane Seeker a résumé)

3) Change line 140 (line number before above changes) from
```
Then I click the "Seeker, John" link
```
to
```
Then I click the "Seeker, Jane" link
```

patmbolger:
Two of the CI test failures are due to 
```
undefined local variable or method `offer_download' for #<#<Class:0x007f866c9c9cb8>:0x007f866e3409a8> (ActionView::Template::Error)
```
This is because the partial `job_seekers/_info.html.haml` is being referenced in 4 places, and only 3 of these are setting the local variable `offer_download` before rendering the partial.

The 4th place is action `#preview_info` in `job_seekers_controller.rb`.

To fix, you'll need to change line 109-110 from:
```
render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker,
                                                   preview_mode: true }
```
to:
```
render partial: '/job_seekers/info', locals: { job_seeker: @jobseeker,
                                                   preview_mode: true, offer_download: false }
```

app/controllers/job_applications_controller.rb
```diff
@@ -1,54 +1,81 @@
...
+  def find_application
+    begin
+      @job_application = JobApplication.find(params[:id])
+    rescue
+      flash[:alert] = "Job Application Entry not found."
+      redirect_back_or_default
+    end
+  end
+
+  def download_resume
```
joaopapereira: What was the rational behind placing this function here instead of the JobSeekerController?

features/step_definitions/devise_steps.rb
```diff
@@ -28,3 +28,7 @@
   step %{I am on the home page}
   step %{I login as "#{person.email}" with password "qwerty123"}
 end
+
+Given(/^I am on "(.*?)" page$/) do |pagename|
```
joaopapereira: This step already exist in navigation_steps.rb
When(/^I go to the (.+) page$/)
Given(/I am on the (.+) page$/)

app/controllers/job_applications_controller.rb
lollypop27: @patmbolger can you help answer @joaopapereira question please?

features/step_definitions/devise_steps.rb
lollypop27: @joaopapereira ok will look into that and get it removed

app/controllers/job_applications_controller.rb
joaopapereira: I think we might have made similar assumptions in the past. The rule of thumb that we should use might be. If it needs a job seeker ID it should go to the jobseeker  controller. If need the company ID we use the company controller, something like
/jobseeker/:id/download_resume
Because the action is over the job seeker

lollypop27:
@patmbolger I pushed up again and no merge conflicts, happy with that :100: 
Just hound now failing :(, I am away from this afternoon and back on Monday if anyone wants too look at it? As I know this pull request has been a long time waiting.

patmbolger:
@lollypop27 - I will fix the hound issues in my local feature branch.  I will also create a test PR to confirm the fixes when houndci runs.  Then, you can pull my feature branch and push that up to your repo.

lollypop27:
@patmbolger @joaopapereira allchecks passed on Semaphore :8ball:

patmbolger:
I think this is good to merge.

I will create a separate story for refectoring `download_resume` from job_applications_controller` to `job_seekers_controller`.

lollypop27:
@patmbolger Great news thank you!

PR #412, opened 2016-10-20
Closes AgileVentures/MetPlus_tracker#448

joaopapereira:


config/boot.rb
```diff
@@ -1,3 +1,15 @@
...
+
+require 'rubygems'
+require 'rails/commands/server'
+
+module Rails
+  class Server
+    alias :default_options_bk :default_options
+    def default_options
+      default_options_bk.merge!(Host: '0.0.0.0')
+    end
```
patmbolger: Suggest adding comment here explaining what this is doing and why.

PR #426, opened 2016-11-01
Add rendering of Stars

joaopapereira:
Take a look at the stars:

image

The only thing needed to add the stars somewhere is just call the function
show_stars(1)
on the template


app/helpers/stars_renderer.rb
```diff
@@ -0,0 +1,38 @@
...
+  def star_type(value)
+    if value <= 0
+      'fa-star-o'
+    elsif value == 1
+      'fa-star-half-o'
+    else
+      'fa-star'
+    end
+  end
+end
```
patmbolger: I think it's preferable to show a half star when the residual value is between 0.4 and 0.8, exclusive (right now, we show only one star for 1.9, for example, when it might be better to show 1.5 stars).

I played with the code and I think these changes might achieve it:

```
def star_images
    stars = []
    (0..4).map do |position|
      stars << star_image((@rating - position).round(1))
    end
    stars
end
```

```
def star_type(value)
    if value <= 0.3
      'fa-star-o'
    elsif value > 0.3 and value < 0.9
      'fa-star-half-o'
    else
      'fa-star'
    end
end
```

PR #428, opened 2016-11-04
Closes AgileVentures/MetPlus_tracker#482

nchristian777:
-Moves the edit button seen when viewing as Case Manager or Agency Admin from bottom
of page up to under Hide Information.
Note: when viewing as an Agency Admin there is a delete button the accompanies the edit button. I assumed that it was to be moved as well since it is paired with the edit button.

### DIFF
--- a/.hound.yml
+++ b/.hound.yml
@@ -1,1 +1,4 @@
 fail_on_violations: true
+
+ruby:
+  config_file: .rubocop.yml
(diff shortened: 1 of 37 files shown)


patmbolger:
The PR looks good.  One suggested change - use call `btn-xs` for the *edit* and *delete* buttons.  The normal sized buttons - to my eye - look too big in their current position.

That is, this is current button size:
screen shot 2016-11-04 at 8 58 29 am

And I think it looks better like this:
screen shot 2016-11-04 at 8 58 48 am

What do you think?

nchristian777:
changed button size as suggested.

joaopapereira:
Looks good, please rebase this story and we shoud be good to merge

PR #430, opened 2016-11-04
Closes AgileVentures/MetPlus_tracker#388

joaopapereira:
Ability to find jobs that match a specific Job Seeker resume


app/views/agency_people/_assigned_job_seekers.html.haml
```diff
@@ -23,11 +25,18 @@
           - when :last_sign_in_at
             %td.col-md-3
               = js.last_sign_in_at
-         
+          - when :match_rating
+            %td.col-md-3
+              - if js.resumes.length > 0
+                = link_to list_match_jobs_job_seeker_path(js), \
```
patmbolger: The `\` in this line should be removed.

app/views/job_seekers/list_match_jobs.html.haml
```diff
@@ -0,0 +1,34 @@
...
+            %td
+              #{time_ago_in_words(job.created_at)} ago
+            %td
+              = show_stars(@star_rating[job.id])
+              #{@star_rating[job.id]} / 5.0
+  .row.text-center
+    = will_paginate @list_jobs, param_name: 'jobs_page',
+                              :params => {:controller => 'job_seekers',
+                              :action => 'list_match_jobs',
+                              :id => @jobseeker.id}
```
patmbolger: If there are no matching jobs the user is shown a page that does not indicate that.  It just contains an empty table with headers.

Suggest either:
1. Popup a dialog box - other the CP home page, that tell the user there are no matches (preferred), or,
2. Put a flash message at top of page, or,
2. Put a message in the current view that no matched were found.

app/views/job_seekers/list_match_jobs.html.haml
```diff
@@ -0,0 +1,34 @@
...
+      %thead
+        %tr.text-center
+          %td
+            Title
+          %td
+            Employer
+          %td
+            Posted
+          %td
+            Match Rating
```
patmbolger: This header title would be clearer as `Match Jobs` (we're not actually showing a match rating in that column, after all - it is just a link to invoke job-match functionality).

app/controllers/job_seekers_controller.rb
```diff
@@ -123,6 +123,20 @@ def destroy
     redirect_to root_path
   end
 
+  def list_match_jobs
+    @jobseeker = JobSeeker.find(params[:id])
+    if @jobseeker.resumes.empty? then
+      flash[:error] =
+        "#{@jobseeker.full_name(last_name_first: false)} do not have any resume"
```
patmbolger: Suggest changing 
```
do not have any resume
```
to:
```
does not have a résumé on file
```

spec/controllers/job_seekers_controller_spec.rb
```diff
@@ -720,4 +720,84 @@ def assign_role_action(role, action)
...
+        end
+      end
+      context 'Multiple matches' do
+        before(:each) do
+          job2
+          job3
+          job6
+          job8
+          job9
+          stub_cruncher_match_jobs
```
patmbolger: Wouldn't it be more efficient to simply se `let!` for these objects at the top of the describe block?  (these don't have to be created anew for each test)

spec/support/service_stub_helpers.rb
```diff
@@ -133,15 +156,14 @@ def stub_email_validate_invalid
                        \n  }
                     \n}"
       stub_request(:get,
-          /^#{EmailValidateService.service_url}\/validate?.*/).
-          to_return(body: body_json)
+                   %r{/^#{EmailValidateService.service_url}\/validate?.*/})
```
patmbolger: This should be:
```
%r{^#{EmailValidateService.service_url}\/validate?.*}
```
(drop the leading and trailing `/` characters)

spec/support/service_stub_helpers.rb
```diff
@@ -133,15 +156,14 @@ def stub_email_validate_invalid
...
+        .to_return(body: body_json)
     end
 
     def stub_email_validate_error
       stub_request(:get,
-          /^#{EmailValidateService.service_url}\/validate?.*/).
-          to_raise(RuntimeError)
+                   %r{/^#{EmailValidateService.service_url}\/validate?.*/})
```
patmbolger: This should be
```
%r{^#{EmailValidateService.service_url}\/validate?.*}
```

app/views/agency_people/_assigned_job_seekers.html.haml
joaopapereira: I need it to break the line

app/views/job_seekers/list_match_jobs.html.haml
joaopapereira: Is this comment in the correct file? because in this template we show the rating

app/views/agency_people/_assigned_job_seekers.html.haml
patmbolger: HAML knows that the statement continues on the next line because of the comma.  That \ breaks my UI.

app/views/job_seekers/list_match_jobs.html.haml
patmbolger: oops ... you're right.

patmbolger:
PR looks ready to merge.

PR #432, opened 2016-11-06
Closes AgileVentures/MetPlus_tracker#456

sigu:


app/models/event.rb
```diff
@@ -188,36 +188,39 @@ def self.evt_app_accepted(evt_obj)
...
-      Pusher.trigger('pusher_control',
-                     EVT_TYPE[:APP_ACCEPTED],
-                     {id: evt_obj.id,
-                      ap_user_id: evt_obj.job_seeker.job_developer.user.id,
-                      job_title:   evt_obj.job.title,
-                      js_name: evt_obj.job_seeker.full_name(last_name_first: false)
-                     })
+    notifyList = []
+    if job_developer = evt_obj.job_seeker.job_developer
+      unless User.is_case_manager?(job_developer)
```
patmbolger: This is checking if the job_developer (JD) is a case_manager (CM) as well.

However, what we need to check is if the JD is also the CM *for this job seeker*.  (The JD can also have the CM role, but not necessarily be the CM for this specific job seeker).

sigu: Nice, had not looked at it from this angle. Let me work on that

sigu:
Is there a way to run this hound locally before pushing and coming across this errors/warnings?

spec/models/event_spec.rb
```diff
@@ -291,19 +344,20 @@
...
-                agency_name: job_developer.agency.name})
+        with('pusher_control',
+             'jobseeker_assigned_cm',
+             {js_id:   job_seeker.id,
+              js_user_id: job_seeker.user.id,
+              js_name: job_seeker.full_name(last_name_first: false),
+              cm_name: case_manager.full_name(last_name_first: false),
+              cm_user_id: case_manager.user.id,
+              agency_name: job_developer.agency.name
+      })
```
sigu: I think this is the format used in the rest of the spec file

spec/models/event_spec.rb
```diff
@@ -237,30 +265,55 @@
...
-                job_title: job.title,
-                js_name: application.job_seeker.full_name(last_name_first: false)
-               })
+        with('pusher_control',
+             'job_application_rejected',
+             {id: application.id,
+              ap_user_id: application.job_seeker.job_developer.user.id,
+              job_title: job.title,
+              js_name: application.job_seeker.full_name(last_name_first: false)
+      })
```
sigu: @joaopapereira, @patmbolger  this is the format that is present on the whole spec file, still throwing errors at me :angry:

arun1595:
@sigu Use `rubocop filename` to see the warnings locally.

sigu:
@arun1595 Its showing me about 117 errors using `rubocop`

patmbolger:
@sigu Rubocop has an auto-fix feature that is very useful.  This is invoked via the "-a" option:
```
rubocop <filename> -a
```
It does a good job of fixing lots of small issues and also - so far - does not seem to make mistakes in doing that.

I pulled your feature branch locally and run auto-fix on event.rb.  Then I fixed a few remaining items, most importantly:

1. Rubocop wants to see a "guard clause" rather than code inside of a conditonal statement.  For example, replace this:
```
unless <condition>
   <some code here ....>
end
```
with this:
```
return if <condition>
<some code here ...>
```

2. I changed all (10) occurrences of `notifyList` to `notify_list`

After that, I ended up with zero hound violations.

PR #433, opened 2016-11-08
Closes AgileVentures/MetPlus_tracker#437

arun1595:
- Added policies and rspec tests for company_registrations
- Did some refactoring

### DIFF
--- a/app/controllers/company_registrations_controller.rb
+++ b/app/controllers/company_registrations_controller.rb
@@ -2,6 +2,7 @@ class CompanyRegistrationsController < ApplicationController
   include UserParameters
   include CompanyPeopleViewer
 
+  before_action :user_logged!, except: [:new, :create]
   before_action :load_and_authorize_company, except: [:new, :create]
 
   def new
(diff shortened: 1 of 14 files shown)


app/controllers/company_registrations_controller.rb
```diff
@@ -2,22 +2,25 @@ class CompanyRegistrationsController < ApplicationController
...
+    authorize company_registration(@company)
     @company_admins = Company.company_admins(@company)
     @people_type    = 'company-all'
   end
 
   def destroy
-    company = Company.find(params[:id])
-    company.destroy
-    flash[:notice] = "Registration for '#{company.name}' deleted."
+    authorize company_registration(@company), :update?
```
joaopapereira: Why are we using update here?

app/controllers/company_registrations_controller.rb
```diff
@@ -54,13 +57,12 @@ def create
   end
 
   def edit
-    @company = Company.find(params[:id])
+    authorize company_registration(@company), :update?
```
joaopapereira: Why using update here?

app/controllers/company_registrations_controller.rb
```diff
@@ -93,27 +95,29 @@ def approve
...
 
     flash[:notice] = "Company contact has been notified of registration approval."
-    redirect_to company_path(company.id)
+    redirect_to @company
 
   end
 
   def deny
     # Deny the company's registration request.
+    authorize company_registration(@company), :update?
```
joaopapereira: Why use update here?

app/controllers/company_registrations_controller.rb
joaopapereira: I understand that, but I would prefer to have the policy related stuff inside the policy because like this we have 2 places where you set the Policy, in this file and in the Policies file

joaopapereira:
@arun1595 I cannot merge it says there are conflicts can you check it out. Thanks

PR #436, opened 2016-11-10
Closes AgileVentures/MetPlus_tracker#50

joaopapereira:


app/jobs/agency_mailer_job.rb
```diff
@@ -0,0 +1,15 @@
...
+    when Event::EVT_TYPE[:OTHER_JD_APPLY]
+      # args[0] = job_developer
+      # args[1] = job_developer
+      # args[2] = job
+      AgencyMailer.job_applied_by_other_job_developer(job_seeker, args[0],
+                                                      args[1],
+                                                      args[2]).deliver_later
+    end
+  end
+end
```
patmbolger: Should this be combined with `notify_email_job.rb`?  That file sets up jobs for the other agency emails.  

If so, we should move the methods from `notify_email_job.rb` into this file since it is named more appropriately.

app/models/event.rb
```diff
@@ -153,15 +155,24 @@ def self.evt_jd_apply(evt_obj)  # evt_obj = job_app
...
-                                  evt_obj.job_seeker.job_developer, evt_obj.job)
+                                  job_developer, evt_obj.job)
+
+    if job_developer != evt_obj.job_seeker.job_developer
+      AgencyMailerJob.set(wait: delay_seconds.seconds)
+                     .perform_later(EVT_TYPE[:OTHER_JD_APPLY],
+                                    evt_obj.job_seeker,
+                                    evt_obj.job_seeker.job_developer,
+                                    job_developer, evt_obj.job)
+    end
```
patmbolger: We should send a Pusher notification to the primary JD as well.

spec/models/event_spec.rb
```diff
@@ -137,6 +143,27 @@
       stub_cruncher_file_download  testfile_resume
     end
 
+    describe 'Job Developer not the primary JD' do
```
patmbolger: The message actually goes to the job seeker, not the JD.

spec/models/event_spec.rb
```diff
@@ -137,6 +143,27 @@
...
+      it 'trigger message to the primary JD' do
+        Event.create(:JD_APPLY, application_diff_jd)
+        expect(Pusher).to have_received(:trigger)
+          .with('pusher_control',
+                'job_applied_by_job_developer',
+                job_id:  job_wo_cp.id,
+                js_user_id: job_seeker.user.id)
+      end
```
patmbolger: This notify message says "Your job developer has applied to \<job\> for you."

This is misleading in this case.  

We could use the pusher method but perhaps the message should be changed to read:

"Job developer \<name\> has applied to \<job\> for you." (with \<name\> being an email link).

spec/models/event_spec.rb
patmbolger: ^^ that comment is meant for the line immediately below.

app/jobs/agency_mailer_job.rb
joaopapereira: I cannot merge for now because  they have 2 different signature. The notify receive emails why the agency doesnt

app/models/event.rb
joaopapereira: Changes to the pusher were not accounted for in the story but I can add that

PR #443, opened 2016-11-16
Closes AgileVentures/MetPlus_tracker#389

patmbolger:
## Story:

The job view page, for a job seeker, should have a control (button or link) that allows the JS to run a match process for this job against his/her résumé.

See initial mock-up for this in #350.

The match level (1.0 through 5.0) will be shown via graphical symbols - stars, for example - and that graphical representation will include depiction of match level that is not necessarily an integer (e.g. 3.8). This graphic representation has to be the same for the other "matching" stories.

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -204,8 +204,8 @@ class JobsController < ApplicationController
     return render(json: { message: 'No résumé on file',
                           status: 404 }) unless resume
 
-    # result = ResumeCruncher.match_resume_and_job(resume.id, @job.id)
-    result = ResumeCruncher.match_resume_and_job(1, 3) # << DEV PURPOSES
+    result = ResumeCruncher.match_resume_and_job(resume.id, @job.id)
+    # result = ResumeCruncher.match_resume_and_job(1, 3) # << DEV PURPOSES
 
     return render(json: { message: result[:message],
                           status: 404 }) if result[:status] == 'ERROR'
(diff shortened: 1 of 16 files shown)


app/controllers/jobs_controller.rb
```diff
@@ -85,6 +86,8 @@ def create
   end
 
   def show
+    @resume = nil
+    @resume = pets_user.resumes[0] if pets_user.is_a?(JobSeeker)
```
joaopapereira: What happens if user do not have resume? Does this return nil?

app/controllers/jobs_controller.rb
```diff
@@ -210,6 +213,28 @@ def revoke
...
+    raise 'Unsupported request' unless request.xhr?
+
+    job_seeker = JobSeeker.find(params[:job_seeker_id])
+    resume = job_seeker.resumes[0]
+
+    return render(json: { message: 'No résumé on file',
+                          status: 404 }) unless resume
+
+    result = ResumeCruncher.match_resume_and_job(resume.id, @job.id)
+    # result = ResumeCruncher.match_resume_and_job(1, 3) # << DEV PURPOSES
```
joaopapereira: We can remove comments

app/controllers/jobs_controller.rb
patmbolger: Yes, this statement returns nil if no résumé.

joaopapereira:
Please rebase and we can merge

patmbolger:
Merge conflicts resolved.

PR #446, opened 2016-11-22
Closes AgileVentures/MetPlus_tracker#441

jeanwei:
1. remove hidden_tag when job created by company person
    - jobs_controller : redefine #new, #create, #edit, #update
    - view/jobs/_form.html.haml : 
      - hidden_tag for company_person_id & company_id are removed
      - logics for setting companies & company_addresses are removed
    - new.html.haml & edit.html.haml: add locals
2. remove duplicated files in view/jobs/
    - combine _compact_list & _list_all --> _list_jobs
    - jobs_controller : #list will only render _list_jobs
    - jobs_viewer : add 'recent-jobs' to job_type
    => view involved : company_person/home, job_seeker/home
3. clean up jobs_controller.rb
4. rewrite jobs_controller_spec.rb & add job_policy_spec.rb
5. add authorization to jobs_controller.rb
6. clean up conditions in view/jobs
    jobs show view:
    - job seeker (when job is not applied)
image
   - job seeker (when job is applied)
image
  - job developer
image
 - company person
image


jeanwei:
Please note currently there is no authorization for the following actions:
- index 
- list_search_jobs
- list
- update_addresses
- match_resume

patmbolger:
@jeanwei - could you please add a story for the missing authorization/login checks?

jeanwei:
@patmbolger , I have corrected my comments. Log-in has been implemented for those should be necessary. But no specific authorization to them because they are returning a list of data, which the data will be filtered accordingly based on user roles.

jeanwei:
I need some investigation on the request.referer because I suspect that it is not working

spec/controllers/jobs_controller_spec.rb
```diff
@@ -14,227 +14,56 @@
...
+      job_skills_attributes: { '0' => { skill_id: skill.id,
+                                        required: 1,
+                                        min_years: 2,
+                                        max_years: 5,
+                                        _destroy: false } } }
+  end
+  before(:each) do
     stub_cruncher_authenticate
     stub_cruncher_job_create
     stub_cruncher_job_update
```
patmbolger: These stubs run in the context of each test, of course.  However, the `let!` statements above run *before* the tests, and not in the context of a test.  That means that we are actually bypassing the stubs when we create an object that involves an external service call, such as line 23.

You can see this if you disable non-stubbed service calls.  That is done by commenting out line 34 in `spec_helper.rb`:
```
WebMock.allow_net_connect!
```
You'll see many failures.

You can fix this by creating any such objects within the test context.  For instance:
1. change line 23 to use `let` rather than `let1`
2. change the test at line 51 to this:
```
it 'increases job count by 1' do
  bosh_job
  expect(assigns(:jobs).count).to eq 2
end
```

patmbolger: You could achieve the same affect by adding this at the top of the describe block:
```
let!(:stub) do
    stub_cruncher_authenticate
    stub_cruncher_job_create
    stub_cruncher_job_update
    allow(Pusher).to receive(:trigger)
end
```

spec/policies/job_policy_spec.rb
```diff
@@ -0,0 +1,128 @@
+require 'rails_helper'
+include ServiceStubHelpers::Cruncher
+
+RSpec.describe JobPolicy do
+  let(:agency)  { FactoryGirl.create(:agency) }
+  let!(:bosh) { FactoryGirl.create(:company, name: 'Bosh', agencies: [agency]) }
+  let!(:bosh_mich) { FactoryGirl.create(:address, location: bosh) }
+  let!(:bosh_job) { FactoryGirl.create(:job, company: bosh, address: bosh_mich) }
```
patmbolger: This non-lazy create is causing tests to fail since it is running in non-stubbed context.  Replace this with:
```
let(:bosh_job) { FactoryGirl.create(:job, company: bosh, address: bosh_mich) }
```
to allow this create to run in the context of each test that requires the object.

jeanwei:
Thanks for the complete reviews. Final change done!

app/policies/job_policy.rb
```diff
@@ -1,9 +1,52 @@
...
+      User.is_agency_admin?(user) || correct_company_person?
+  end
+
   def apply?
-    not user.nil? and
-      (user.is_job_seeker? or
-          (user.is_a?(AgencyPerson) and
-              (user.is_job_developer?(user.agency) or
-              user.is_case_manager?(user.agency))))
+    record.active?
```
jeanwei: It is not removed. It is in  'JobSeekerPolicy'. because job_seeker's approval is needed.

jeanwei:
Sorry about the number of files involved. It's mainly due to cleaning up of views and test files

patmbolger:
I am OK with this PR as is at this point.  Good work !

joaopapereira:
Going to merge

PR #448, opened 2016-11-23
Closes AgileVentures/MetPlus_tracker#476

Orangeburst:
non-agency person redirected if trying to access agency admin home page

### DIFF
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -28,11 +28,8 @@ class BranchesController < ApplicationController
   end
 
   def new
-    begin
-     @agency = Agency.this_agency(current_user)
-    rescue
-     @agency = nil
-    end
+    @agency = Agency.this_agency(current_user)
+   
     @branch = Branch.new(agency: @agency)
     
     self.action_description ="create a branch"
--- a/app/controllers/branches_controller.rb
+++ b/app/controllers/branches_controller.rb
@@ -28,13 +28,15 @@ class BranchesController < ApplicationController
   end
 
   def new
-    @agency = Agency.this_agency(current_user)
-   
+    if Agency.this_agency(current_user).nil?
+      flash[:alert] = 'You are not authorized to perform this action.'
+      redirect_to(root_path) && return
+    else
+      @agency = Agency.this_agency(current_user)
+    end  
     @branch = Branch.new(agency: @agency)
-    
     self.action_description ="create a branch"
     authorize @branch
-  
     @branch.build_address
   end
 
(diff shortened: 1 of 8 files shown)


app/controllers/branches_controller.rb
```diff
@@ -28,11 +28,8 @@ def create
   end
 
   def new
-    begin
-     @agency = Agency.this_agency(current_user)
-    rescue
-     @agency = nil
-    end
+    @agency = Agency.this_agency(current_user)
```
patmbolger: Logic after this line is needed so that the behavior is similar to AgencyController #home method - that is, set a flash message and redirect to root if agency==nil.

app/models/agency.rb
```diff
@@ -16,10 +16,11 @@ def self.agency_admins(agency)
   end
 
   def self.this_agency(user)
-    raise RuntimeError, 'Logged in user is not an agency person' unless
-            user.actable.is_a? AgencyPerson
-
+    unless user.actable.is_a? AgencyPerson
+      return nil
+    end
```
patmbolger: Lines 19 - 20 might be better as:
```
return nil unless user.actable.is_a? AgencyPerson
```

features/step_definitions/navigation_steps.rb
```diff
@@ -48,6 +48,11 @@ def page_translator name
   end
 end
 
+When(/^I type agency_admin home the search bar/) do
```
patmbolger: Suggest this wording:
```
I type agency_admin home in the URL address bar
```

app/controllers/branches_controller.rb
Orangeburst: Should the message be 'You are not authorized to perform this action.'
or 'Current agency cannot be determined’

app/policies/agency_policy.rb
```diff
@@ -1,5 +1,5 @@
 class AgencyPolicy < ApplicationPolicy
   def update?
-    user.is_agency_admin? record
+    user.is_agency_admin? record == nil ? false : record
```
patmbolger: Not to be picky but I find this awkward to read, since we are passing in `false` as the `agency` argument to `is_agency_admin?`.   I think it would be more straight-forward to simply return `false` if `record` is nil:
```
record.nil?  ?  false : user.is_agency_admin?(record)
```
Reading this from left to right it clearly says "if agency is nil return false; otherwise check if user is an admin of that agency".

Orangeburst: yes i agree, reads better

patmbolger:
This PR looks good .... good job, and thanks for making requested changes.

PR #456, opened 2016-12-03
487 refactor download resume

pcaston:
The action (download_resume) was moved from job_applications_controller to job_seekers_controller.
In addition the cucumber controller features and rspec examples and steps were refactored and a new file job_seeker_steps.rb was added. 
All rspec and cucumber tests ran successfully after merging.


app/controllers/job_seekers_controller.rb
```diff
@@ -140,6 +140,29 @@ def list_match_jobs
     end
   end
 
+  def download_resume
```
patmbolger: This method should receive the `id` params as the job_seeker ID, *not* the ID of the job_application.

That latter ID should also be passed in as well - need to adjust route for that.

Note that the call to method `download_resume_path`, in views/job_seekers/_info.html.haml, needs to change to accommodate the changed route and params.

config/routes.rb
```diff
@@ -156,6 +154,8 @@
   end
 
   get 'job_seekers/:id/preview_info' => 'job_seekers#preview_info'
+  get 'job_seekers/:id/download_resume' => 'job_seekers#download_resume',
+      as: :download_resume
```
patmbolger: See comments above.  The route should include the ID of the job seeker (`:id`) as well as the ID of the particular job  application (`:job app_id`, for example).

(although this is not a pure RESTful action, it still needs to adhere to Rails convention)

joaopapereira: Hmm I thin I disagree with  this the route should look like
/job_seekers/:id/download_resume/:resume_id
And the views that show the link to this route is the one responsible to get the resume I'd.
For now the resume I'd is the same but we will have support for multiple resumes in the future

patmbolger: I think we need the job application ID, since 1) we can get the resume from the job application, and 2) we need to authorize that a company person - who wants to DL the resume - is associated with the company that posted the job that the job application is for.

If we support multiple resumes in the future we would need to associate the specific resume used for the application to the JobApplication model, so we could still get the resume from the job application.

joaopapereira: Hmm given your argument I am not sure if job_seekers is the correct controller to have this action.
Nevertheless job_seekers I'd and the resume I'd we can check if the user is from the a company that the Job Seeker applied to.

patmbolger: ok we can go with that.

Paul - send the resume ID associated with the job application to the controller action.

Then, we you authorize the action, confirm that the job seeker has **at least one** *active* job application submitted to a job posted by the user's (company person's) company.  If there are no *active* job applications then do not authorize the download.

app/controllers/job_seekers_controller.rb
```diff
@@ -140,6 +141,29 @@ def list_match_jobs
...
+    job_seeker = JobSeeker.find(params[:id])
+    resume = Resume.find_by_id(params[:resume_id])
```
patmbolger: Best to just use `Resume.find()` .... `find_by_id()` might result in `method_missing` exception and the method being created on-the-fly (that is, more processing than required).

app/controllers/job_seekers_controller.rb
```diff
@@ -140,6 +141,29 @@ def list_match_jobs
...
+    if resume.nil? 
+      raise 'Resume not found in DB'
+    else
+      authorize job_seeker
```
patmbolger: Suggest moving this to immediately below line 145 (no use in doing additional processing if user is not authorized).

app/policies/job_seeker_policy.rb
```diff
@@ -83,5 +83,7 @@ def permitted_attributes
       ]
     end
   end
-
-end
\ No newline at end of file
+  def download_resume?
+    User.is_company_person?(user) && record.job_applications.where(job: user.company.jobs).exists?
```
patmbolger: We want to limit this match to jobs that are in status `active`.  This can be done with:
```
record.job_applications.where(job: user.company.jobs.where(status: 'active')).exists?
```
or, because the `status` field is an enum, a simpler way:
```
record.job_applications.where(job:  user.company.jobs.active).exists?
```

app/controllers/job_seekers_controller.rb
```diff
@@ -104,6 +104,7 @@ def index
 
   def show
     @jobseeker = JobSeeker.find(params[:id])
+    @job_application_id = params[:job_application_id]
```
patmbolger: I don't believe we need this instance var.

pcaston:
Closing to re-open with changes

app/controllers/job_seekers_controller.rb
```diff
@@ -100,6 +100,7 @@ def home
   def index
     @jobseeker = JobSeeker.all
     authorize @jobseeker
+    @offer_download = pets_user.is_a?(CompanyPerson)
```
patmbolger: This instance var is not used in the `index` view - please remove.

pcaston:
Removed unused instance variable from index action and ran rubocop with update on job_seekers_controller.rb.

joaopapereira:
Will try to merge it at work today

PR #459, opened 2016-12-08
Closes AgileVentures/MetPlus_tracker#477

sigu:


features/job_seeker.feature
```diff
@@ -56,6 +56,10 @@ Scenario: new Js Registration
   And I fill in "Password" with "password"
   And I fill in "Password Confirmation" with "password"
   And I select "1990" in select list "Year Of Birth"
+  And I fill in "Street" with " 123 Main Street"
+  And I fill in "City" with "Detroit"
+  And I fill in "Zipcode" with "48201"
+  And I select "Michigan" in select list "State"
   Then I select "Employed Not Looking" in select list "Status"
   And I choose resume file "Admin-Assistant-Resume.pdf"
   Then I click the "Create Job seeker" button
```
patmbolger: This scenario fails right now because it is looking for an error on résumé file type, but this is not appearing because the `save` of the JobSeeker instance fails with these errors:
```
Address street can't be blank Address city can't be blank Address state can't be blank
```
This is due to problem noted in JobSeekersController#create.

We need to add attributes to the address so the `save` will not fail (after the JS save we try to save the résumé file and, if that fails, rollback the JS save and display an error to the user), or, fix the controller issue and confirm that this test does not fail event without the address attributes.

app/controllers/job_seekers_controller.rb
```diff
@@ -16,7 +17,7 @@ def create
     models_saved = @jobseeker.save
 
     if models_saved
```
patmbolger: At this point we don't require that a job seeker supply an address. (perhaps we should - need client input)

However, this logic is requiring an address because we are presenting address attributes to the Address model, and that is failing if the required attributes (for an Address) are missing:
```
validates_presence_of :street
validates_presence_of :city
validates_presence_of :state
```

We want to strip the address attributes from params if they are *all* empty.  This should prevent an attempted `save` of the address and allow the saving of the other JS attributes.

features/job_seeker.feature
patmbolger: Also, when I run `cucumber features/job_seeker.feature:150` I get this error:
```
Then I click the "Update Job seeker" button                         # features/step_definitions/web_steps.rb:132
      Render and/or redirect were called multiple times in this action. Please note that you may only call render OR redirect, and at most once per action. Also note that neither redirect nor render terminate execution of the action, so if you want to exit an action after redirecting, you need to do something like "redirect_to(...) and return". (AbstractController::DoubleRenderError)
      ./app/controllers/job_seekers_controller.rb:86:in `update'
      ./features/step_definitions/web_steps.rb:134:in `/^(?:I|they) click(?: the)?( \w*)? "([^"]*)" button$/'
      features/job_seeker.feature:160:in `Then I click the "Update Job seeker" button'
```
This might be caused by changes in the `update` action - see lines 84 and 85 (your version of the file).

Check the advice in the error message (long line beginning "Please note that you may only call render ....".

patmbolger: Note: same error occurs when I run `cucumber features/job_seeker.feature:166`.

app/controllers/job_seekers_controller.rb
patmbolger: You will also need similar logic in #update.  If the JS has not entered an address before, we need to create an empty address for them in order to do so and, if empty, remove from params for the same reasons as for #create.

app/controllers/job_seekers_controller.rb
```diff
@@ -50,15 +51,16 @@ def update
     authorize @jobseeker
     jobseeker_params = handle_user_form_parameters(permitted_attributes(@jobseeker))
     dispatch_file    = jobseeker_params.delete 'resume'
+    jobseeker_params.delete 'address_attributes' if address_is_empty?
```
sigu: I removed the address from new to make the sign up form shorter for user experience.
Must the user supply their address? What if we made the users to add their addresses via the edit page. Feels like a better UX to me 😃

app/controllers/job_seekers_controller.rb
```diff
@@ -160,6 +162,11 @@ def download_resume
 
   private
 
+  def address_is_empty?
+    address = form_params[:address_attributes]
+    address[:street].empty? && address[:city].empty? && address[:state].empty?
+  end
```
sigu: I am not getting the benefits of doing this. Kindly ellaborate what difference it makes

app/controllers/job_seekers_controller.rb
patmbolger: OK, that makes sense .... lets go with that.

app/controllers/job_seekers_controller.rb
patmbolger: Since we will not be calling the ```form_params``` method in the #create action then my prior comment does not apply.

However, this statement in the #update action:
```
jobseeker_params.delete 'address_attributes' if address_is_empty?
```
is checking if all address fields (except zip code) are empty in the params - but it is working on an entirely different instance of those params because it is calling ```form_params``` internally.  Thus, `jobseeker_params` != the params instance examined by ```address_is_empty?```.  

The fact that they are identical is not the point.  If for some reason the two instances of params vary then this code could break.

It is also not obvious to the reader of the code exactly what is happening.  If I see
```
address_is_empty?(jobseeker_params)
```

PR #460, opened 2016-12-12
Spike - #427 

jeanwei:
SPIKE: The story creates inline view of job seeker's resume:
- To support .doc, .docx file, they need to be converted to .pdf before render inline.
- This spike uses 'docsplit' gem for the file conversion. https://github.com/documentcloud/docsplit
- mechanism: 
  - it checks the operating system of the user. Then look for office reader: LibreOffice  / OpenOffice in the user machine & Use it to read and convert the file to .pdf
  - there is another one which is using heroku-buildpack-libreoffice.. I havnt figured out yet. Wonder if that refers to the cloud server??
https://github.com/documentcloud/docsplit/blob/master/lib/docsplit/pdf_extractor.rb (line 55)
- Also. need to consider how to manage the created file. want to do conversion once only for every file and render accordingly when needed or convert everytime rendering?

Another way of doing without this libreoffice is to use converter API.. but they will require fee. 
Just curious, Is it possible that  we use 'curl CLI from online conversion website?

To test the following code:
- log in as company person. 
- go to /job_seekers/202 for .doc file, /job_seekers/203 for .docx file. 
- Right now the converted file is stored in .tmp. and they are not deleted after created.

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -106,6 +106,35 @@ class JobSeekersController < ApplicationController
     @jobseeker = JobSeeker.find(params[:id])
     authorize @jobseeker
     @offer_download = pets_user.is_a?(CompanyPerson)
+    respond_to do |format|
+      format.html
+      format.pdf do
+        begin
+          resume = @jobseeker.resumes[0]
+          raise 'Resume not found in DB' unless resume
+          
+          tempfile = ResumeCruncher.download_resume(resume.id)
+          raise 'Resume not found in Cruncher' unless tempfile
+
+          pdf_filename = Docsplit.extract_pdf(tempfile.path, output: './tmp')
+          # pdf_file = File.open(pdf_filename[0])
+          
+          send_file "#{pdf_filename[0]}.pdf", 
+                    # filename: "#{resume.file_name[/(.+)\./]}pdf",
+                    type: 'application/pdf',
+                    disposition: 'inline'
+          
+        rescue RuntimeError => e
+          flash[:alert] = "Error: #{e.message}"
+          redirect_back_or_default
+        ensure
+          if tempfile
+            tempfile.close
+            tempfile.unlink
+          end
+        end
+      end
+    end
   end
 
   def preview_info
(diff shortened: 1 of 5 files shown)


app/controllers/job_seekers_controller.rb
```diff
@@ -106,6 +106,35 @@ def show
...
+      format.pdf do
+        begin
+          resume = @jobseeker.resumes[0]
+          raise 'Resume not found in DB' unless resume
+          
+          tempfile = ResumeCruncher.download_resume(resume.id)
+          raise 'Resume not found in Cruncher' unless tempfile
+
+          pdf_filename = Docsplit.extract_pdf(tempfile.path, output: './tmp')
+          # pdf_file = File.open(pdf_filename[0])
```
joaopapereira: Please remove comment

app/controllers/job_seekers_controller.rb
```diff
@@ -106,6 +106,35 @@ def show
...
+          
+          send_file "#{pdf_filename[0]}.pdf", 
+                    # filename: "#{resume.file_name[/(.+)\./]}pdf",
```
joaopapereira: Please remove comment

app/views/job_seekers/_info.html.haml
```diff
@@ -36,9 +36,13 @@
           &nbsp; &nbsp (updated #{time_ago_in_words(resume.created_at)} ago)
       %td
         - if offer_download
-          = button_to "Download Resume",
-            download_resume_path(job_seeker.job_applications.last), target: '_self',
+          = button_to "View Resume",
+            job_seeker_path(job_seeker, format: 'pdf'), target: '_self',
             method: :get, class: 'btn btn-success btn-s pull-right'
+        / - if offer_download
```
joaopapereira: Please remove comment

jeanwei:
@joaopapereira shall I just remove this spike? because after discussion with patrick, we need to look for 3rd party converter service as there is no way to ask the user to install those open office to run the inline viewer.

patmbolger:
@jeanwei - you can close this PR (but please don't remove the feature branch from your repo).

I have marked this feature for "Release 2.0", and adjusted the description in then story accordingly.

PR #469, opened 2016-12-24
Closes AgileVentures/MetPlus_tracker#507

hin101:
Change the flash message when the email address is updated.


hin101:
Will redo and open new pull request

PR #470, opened 2016-12-27
Closes AgileVentures/MetPlus_tracker#502

jeanwei:
Issue:
- 'bullet' complains about < N+1 Query method call stack /app/models/skill.rb:13:in 'has_job?' >
- '#has_job?' calls 'jobs.any?' which relates to 'jobs.count'

Completed:
- jobs_count COLUMN is added to skills TABLE
- counter_cache OPTIONS is added to JobSkill CLASS
- spec test added

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -31,9 +31,9 @@
 # Ignore application configuration
 /config/application.yml
 
-
 # Ignore Editor files
 .idea/*
 
+/spec/examples.txt
 
 .DS_Store
--- a/app/models/job_skill.rb
+++ b/app/models/job_skill.rb
@@ -1,6 +1,6 @@
 class JobSkill < ActiveRecord::Base
   belongs_to :job, inverse_of: :job_skills
-  belongs_to :skill
+  belongs_to :skill, counter_cache: :jobs_count
 
   validates_presence_of :job, :skill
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -31,9 +31,9 @@
 # Ignore application configuration
 /config/application.yml
 
-
 # Ignore Editor files
 .idea/*
 
+/spec/examples.txt
```
joaopapereira: Was there any particular reason for this file to be added to the ignore?

jeanwei: I do not know when is it being added. never seen them before. shall un-ignore?

joaopapereira: Does that exist? It is strange that the file gets created on the fly

patmbolger: I added this ignore directive in the PR I just merged (waffle #515).

For some reason, this just started to appear recently in my environment as well.

In ``spec_helper.rb` there is this comment:
```
# Allows RSpec to persist some state between runs in order to support
  # the `--only-failures` and `--next-failure` CLI options. We recommend
  # you configure your source control system to ignore this file.
  config.example_status_persistence_file_path = "spec/examples.txt"
```
and later in the file the option is turned on. (and it seems to have been turned on since the beginning - so not sure why all of a sudden rspec is creating this file).

PR #473, opened 2017-01-06
Closes AgileVentures/MetPlus_tracker#505

arun1595:


app/controllers/agency_people_controller.rb
```diff
@@ -153,21 +151,27 @@ def assign_job_seeker
...
 
     person_params = handle_user_form_parameters agency_person_params
     if @agency_person.update_attributes(person_params)
       sign_in :user, @agency_person.user, bypass: true
-      flash[:notice] = "Your profile was updated successfully."
-      redirect_to root_path
+
+      # Check if user attempted to update the email address
+      if current_email != person_params[:email]
+        flash[:notice] = 'Please check your inbox to confirm your email address'
```
patmbolger: please change to ```flash[:warning]``` so user does not assume that the email change is successfully completed (see similar comment on Hinesh's PR #469)

app/controllers/agency_people_controller.rb
```diff
@@ -153,21 +151,27 @@ def assign_job_seeker
...
-    self.action_description= "edit agency person's profile"
+    self.action_description = "edit agency person's profile"
     authorize @agency_person
   end
 
   def update_profile
     @agency_person = AgencyPerson.find(params[:id])
-
-    self.action_description= "edit agency person's profile"
+    current_email = @agency_person.email
```
patmbolger: Another way to do this is to exploit Rails auto-create of predicate methods for string fields, ala http://blog.plataformatec.com.br/2012/05/active-record-attribute-method/.

See Hinesh's PR for example.

arun1595: Done.

app/controllers/agency_people_controller.rb
arun1595: Done.

PR #480, opened 2017-01-12
Closes AgileVentures/MetPlus_tracker#249

hin101:


lollypop27:
@hin101 thank you for this, you will need to sortout the hound issues please.
Another change on the dropdown menu, once logged in and one clicks on a link from the dropdown menu (Admin), it needs to be highlighted in the yellow active state colour so one know they have chosen that link.

app/views/layouts/_footer.html.haml
```diff
@@ -1,19 +1,28 @@
...
-		-if pets_user.is_a?(CompanyPerson)
+    %div{ :class => "container" }
+        %div{:class => "footer-left"}
+            %span= "&reg;".html_safe+" MetPlus #{Date.today.year}"
+            %span= " | Terms and Conditions Apply"
+            %span= " | All Rights Reserved"
+        %div{:class => "footer-right"}
+            For employers:
+            - if pets_user.is_a?(CompanyPerson)
+                = link_to "Post jobs", new_job_path(company_id: current_user.try(:actable).try(:company_id), company_person_id: current_user.try(:actable_id)), class: "#{yield(:new_job_active)}"
```
joaopapereira: Do we need all these tries? If the user is a company person I assume that actable exists and also the company

app/views/layouts/_footer.html.haml
```diff
@@ -1,19 +1,28 @@
...
+                |
+                = link_to "Search Resumes", '#', class: "#{yield(:search_resume_active)}"
+                |
+            - elsif pets_user.try(:agency)  && pets_user.is_job_developer?(pets_user.try(:agency))
```
joaopapereira: Should this line be:
- elsif pets_user.is_job_developer? current_agency?

To make sure the user is job developer of the current agency and not of a agency?

app/views/layouts/_footer.html.haml
hin101: done

patmbolger:
1) The landing page looks good - main menu and footer.  There are two issues:

a) I think we should drop the "Home" link in the footer:
screen shot 2017-01-23 at 5 43 49 pm
This is because it really just goes to the landing page, not the user's home page (described below).  So, keeping it there will lead to confusion when the user logs in and sees a "Home" option in the main menu, which takes the user to their actual home page.

b) when I have the browser showing a wide view and I click on the "Jobs" link, the menu contracts as if I have reduce the width of the screen.   That is, it goes from this:

screen shot 2017-01-23 at 5 49 51 pm

To this:
screen shot 2017-01-23 at 5 50 16 pm

patmbolger:
2) When a registered user logs in, they end up in the "Home" page.  That page is different depending upon the type of person – jobseeker, company person or agency person.  For instance, a company person home looks like this:

screen shot 2017-01-23 at 5 53 36 pm

Note that the main menu is now missing a "Home" option - so once the user leaves their home page there is no way to get back.  So, we need to add the "Home" option back to the main menu - once a user is logged in - and the destination will be one of the the three home types (there is existing logic for that).  (and, as noted, we should drop the "Home" option in the footer - or make it go to the same page as the Home option in the main menu.

patmbolger:
3) The footer has a lot of things in it - which is no problem except that it shifts to two lines very soon as the browser width is reduced:

screen shot 2017-01-23 at 5 59 27 pm
I think we can remove:
- Terms and Conditions Apply (we don't have any yet, for that matter)
- For employers: Post jobs | Search Resumes | Partner with Us | (this is significant functionality that should not be buried in the footer, and the landing page will contain information about how an employer can "partner with us")

patmbolger:
Almost there - please prevent the "Home" option from appearing in the main menu *until* the user logs in.  The "Home" page only has relevance for a logged in user.  

screen shot 2017-01-29 at 6 29 25 am

This was not handled in the UI mockups, but we can fix that now.

PR #494, opened 2017-01-28
Closes AgileVentures/MetPlus_tracker#527

sigu:
This is work in progress to catch any feature test that will fail due to my changes 😛


sigu:
Finished up this issue.
Please review @patmbolger

spec/policies/job_seeker_policy_spec.rb
```diff
@@ -73,7 +73,7 @@
 
   permissions :show? do
     it 'only allows access if user is account owner' do
```
patmbolger: please change this description to:
```
it 'does not allow access if user is the job seeker' do
```

PR #496, opened 2017-02-01
WIP -Test ruby upgrade

StlMaris123:
Test ruby upgrade from 2.2.2 to 2.3.3


sigu:
@patmbolger this did not trigger semaphoreci, 
1. how can we get it triggered to confirm all tests are okay. 
2. is there a way of telling the ci in semaphore to use a particular ruby version?

thanks for this @StlMaris123 😃

patmbolger:
@StlMaris123 - It may be that semaphore was not triggered because there were no code changes in the PR.   There was a change in the Gemfile for the ruby version, but that is commented out.

You'll need to uncomment that, run bundle install, and resubmit.

patmbolger:
@StlMaris123 - I set the ruby version for the CI tests of this PR to 2.3.3 and the tests all pass.  

Can we try to go to 2.4 and see if that works?

(I have to set the ruby version each time we want to perform the CI tests, and then reset back to 2.2.2 until we formally upgrade - I will watch for an updated PR from you and set the version to 2.4 for CI tests).

patmbolger:
Looks good - merging

PR #498, opened 2017-02-01
 Closes AgileVentures/MetPlus_tracker#520

hin101:
My Profile Page

### DIFF
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -169,6 +169,7 @@ class JobSeekersController < ApplicationController
   def my_profile
     @jobseeker = JobSeeker.find(params[:id])
     authorize @jobseeker
+    @current_resume = @jobseeker.resumes[0]
   end
 
   private
--- a/app/controllers/job_seekers_controller.rb
+++ b/app/controllers/job_seekers_controller.rb
@@ -96,7 +96,6 @@ class JobSeekersController < ApplicationController
   end
 
   def home
-    logger.info("Person: #{pets_user.inspect}")
     @jobseeker = JobSeeker.find(params[:id])
     @recent_jobs_type = 'recent-jobs'
     authorize @jobseeker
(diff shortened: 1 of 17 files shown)


patmbolger:
Looking at this I think we want to change the design a little - to better reflect the data items that the user can edit in their profile.  

1) For a job seeker, these are the items that can be edited:

screen shot 2017-02-02 at 6 21 21 am

So, these items should be listed on the "My Profile" page.

2) On the profile page, I recommend just listing the full name (as shown above) as opposed to separate `first name` and `surname`.   (the profile edit page, of course, separates those out).

3) The item name (e.g. "Email") should be in bold, and the item value should _not_ be in bold.

4) Also, I think it looks better to left justify each column as shown above.

4) Similar to the above items for company person profile.  Show these items (plus address):

screen shot 2017-02-02 at 6 24 02 am

5) Similar comments as above for an agency person.

lollypop27:
@hin101 good job thank you. Over and above @patmbolger changes please can you make sure that the form is centred left and right on the page, at the moment the space on the left of the page is less than the right of the page. Thank you.

app/controllers/job_seekers_controller.rb
```diff
@@ -96,6 +96,7 @@ def update
   end
 
   def home
+    logger.info("Person: #{pets_user.inspect}")
```
patmbolger: This looks like a holdover from development(?)

app/views/job_seekers/my_profile.html.haml
```diff
@@ -0,0 +1,31 @@
...
+        %tr
+          %td
+            %strong Resume
+          %td
+            - if @current_resume
+              Your current resume is #{@current_resume.file_name} and was uploaded on
+              %span.utc_to_local_time= @current_resume.updated_at
+            - else
+              You have no current resume
+    = link_to "Edit", edit_job_seeker_path(@jobseeker), class: "btn btn-lg edit-profile-btn"
```
patmbolger: We're missing ```address``` in this view.

PR #504, opened 2017-02-12
Closes AgileVentures/MetPlus_tracker#536

sigu:
Feature tests for:
1. user login
2. User logout
Controller tests proved to be a pain in the


features/step_definitions/users_steps.rb
```diff
@@ -21,4 +21,22 @@
...
+Then(/^I should( not)? be remembered$/) do |remember|
+  cookies = page.driver.cookies
+  user_id = cookies['user_id']
+  person_type = cookies['person_type']
+  if remember.nil?
+    expect(user_id.expires).to be_future
+    expect(person_type.expires).to be_future
+  else
+    expect(user_id.expires).to be nil
+    expect(person_type).to be nil
```
patmbolger: The parameter to the block, `remember`, results in code that is a bit confusing.  That is, if `remember` is nil, that means we actually _do_ want to remember the user.  This results in this code:
```
if remember.nil?  # << this looks like "if we don't to remember, then make sure we are remembering"
  expect(user_id.expires).to be_future
  expect(person_type.expires).to be_future
```

I would suggest changing the parameter name to make it clear what it represents, and adjust logic accordingly, e.g.:
```
Then(/^I should( not)? be remembered$/) do |not_remember|
  cookies = page.driver.cookies
  user_id = cookies['user_id']
  person_type = cookies['person_type']
  if not_remember
    expect(user_id.expires).to be nil
    expect(person_type.expires).to be nil
  else
    expect(user_id.expires).to be_future
    expect(person_type.expires).to be_future
  end
end
```

Also, note that I changed ```expect(person_type).to be nil``` to ```expect(person_type.expires).to be nil``` (see other comment for explanation).

features/step_definitions/users_steps.rb
```diff
@@ -21,4 +21,22 @@
...
+  end
+end
+j
+Then(/^I should be logged out$/) do
+  cookies = page.driver.cookies
+  expect(cookies['person_type']).to be nil
+  expect(cookies['person_id']).to be nil
```
patmbolger: If you haven't done so already please enter a story (bug) to fix person ID (to user ID) in the controller action.

features/step_definitions/users_steps.rb
patmbolger: The last note above is related to the user object that is being used for these tests.  The user here (email: ```salemamba@gmail.com```) is set up _only_ as a User instance.  The actual users in the system, however, are always of a particular "people" type (AgencyPerson, CompanyPerson, JobSeeker), which also includes User attributes by inheritance.

To fix this I added a job seeker to the background, and used that person in your added tests.

I will send you my version of the feature file on Slack.

features/step_definitions/users_steps.rb
sigu: this is aready in https://github.com/AgileVentures/MetPlus_tracker/issues/561

features/manage_users.feature
```diff
@@ -129,3 +134,35 @@ Scenario: Resend confirmation email - sad path
...
+  Scenario: User logs out
+    Given I am on the home page
+    Then I click the "Log In" link
+    And I fill in "user_email" with "mike.smith@gmail.com"
+    And I fill in "user_password" with "password"
+    And I check "user_remember_me"
+    And I click "Log in" button
+    Then I click the "dd_menu" link
+    And I log out
+    Then I wait for 1 second
```
patmbolger: The 1 second wait is probably required because the next step does not require Capybara to find or see an element on the screen (or in the DOM).  If that _had_ been the case, then most Capybara methods will wait (up to a confirmed default timeout_ for the item to appear.  This provides plenty of time for most actions to complete and the page to reload.  Without that implicit wait, we sometimes have to put in a wait step as a result.

features/manage_users.feature
```diff
@@ -129,3 +134,35 @@ Scenario: Resend confirmation email - sad path
...
+    And I fill in "user_password" with "secret1234"
+    And I click "Log in" button
+    Then I should not be remembered
+
+  @javascript
+  Scenario: Remember user on log in
+    Given I am on the home page
+    Then I click the "Log In" link
+    And I fill in "user_email" with "salemamba@gmail.com"
+    And I fill in "user_password" with "secret1234"
```
patmbolger: ^^ Need to login with Job Seeker user (also in scenario above this).  This user does not have an "actable_type", thus this line (21) in the controller:
```
cookies[:person_type] = current_user.actable_type
```
assigns ```cookies[:person_type]``` to ```nil```, and this breaks the test.

sigu: This makes sense but it would not make the scenario to fail since we are not  checking on the cookie value

features/manage_users.feature
```diff
@@ -129,3 +134,35 @@ Scenario: Resend confirmation email - sad path
   And I click the "Resend confirmation instructions" button
   Then I should see "1 error prevented resending a confirmation email:"
   And I should see "Email was already confirmed, please try signing in"
+
+  @javascript
+  Scenario: Do not remember user on login
+    Given I am on the home page
+    Then I click the "Log In" link
+    And I fill in "user_email" with "salemamba@gmail.com"
+    And I fill in "user_password" with "secret1234"
```
patmbolger: Need to use the job seeker to login as this user is not a valid user of PETS (user is not one of JobSeeker, CompanyPerson or AgencyPerson).  These models incorporate the User model by inheritance.

See ERD for structure: https://github.com/AgileVentures/MetPlus_tracker/blob/master/metsplus_postgres_altdesign.png

See this gem for how we do this multi-table inheritance: https://github.com/hzamani/active_record-acts_as

PR #509, opened 2017-02-17
 Closes AgileVentures/MetPlus_tracker#557

Danielotieno:
Add test for the has_job? method


patmbolger:
Looks good ... merging.

PR #511, opened 2017-02-18
Closes AgileVentures/MetPlus_tracker#569

hin101:


app/views/jobs/_searched_job_list.html.haml
```diff
@@ -33,8 +33,14 @@
...
-          - unless job.active?
-            %span.label.label-info
+          - if job.status.capitalize == "Active"
+            %span.label.label-active
+              #{job.status.capitalize}
+          - elsif job.status.capitalize == "Filled"
+            %span.label.label-filled
+              #{job.status.capitalize}
+          - elsif job.status.capitalize == "Revoked"
+            %span.label.label-revoked
```
patmbolger: In order to avoid the work of capitalization two times, perhaps create an local var for the the ```if``` statement, e.g.
```
job_status = job.status.capitalize
if job_status == 'Active'
  etc ....
```

Or, you could replace the whole ```if``` block with something like:
```
%span.label{ class: "label-#{job.status}" }
  #{job.status.capitalize}
```

patmbolger:
Looks good .... merging

PR #515, opened 2017-02-19
Closes AgileVentures/MetPlus_tracker#537

StlMaris123:
add tests for CompanyPeople helper


sigu:
screenshot from 2017-02-19 19-28-27
With this PR, this is the coverage

features/company_person.feature
```diff
@@ -66,6 +66,8 @@ Feature: Company Person
     Then I click the "Smith, Jane" link
     And I should see button "Edit Person"
     And I should see button "Delete Person"
+    And I click "Edit Person" button
+    Then I should see "Company Contact"
```
patmbolger: This is not a valid test as the test will always "see" "Company Contact".

patmbolger: The intent of the `disable_company_admin?(company_person, role)` method is to check if the `company_person` is the sole admin for the company.  If so, we disable the checkbox for that role, so that a company admin cannot uncheck himself and thus leave the company with no admin (in which case no one in the company can administer the company information - including designating someone as the admin).

This is what the form looks like if the method returns true:
<hr>
screen shot 2017-02-20 at 11 24 33 am
<hr>

We need to add a test to `company_people_helper_spec.rb` and test the method's return values under all possible scenarios:
1. Company admin editing himself when he is sole admin (method returns true - checkbox should be disabled)
2. Company admin editing another company admin (should return false)
3. Company admin editing a company contact who is _not_ also an admin (returns false)

The return values noted above are those for when the `role` argument is `CompanyRole::ROLE[:CA]`.

The method _always_ returns false when the `role` argument is `CompanyRole::ROLE[:CC]` (which should also be tested).

StlMaris123: The role that is being passed to the `disable_company_admin?` method, is it for the company person? And if so what if the company_person has two roles?

patmbolger: good questions .... Yes, the role is for the company person.

If the person has two roles, the method will be called twice - once for each role.

patmbolger: @StlMaris123 - please remove those two lines from this test as they are not doing anything for us.

Gemfile
```diff
@@ -39,6 +39,7 @@ gem 'devise',           '~> 3.5.2'
 gem 'devise_invitable', '~> 1.5.2'
 
 gem 'figaro'
+gem 'pry'
```
patmbolger: What does `pry` give us that `byebug` doesn't?  I'd like to have just one such tool loaded.

patmbolger:
@StlMaris123 - almost there.  Please add this to `.gitignore` file:
```
coverage/
```
That will cause git to ignore the simplecov output files (which should not be in git).

Then, remove the coverage files from your repo and resubmit the PR.

(to remove a file from tracking by git, you can use this command:
```
git rm --cached coverage/index.html
```
Thank you.

patmbolger:
@StlMaris123 - there are still many files in the `coverage` directory that are included in this PR.   Please remove those (I see that you have removed 2 such files so far).

StlMaris123:
Hi @patmbolger i have removed all the cached files

patmbolger:
Looks good - merging.

PR #520, opened 2017-02-24
WIP-Closes AgileVentures/MetPlus_tracker#546

anyangocynthia:


anyangocynthia:
I realized the rest of the methods are private. Is there a way of testing those private methods?

I noticed that views are not rendered in controller tests hence could not test the contents, which was the easier option.

patmbolger:
@anyangocynthia - good work.

Regarding testing private methods - I think the consensus is that these don't need to be tested explicitly, since they represent an internal structure used for producing the output that is visible via non-private methods, and hence are tested implicitly.   There are some good comments here: http://stackoverflow.com/questions/16197332/should-i-test-private-methods-using-rspec

Also, you can render the views in controller tests - via `render_views` (https://www.relishapp.com/rspec/rspec-rails/v/3-5/docs/controller-specs/render-views)

With that, you can then add tests that assert the presence of certain contents in the view.

Why don't you enhance your tests (which are quite good) using that and resubmit?

PR #521, opened 2017-02-25
Redesign admin job properties page

arun1595:
Closes AgileVentures/MetPlus_tracker/issues/535

### Old
job_properties_page

### New

image

### DIFF
--- a/app/assets/stylesheets/agency_admin.scss
+++ b/app/assets/stylesheets/agency_admin.scss
@@ -7,6 +7,7 @@
   width: 20%;
   text-align: left;
 }
+
 .div_align_right {
   float: left;
   width: 50%;
@@ -15,4 +16,14 @@
 
 #bold_text {
   font-weight: bold;
+}
+
+#job_categories_table,
+#skills_table {
+  margin-top: 2em;
+  padding: 0;
+}
+
+.job-properties-container {
+  margin-top: 2em;
 }
--- a/app/views/agency_admin/job_properties.html.haml
+++ b/app/views/agency_admin/job_properties.html.haml
@@ -1,52 +1,51 @@
 - provide(:agency_admin_job_active, 'active')
 
-%h3 PETS Administration - Job Properties
+.row
+  .col-sm-10.col-sm-offset-1
+    %h3 PETS Administration - Job Properties
 
 .row
-  .col-sm-3
-    %h4 Job Specialties
+  .col-sm-10.col-sm-offset-1.job-properties-container
+    %ul.nav.nav-pills.custom-pills{ role: 'tablist' }
+      %li.active{ role: 'presentation' }
+        %a{ href: '#job-specialties-pane',  'aria-controls' => 'job-specialties', role: 'tab', data: { toggle: 'tab' } }
+          Job Specialties
+        %li{ role: 'presentation' }
+          %a{ href: '#job-skills-pane',  'aria-controls' => 'job-skills', role: 'tab', data: { toggle: 'tab' } }
+            Job Skills
 
-  - if @job_categories.empty?
-    .col-sm-3
-      %h5 There are no job specialties.
-  - else
-    .col-sm-9
-      = render partial: 'job_specialties', object: @job_categories,
-                locals: {data_type:  'job_categories',
-                         partial_id: 'job_categories_table',
-                         show_property_path:   :job_category_path,
-                         delete_property_path: :job_category_path}
-    .clearfix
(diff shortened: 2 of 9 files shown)


arun1595:
I haven't modified the following fixture files:

- agency_admin/job_categories.html
- agency_admin/skills.html

How should I go about it?

/cc @patmbolger @joaopapereira

app/views/agency_admin/job_properties.html.haml
```diff
@@ -1,52 +1,51 @@
...
+                               show_property_path:   :skill_path,
+                               delete_property_path: :skill_path}
+          .clearfix
 
-  .col-sm-3.col-sm-offset-3.table_action_button
-    - text_modal_button 'Add Job Skill', 'add_skill',
-                        'btn btn-primary btn-sm pull-right'
+        .table_action_button.floating-btn-container
+          %button.floating-btn#add-job-skill-modal-btn{ data: { toggle: 'modal', target: '#add_skill' } }
+            %span.glyphicon.glyphicon-plus
```
patmbolger: See comment above re "+" sign.

app/views/agency_admin/job_properties.html.haml
```diff
@@ -1,52 +1,51 @@
...
+                               show_property_path:   :job_category_path,
+                               delete_property_path: :job_category_path }
+          .clearfix
 
-.row
-  .col-sm-3
-    %h4 Job Skills
+        .table_action_button.floating-btn-container
+          %button.floating-btn#add-job-specialty-modal-btn{ data: { toggle: 'modal', target: '#add_job_category' } }
+            %span.glyphicon.glyphicon-plus
```
patmbolger: I'd like to retain the "Add Skill" button as I am not sure it would be apparent to a typical user what the "+" icon represents.

app/assets/stylesheets/agency_admin.scss
```diff
@@ -15,4 +18,71 @@
...
   font-weight: bold;
-}
\ No newline at end of file
+}
+
+.job-properties-pills {
+  a {
+    background: $light-grey;
+    border-bottom: 3px solid darken($light-grey, 20%);
+  }
```
patmbolger: Please make this and similar styles generic if they are to be used in other screens.  This one in particular will be used in other redesigned pages for inactive buttons.

app/views/agency_admin/job_properties.html.haml
arun1595: Yeah, I agree. The reason I floated it is because to gain attention instantly. In the existing version, the add button was all the way on the bottom and wasn't visible. I tried adding it to the top but it didn't help.

Can we add a bootstrap [tooltip](http://getbootstrap.com/javascript/#tooltips) to display 'Add new skill'?

patmbolger: I reviewed that with Lara and Paul this morning after scrum and we all agreed that the + is too obscure.  Lara would prefer a button place at the top (agree on bottom placement not optimal).

Can you give that another try?

arun1595: Okay. Let me give it a try.

patmbolger: BTW, we all like the "delete" notation.  Also, Paul is working on the companion story (for companies, people, branches) and will be reusing the generic styling you have developed (see comment about identifiers for such styling).

patmbolger:
I would not worry about the fixture files as those are stripped-down files with the structure of the HTML needed for testing JS associated with those pages.  The changes you are making for this PR should not affect those files, so we should be OK. (the tests are concerned with interactions with the modals and those have not changed with this PR).

arun1595:
@lollypop27 @patrick I have updated the design with the suggested changes. PTAL

features/agency_admin.feature
```diff
@@ -284,7 +274,7 @@ Scenario: manage job properties
   And I wait 2 seconds
   And I fill in "Description:" with ""
   And I click the "Update Specialty" button
-  And I wait 2 seconds
+  And I wait 10 seconds
```
patmbolger: This wait seems excessive (although a wait is required because the preceding step is an XHR action).

Can it be reduced?

features/agency_admin.feature
```diff
@@ -245,7 +235,7 @@ Scenario: assign job seeker to agency person
 Scenario: manage job properties
```
patmbolger: This test is failing in CI as well as locally for me.

patmbolger:
looks good - merging

PR #522, opened 2017-02-27
Fix bug on the status field of companies table

arun1595:
Fixes AgileVentures/MetPlus_tracker/issues/579


patmbolger:
Merging this PR since it is failing solely due to 1) bug in cruncher involving crate of jobs, and 2) unstubbed respectively tests that fail due to the cruncher bug.

Need this PR to be merged to fix a bug in Heroku which is caused by a failing migration (which is addressed by this PR).

PR #523, opened 2017-02-28
Closes AgileVentures/MetPlus_tracker#584

patmbolger:
1. Remove webmock and email stub from seeds.rb

2. Replace instance method in class EmailValidator with stub that does nothing

3. Also, stub some unstubbed rspec tests


patmbolger:
Merging in spite of cucumber test failures.  Failures are due to cruncher bug when trying to save a job.  Need this PR to be merged so that db:seed can be run in Heroku env.

PR #525, opened 2017-03-01
Fix sorting jobs on associated models

arun1595:
Fixes AgileVentures/MetPlus_tracker/issues/581


patmbolger:
Looks good - merging

PR #533, opened 2017-03-14
WIP job seeker  navigation buttons

cesswairimu:


sigu:
@cesswairimu please remove the hound errors, I would suggest you try using `rubocop` to check your code. Install rubocop (`gem install rubocop`) in your machine then run `rubocop 'filename'` where filename is the file you have just modified.
Rubocop checks your code the same way hound does so it will save you a little 'trouble' with hound.

sigu:
Currently the heading of the section is being hidden behind the top bar which looks a bit odd, is it possible you align it just below the header? 
screenshot from 2017-03-15 10-36-32
Align it to look like this
screenshot from 2017-03-15 10-41-22

For the accuracy of the intention of the PR, I do not know if this is what @lollypop27 had in mind to be implemented.

cesswairimu:
@sigu i am still struggling with the top bar thing. Any suggestions?

app/views/job_seekers/home.html.haml
```diff
@@ -11,11 +14,11 @@
   .col-sm-9
     = render partial: 'info', locals: {job_seeker: @jobseeker, offer_download: false}
```
hin101: If you wrap the below in a row this should bring it below the nav bar.

patmbolger:
Please review the changes that occurred to the `Admin > Job Properties` page.  That was recently styled with the new UI changes, and the css/logic for the buttons at the top of the page should be similar to what is required for this story.

sigu:
@cesswairimu , for the haml files please use https://github.com/brigade/haml-lint for syntax checking. Rubocop is for ruby files

PR #535, opened 2017-03-15
WIP Adds test for company people viewer

lollypop27:


spec/controllers/concerns/company_people_viewer_spec.rb
```diff
@@ -0,0 +1,25 @@
+require 'rails_helper'
+
+class TestConcernCompanyPeopleClass
```
patmbolger: Since the concern methods operate in the context of a controller action, this class should inherit from `ApplicationController` (the standard rails class that all controllers inherit from).

spec/controllers/concerns/company_people_viewer_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+  extend CompanyPeopleViewer
```
patmbolger: The concern methods are _instance_ methods of the controller class (they are _not_ defined as `def self.display_company_people ...`, for example).

The `extend` method will add the module methods as _class_ methods.  We want them added as _instance_ methods, so we should use the `include` method rather than `extend`.

spec/controllers/concerns/company_people_viewer_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+end
+
+RSpec.describe CompanyPeopleViewer do
```
patmbolger: The _subject_ of the test is the class created at the top of the file (which mimics the concern modules operating in the context of a controller).  So, this line should be:
```
RSpec.describe TestConcernCompanyPeopleClass do
```

spec/controllers/concerns/company_people_viewer_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    it 'no company people present' do
+    expect(TestConcernCompanyPeopleClass.display_company_people no_company_people).to eq(
+                               email: 'bob@hrsolutions.com',
+                               phone: '011 555 555',
+                               roles: 'Admin',
+                               status: 'Active',
+                               company: 'HR Soultions')
+  end
+  end
+end
```
patmbolger: To better structure the tests, I suggest
1) Create your test objects in the top-most describe block, before the actual test blocks.  This is a clean way to declare the objects your tests will use, and doesn't clutter the test blocks themselves  with object create statements.  Something like this:
```
RSpec.describe TestConcernCompanyPeopleClass do
  # Create test objects - a company and company people
  let(:company)      { FactoryGirl.create(:company) }
  let(:cmpy_person1) { FactoryGirl.create(:company_contact, company: company) }
  let(:cmpy_person2) { FactoryGirl.create(:company_contact, company: company) }
  let(:cmpy_person3) { FactoryGirl.create(:company_contact, company: company) }
```
2. Now you can use these objects in your test blocks, for example this tests one of ways that the `display_company_people` method can be called:
```
describe '#display_company_people' do
    it 'returns all company people for a specified company' do
      expect(subject.display_company_people('company-all', company))
        .to match_array [cmpy_person1, cmpy_person2, cmpy_person3]
end
```
Note that I am using `subject` above.  This is an implicit reference to an instance of the class declared in the outer-most block. (http://www.relishapp.com/rspec/rspec-core/v/3-5/docs/subject/implicitly-defined-subject).  So, subject == instance of `TestConcernCompanyPeopleClass`.  Now, we can access the methods in the concerns module by calling them on `subject`.

patmbolger: Note that to test calling the `display_company_people` method with the parameter `my-company-all` you'll have to stub out the return value of `pets_user`.  See `spec/controllers/concerns/task_spec.rb` for an example of how to do that.

patmbolger: Note also that you can access the constant `FIELDS_IN_PEOPLE_TYPE` (for test purposes) by referencing that constant in the test class, that is `TestConcernCompanyPeopleClass::FIELDS_IN_PEOPLE_TYPE`.

This will return the hash: 
```
{:"my-company-all"=>[:full_name, :email, :phone, :roles, :status], :"company-all"=>[:full_name, :email, :phone, :roles, :status]}
```

lollypop27: @patmbolger thanks for this in depth feedback really appreciate it.

patmbolger:
Looks good - merging.

PR #547, opened 2017-03-27
Closes AgileVentures/MetPlus_tracker#552

hin101:
Add test for exceptions and clean up spec file to comply with Rubocop.

### DIFF
--- a/spec/models/resume_spec.rb
+++ b/spec/models/resume_spec.rb
@@ -15,40 +15,36 @@ RSpec.describe Resume, type: :model do
   end
 
   describe 'Resume instance' do
-
     it 'is valid with all required fields' do
-
       stub_cruncher_authenticate
       stub_cruncher_file_upload
 
       file = fixture_file_upload('files/Janitor-Resume.doc')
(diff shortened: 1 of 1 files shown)


spec/models/resume_spec.rb
```diff
@@ -88,14 +81,23 @@
...
+    it 'raises an error when the external cruncher raises an exception' do
+      stub_cruncher_file_upload_error
+      expect do
+        file = fixture_file_upload('files/Test File.zzz')
+        resume = Resume.new(file: file,
+                            file_name: 'Admin-Assistant-Resume.pdf',
+                            job_seeker_id: job_seeker.id)
+        resume.save
+      end .to raise_error(RuntimeError)
+    end
```
patmbolger: This test confirms that the cruncher exception is re-raised by the `save` method (in the rescue clause).

We also need to test:
1) An error message is added to the model instance
2) The model instance is destroyed (can be checked with `destroyed?`)

hin101:
Will reopen in new PR

PR #551, opened 2017-04-01
Job#show back button - closes AgileVentures/MetPlus_tracker#574

Thomascountz:
I wrote two new private methods to store the `request.url` in the `session` hash when a user navigates the `job#index` page. 

I looked into simply doing something like this:
```
<%= link_to :back %>
```
But found that this, 1) emulates the browser's back button. This means that if a user found a job via a google search and they pressed the `back` button, they would be taken off site. And, 2) does not return a user to the correct "paginated" page.


patmbolger:
@Thomascountz - this change does not work if the user comes at a job from another starting point - such as the job seeker home page, for instance.

Also, this mechanism is complex and unneeded, since it is attempting to do what turbolinks already does for us.

I think the solution is just to remove the "Back" button from the job view altogether.  The user can use the back function in the browser to go back (and the prior page is recovered with all pagination links consistent with their prior positions).

patmbolger:
@Thomascountz - yes, please submit the changed code to this same PR.

patmbolger:
Looks good - merging.

PR #552, opened 2017-04-03
Closes AgileVentures/MetPlus_tracker #560

joesasson:
This adds a test for a failed email validation. 

Due to the failing validation, I had to instantiate a new instance of user inside the `it` block, because ActiveRecord was not letting the test run. 

This test only covers a case where the non-valid email resembles a valid one, which would populate the `did_you_mean` field of the API response. A case where that field is not populated (i.e. if the email doesn't resemble anything) is not covered by this test.

Thanks to @Thomascountz for the assistance. This is my first PR ever, so please go easy on me.

### DIFF
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -84,8 +84,8 @@ RSpec.describe User, type: :model do
 
          user = FactoryGirl.build(:user, email: 'emailaddress@gmal.com')
          user.valid?
-         expect(user.errors[:email]).to include('is not valid (did you mean\
-                                                  ... myaddress@gmail.com?)')
+         expect(user.errors[:email]).to include(
+           'is not valid (did you mean ... myaddress@gmail.com?)')
        end
      end
    end
--- a/spec/models/user_spec.rb
+++ b/spec/models/user_spec.rb
@@ -84,7 +84,7 @@ RSpec.describe User, type: :model do
 
          user = FactoryGirl.build(:user, email: 'emailaddress@gmal.com')
          user.valid?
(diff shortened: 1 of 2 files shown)


spec/models/user_spec.rb
```diff
@@ -78,6 +78,15 @@
...
          expect(user).to be_valid
        end
+
+       it 'adds an error to object when validation fails' do
+         stub_email_validate_invalid
+
+         user = FactoryGirl.build(:user, email: 'emailaddress@gmal.com')
+         user.valid?
+         expect(user.errors[:email]).to include('is not valid (did you mean ...
+                                                myaddress@gmail.com?)')
```
patmbolger: @joesasson - this test is failing because you have a newline in your expected error message, which should not be there.  To avoid that, you could write this statement like this:
```
expect(user.errors[:email])
  .to include('is not valid (did you mean ... myaddress@gmail.com?)')
```

patmbolger:
@joesasson - good first effort.  Please see my comment on your test.

Also, a couple of things to keep in mind:

1. Please run all tests (rspec and cucumber) locally before submitting your PR (or changes to a PR).  It's always good to see test failures, and fix them, before Semaphoreci tells us about them :)

2. You almost certainly don't want to develop features in your ```development``` branch.  For an overview of a typical dev process, using git, please see this page on the wiki: https://github.com/AgileVentures/MetPlus_PETS/wiki/Developing-a-feature-(or-bug,-chore)

joesasson:
Thanks @patmbolger for reviewing the PR and for the useful feedback. I'm going to redo the PR taking into consideration your suggestions. Thanks for the help!

PR #557, opened 2017-04-06
Closes AgileVentures/MetPlus_tracker#597

patmbolger:


patmbolger:
Not so much redundant as never used.

PR #559, opened 2017-04-09
Remove partial  list.html #603

pcaston:
Removed jobs/_list.html.haml and retested


patmbolger:
Looks good .... merging

PR #566, opened 2017-04-15
Closes AgileVentures/MetPlus_tracker#605

hin101:
Add Job Seeker from within application.

### DIFF
--- a/app/views/agency_people/home.html.haml
+++ b/app/views/agency_people/home.html.haml
@@ -1,6 +1,10 @@
 - provide(:person_home_active, 'active')
 
 .row
+  .col-sm-10.col-sm-offset-1.margin-top
+    = link_to "Add Job Seeker", new_job_seeker_path, class: 'btn btn-primary btn-lg'
+
+.row
   .col-sm-10.col-sm-offset-1.tasks.margin-top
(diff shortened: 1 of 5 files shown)


app/views/agency_people/home.html.haml
```diff
@@ -1,6 +1,10 @@
 - provide(:person_home_active, 'active')
 
 .row
+  .col-sm-10.col-sm-offset-1.margin-top
+    = link_to "Add Job Seeker", new_job_seeker_path, class: 'btn btn-primary btn-lg'
+
+.row
```
patmbolger: I would like to see the new button be less obtrusive, and be positioned below the tasks tables and be more clearly associated with the job seekers table - something like this:

screen shot 2017-04-17 at 10 51 47 am

You can achieve this by adding the 3rd statement shown here:
```
.row
  .col-sm-10.col-sm-offset-1.agency-people-home
    .pull-right
      = link_to 'Add Job Seeker', new_job_seeker_path,
                class: 'btn btn-primary btn-sm pull-right'
```
Note that the first two statements above already exist, and are lines 50 and 51.

patmbolger:
Looks good ... merging

PR #569, opened 2017-04-23
Add company person as ca cc #446

pcaston:
Re-submitting after adding controller rspec tests

### DIFF
--- a/spec/controllers/company_registrations_controller_spec.rb
+++ b/spec/controllers/company_registrations_controller_spec.rb
@@ -258,6 +258,13 @@ RSpec.describe CompanyRegistrationsController, type: :controller do
       it 'sets User approved to false' do
         expect(assigns(:company).company_people[0].user.approved).to be false
       end
+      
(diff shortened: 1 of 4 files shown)


spec/controllers/company_registrations_controller_spec.rb
```diff
@@ -259,6 +259,13 @@
         expect(assigns(:company).company_people[0].user.approved).to be false
       end
 
+      it 'sets User roles to CC and CA' do
+        expect(assigns(:company).company_people[0].company_roles[0].role)
+          .to eq 'Company Contact'
+        expect(assigns(:company).company_people[0].company_roles[1].role)
+          .to eq 'Company Admin'
+      end
```
patmbolger: This test could break if we change the order of adding roles for a new company admin/contact.  Suggest instead comparing `company_roles` array to expected array contents using `contains_exactly`.

pcaston:
Changed to make it order independent as requested.

PR #572, opened 2017-04-25
Add tests url validator #601

pcaston:
Tests added in validators/url_validator_spec.rb

### DIFF
--- a/spec/validators/state_validator_spec.rb
+++ b/spec/validators/state_validator_spec.rb
@@ -1,37 +1,28 @@
 require 'rails_helper'
+require 'support/validator_helper'
+include ValidatorHelper
 
 RSpec.describe StateValidator do
-  let(:model) do
-    Class.new do
-      include ActiveModel::Validations
-
-      attr_accessor :test_State
-
-      def self.model_name
-        ActiveModel::Name.new(self, nil, 'TestModel')
-      end
-    end
-  end
-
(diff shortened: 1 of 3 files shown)


spec/validators/state_validator_spec.rb
```diff
@@ -0,0 +1,39 @@
...
+    Class.new do
+      include ActiveModel::Validations
+
+      attr_accessor :test_State
+
+      def self.model_name
+        ActiveModel::Name.new(self, nil, 'TestModel')
+      end
+    end
+  end
```
patmbolger: I like this idea ..... but I think we should DRY it out by extracting into a helper module, perhaps in `support/validator_helper.rb`, and then include that module in both the URL and state validator specs.

The module could look like this:
```
module ValidatorHelper
  def test_model_class
    Class.new do
      include ActiveModel::Validations

      attr_accessor :test_attr

      def self.model_name
        ActiveModel::Name.new(self, nil, 'TestModel')
      end
    end
  end
end
```
Then, in each validator you need declare your model class as:
```
let(:model) { ValidatorHelper::test_model_class }
```
Note that the attribute under test has been made generic: `test_attr`, so your validators have to reference that.

pcaston: OK I'll take a look

patmbolger:
Changes look good - merging.

PR #544, opened 2017-03-23
Closes AgileVentures/MetPlus_tracker#548

joaopapereira:
Add unit tests for #index, #assign and #tasks

PR #574, opened 2017-05-10
Add job status as searchable field #544

pcaston:
Submitted pull request for comment.

### DIFF
--- a/app/views/jobs/index.html.haml
+++ b/app/views/jobs/index.html.haml
@@ -86,7 +86,7 @@
         = f.collection_select :status_in, Job.statuses, :last, :first,
                {include_blank: false},
                {multiple: true, size: 5,
-                class: 'form-control', style: 'width: 100%;'}              
+                class: 'form-control', style: 'width: 100%;'}     
 
     .form-group
       .col-sm-3.col-sm-offset-3
@@ -192,10 +192,10 @@
                 %strong Status
               %td
                 All possible statuses for jobs are listed with
-                %i active 
+                %i active
                 as the default.
                 %br
-                You can select another status for searching by clicking
+                You can select multiple statuses by repeatedly clicking
                 in the search box.
 
       .modal-footer
--- a/app/views/jobs/index.html.haml
+++ b/app/views/jobs/index.html.haml
@@ -81,6 +81,13 @@
                 class: 'form-control', style: 'width: 100%;'}
 
     .form-group
+      = f.label :status_eq, 'Status', class: 'control-label col-sm-3'
+      .col-sm-6
+        = f.collection_select :status_eq, Job.statuses, :last, :first,
+              {multiple: true, size: 5,
+                class: 'form-control', style: 'width: 100%;'}              
+
+    .form-group
       .col-sm-3.col-sm-offset-3
         = submit_tag 'Search Jobs', class: 'btn btn-primary'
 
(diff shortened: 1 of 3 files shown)


pcaston:
Re-opening pull request to re-run tests

patmbolger:
Paul - this is good progress.  You'll need to make a few more changes to make it conform with how the rest of the search form works.  This means that the "status" search field will have to use "select2" javascript package to allow for a better UI.

See code comments for specifics.

Once this is complete, the search form should looks like this, for example:

screen shot 2017-05-11 at 8 20 13 am

---
And the resulting search should show all jobs matching (one or more) status values:

screen shot 2017-05-11 at 8 20 48 am

app/views/jobs/index.html.haml
```diff
@@ -81,6 +81,13 @@
                 class: 'form-control', style: 'width: 100%;'}
 
     .form-group
+      = f.label :status_eq, 'Status', class: 'control-label col-sm-3'
+      .col-sm-6
+        = f.collection_select :status_eq, Job.statuses, :last, :first,
+              {multiple: true, size: 5,
+                class: 'form-control', style: 'width: 100%;'}
```
patmbolger: Need to change this code to 1) allow for multiple status values to be selected, and 2) fix an issue with collection_select arguments.

Line 84 should be:
```
= f.label :status_in, 'Status', class: 'control-label col-sm-3'
```
With 'status_in` we are telling Ransack (in controller) that we want to search for any status values that match the values chosen by the user in the associated `collection_select` statement.

Lines 86 - 88 should be:
```
= f.collection_select :status_in, Job.statuses, :last, :first, {},
              {multiple: true, size: 5,
                class: 'form-control', style: 'width: 100%;'}
```
We changed to `status_in` consistent with the label change.  Also, note the empty hash - see the documentation for `collection_select` - it takes two hash args - one for options to `collection_select` itself and one for specifying HTML options for the generated HTML input/select code.  The empty hash above is for the first hash arg.  Without that, the second hash arg was being interpreted as `collection_select` options and thus ignored.

patmbolger: Also, to enable the `status` field to act as a `select2` field, we need to add the fourth line shown here to `app/assets/javascripts/jobs.js`:
```
$('#q_address_city_in').select2();
$('#q_skills_id_in').select2();
$('#q_company_id_in').select2();
$('#q_status_in').select2();
```
(the first 3 lines are already in that file).
This Code runs when the page is loaded, and converts a normal select field to a select2 field.

You can check out `select2` here:  https://select2.github.io.

features/search_jobs.feature
```diff
@@ -160,4 +161,30 @@ Scenario: Search by company
...
+  And I click the "Show Search Form" link
+  And I wait 1 second
+  Then I select "revoked" in select list "Status"
+  And I click the "Search Jobs" button
+  Then I should see "Job6"
+  And I should not see "Job1"
+  And I should not see "Job2"
+  And I should not see "Job3"
+  And I should not see "Job4"
+  And I should not see "Job5"
```
patmbolger: These tests look good but please add a test where the search occurs for _two_ values of status.

patmbolger:
One thing I almost forgot - the help text for the `status` search field should note that multiple values can be selected (like some of the other search fields).

app/views/jobs/index.html.haml
```diff
@@ -179,6 +187,16 @@
                 %br
                 You can select multiple companies by repeatedly clicking
                 in the search box.
+            %tr
+              %td
+                %strong Status
+              %td
+                All possible statuses for jobs are listed with
+                %i active
+                as the default.
```
patmbolger: Actually, there is no default value to `status`.  Please adjust this help text.

patmbolger:
Looks good.

PR #575, opened 2017-05-16
Job search companies regardless of job status #595

pcaston:


pcaston:
Reopen to re-run tests

app/models/company.rb
```diff
@@ -40,14 +39,8 @@ def registration_denied
...
-    companies = []
-    Company.active.order(:name).each do |cmpy|
-      unless cmpy.jobs.where(status: 'active').empty?
-        companies << cmpy
-      end
-    end
-    companies
+  def self.all_with_jobs
+    Company.order(:name).joins(:jobs).uniq.all
   end
```
patmbolger: We need to grab all `active` companies with at least one job.  This query is close but needs to limit the results to active companies.  You could add a scope to the query string that looks something like this:
```
scope :all_active, -> {where(status: Company.statuses[:active])}
```
Note that we're using an `enum` for status, and that provides a `statuses` method that allows us to use the status name (or symbol) for looking up the value of the status as stored in the DB (integer).  http://edgeapi.rubyonrails.org/classes/ActiveRecord/Enum.html

arun1595: We can also do `Company.active` to get the active companies

`ActiveRecord Enum` also provide scopes for all possible values of the `enum` field.

https://apidock.com/rails/v4.2.1/ActiveRecord/Enum

patmbolger: @arun1595 - you're right .... I had forgotten about that.  That is a better way to do create the scope.

pcaston: Applied changes as suggested.

pcaston:
Re-openning to re-run tests

features/search_jobs.feature
```diff
@@ -155,6 +157,7 @@ Scenario: Search by company
   And I should not see "Job4"
   And I click the "Show Search Form" link
   And I wait 1 second
+  Then "Acme Inc." should not be an option for select list "Company"
```
patmbolger: This step confirms that `Acme, Inc.` is not selectable because it _has no jobs_.  

However, we should also add a company to the Background that is _not_ active, but that _does_ have jobs, and then confirm that this company is not available for selection in the search form.

pcaston: good point.

pcaston:
Re-run tests

pcaston:
Re-run tests

pcaston:
Re-run tests

pcaston:
Closed and re-openned to re-run semaphore tests

PR #577, opened 2017-05-27
Job search on active companies only #596

pcaston:
Added active companies predicate and spec test to jobs controller

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -44,13 +44,13 @@ class JobsController < ApplicationController
       @description_words += q_params[:description_cont_all]
     end
 
-    params[:q] = if params[:q].present?
-                   params[:q].merge('company_status_eq' => ['active'])
-                 else
-                   { 'company_status_eq' => ['active'] }
-                 end
+    #  ensure only jobs for active are returned.
+    q_params = if q_params.present?
+                 q_params.merge('company_status_eq' => 'active')
+               else
+                 { 'company_status_eq' => 'active' }
+               end
     @query = Job.ransack(params[:q]) # For form display of entered values
-
     @jobs  = Job.ransack(q_params).result
                 .includes(:company)
                 .includes(:address)
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -10,7 +10,6 @@ class JobsController < ApplicationController
                                         :match_jd_job_seekers]
 
   def index
-
     # Make a copy of q params since we will strip out any commas separating
     # words - need to retain any commas in the form (so user is not surprised)
     q_params = params[:q] ? params[:q].dup : params[:q]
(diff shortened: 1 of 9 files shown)


pcaston:
Re-open to run semaphore tests.

app/controllers/jobs_controller.rb
```diff
@@ -45,6 +44,11 @@ def index
       @description_words += q_params[:description_cont_all]
     end
 
+    params[:q] = if params[:q].present?
+                   params[:q].merge('company_status_eq' => ['active'])
+                 else
+                   { 'company_status_eq' => ['active'] }
+                 end
```
patmbolger: There are at least these problems with this statement:

1) We do _not_ use `params[:q]` to actually execute the jobs search.  We actually use the local var `q_params` (assigned at top of the action method), which is set to a copy of `params[:q]` initially.

We keep `params[:q]` untouched so that when the search form is redisplayed the user sees _exactly_ the text they have entered and the selections they have made (including entering terms separated by commas, spaces, etc.)

2) The logic here should be that we adjust `q_params`, to include only companies that are of `active` status, if:
  a. `q_params` is not `nil`, and
  b. `q_params[:company_id_in]` does not contain one or more company IDs (that is, the user is _not_ including one or more companies in the search).

That logic could begin with this statement:
```
if q_params && q_params[:company_id_in] == ['']
```
(note that ransack will include an array with an empty string if a company is not specified in the search, hence the comparison above)

patmbolger: Also, please make sure to include sufficient spec (controller) and/or cucumber tests to confirm the logic works as expected.  Including:

1) A search that does _not_ include a company will not show jobs associated with companies that are of status `pending_registration`, `inactive`, `registration_denied`).

2) Searches that _include_ a company with one of the above status values _will_ show jobs for that company.

pcaston: Ah I see.  OK will do.

pcaston: Also for the initial jobs index display when q_params will be nil we need to set it to { 'company_status_eq' => ['active'] } otherwise any non-active company jobs will be returned.
When q_params is not nil then merging the above predicate has no effect on the jobs found when one or more companies have been selected, since you can only select active companies.  Do we need to worry? would it perhaps increases response time on the query.

pcaston: I can set up tests to ensure that companies in either the 'pending_registration' or 'registration_denied'  state cannot be selected in the search form but is it possible for such companies to have jobs?

patmbolger: 1) Yes, I see your point re `index` initial load.

2) I agree that we can remove the check for a company search since we have limited search on `company` to only active companies.

3) Good question re the status values `pending_registration` or `registration_denied` ... in today's model such a company cannot have jobs associated with them.  However, in the future I could imagine a process where an `active` company moves to `inactive` (or, perhaps, some `suspended` status) - for whatever reason.  And then a subsequent review process occurs to bring the company back to active status.  In this (admittedly contrived) example, we could have a company in one of those states that also has jobs.

pcaston: Ok I'll set up some rspec tests rather than cucumber .  I think that is more appropriate for what are really white box structural tests of the code.  
BTW would it be possible to set up a hang-out at say 8pm tomorrow night to discuss another story?  The Martin Fowler scrum at 7:00 pm is a bit awkward for me.

patmbolger: I agree that rspec (controller) tests are best for this.

And, yes, lets connect up at 8:00 pm (your time) tomorrow.

PR #580, opened 2017-06-09
Add tests for sessions controller #602

pcaston:
Added further controllers tests to cover the create method in the custom devise sessions controller.  To test that the cookies expiring date is set correctly the cookie method is stubbed.


patmbolger:
Nicely done - merging

PR #581, opened 2017-06-11
allow agency persons to download resumes

pcaston:
Changed the policy and controller to enable Agency persons to download jobseeker resumes.


patmbolger:
Looks good .... merging.

PR #583, opened 2017-06-18
Add tests for people invitation controller #551

pcaston:
Set up shared example 'accepts invitation' and used it to test the accept_resource method of the people_invitations_controller.rb for agency and company people types.


patmbolger:
Looks great ... merging.

PR #584, opened 2017-06-19
download JS resume button changes to view, controller and feature test

pcaston:
Added condition to Job_seeker_controller that a job seeker must have at least one resume before the download button will appear.  Adjusted the size of the download button to be the same as that of the 'Assign Me' button.  Added a feature test to verify that the download button does not appear for a job seeker with no resume.


patmbolger:
Looks good - merging

PR #586, opened 2017-06-26
Remove unnecessary text

arun1595:
Fixes AgileVentures/MetPlus_tracker#612

### DIFF
--- a/app/views/company_registrations/new.html.haml
+++ b/app/views/company_registrations/new.html.haml
@@ -2,13 +2,5 @@
 
 %h2 Request Company Registration
 
-%p Thank you for your interest in MetPlus!
-
-%p brief promotional message about the mission goes here]
-
-%p
-  Please enter information about your organization and a means of
-  contact using the form below.
-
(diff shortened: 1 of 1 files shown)


app/views/company_registrations/new.html.haml
```diff
@@ -2,13 +2,5 @@
 
 %h2 Request Company Registration
 
-%p Thank you for your interest in MetPlus!
-
-%p brief promotional message about the mission goes here]
```
joaopapereira: After talking with Chet, this should be the only line to remove, the rest should stay

app/views/company_registrations/new.html.haml
```diff
@@ -2,13 +2,5 @@
...
-
-%p
-  Please enter information about your organization and a means of
```
arun1595: Isn't this line superfluous?

joaopapereira: hmm, ok, lets see how it flies without that line.

PR #588, opened 2017-06-26
Change default font and global font size

arun1595:


patmbolger:
Hi @arun1595 - this looks good.   I was wondering how this story came about.  Was it based upon a conversation with someone on the team, or perhaps your own sense of how the UI should look?

I like the larger font, but just want to understand the motivation.  When we merge this, I may want to weak other views so as to better accommodate the larger font (in some tables, for example).

arun1595:
@patmbolger 

> Was it based upon a conversation with someone on the team, or perhaps your own sense of how the UI should look?

It is the latter.

I now understand what you meant regarding the font size for the table content. Yeah 16px seems too big. Let me change it. 

What do you think about using Open Sans as the global font? It was my personal choice, but I would love to hear your thoughts on choosing a new font-face.

/cc @joaopapereira

patmbolger:
@lollypop27 - could you please review and give your opinion on the font (and size) to be used as the default font in the app?

lollypop27:
@patmbolger and @arun1595 Love OpenSans my fav font, think a 14px font size should be ok

patmbolger:
@arun1595 - please set the default font-size for the body to 14px and then I can merge.

arun1595:
@patmbolger It should be 14px by default. Do you want me to explicitly define it?

patmbolger:
@arun1595 - yes, please set the size explicitly.  Perhaps its. browser-specific setting, but I am seeing 16 px; when I stand up your feature branch.

screen shot 2017-07-13 at 12 03 00 pm

arun1595:
@patmbolger Done :smile:

patmbolger:
Thanks for adjusting .... merging.

PR #592, opened 2017-07-13
Refactor pets_user method

arun1595:


patmbolger:
Merged a copy of this after confirming all tests pass.

PR #593, opened 2017-07-14
[TEST] Pcaston add tests for recaptcha service #599

patmbolger:
Test PR with changes to make all tests work.


features/step_definitions/users_steps.rb
```diff
@@ -11,14 +12,15 @@
...
-   visit  "/users/edit.#{user.id}"
+  user = User.find_by_first_name!(first_name)
+  visit "/users/edit.#{user.id}"
 end
 
-Then(/^I should verify the change of first_name "(.*?)", last_name "(.*?)" and phone "(.*?)"$/) do |first_name, last_name, phone|
-    user = User.find_by_first_name(first_name)
-    expect(user.last_name).to    eql last_name
-    expect(user.phone).to        eql phone 
+Then(/^I\sshould\sverify\sthe\schange\sof\sfirst_name\s"(.*?)",\s
```
patmbolger: This is OK but I don't think it is necessary and it makes the regex harder to read.

The pattern `/text1 text2/` is treated exactly the same as `/text1\stext2/`, so the addition of `\s` is not needed.

features/step_definitions/users_steps.rb
```diff
@@ -33,10 +35,26 @@
...
+  #  since the intent is to prevent automated entry that simulates a human.
+  # Thus, 1) we won't attempt to "check the box", and 2) we will override
+  # server-side verification to just return true (== verified by Google).
+
+  class RecaptchaService
+    def self.verify(_, _)
+      true
+    end
+  end
+end
```
patmbolger: The comment above explains where I ended up with this step, after trying to simulate a user action (checking the box).  The stub method did not work, but simply overriding the `verify` method does the job.

features/step_definitions/web_steps.rb
```diff
@@ -162,6 +162,7 @@ def search_text(text)
 end
 
 And(/^show me the page$/) do
+  save_and_open_page
```
patmbolger: This seems to have been dropped in a prior PR.  Adding it back ....

features/support/env.rb
```diff
@@ -68,40 +71,18 @@
...
-      # debug: true
-    )
-  end
-  Capybara.javascript_driver = :poltergeist
+Capybara.register_driver :selenium do |app|
+  Capybara::Selenium::Driver.new(
+    app,
+    browser: :chrome,
+    args: ['headless']
+  )
```
patmbolger: Some restructuring to simplify, and also add Chrome as browser option for the Selenium web driver.

features/pages.feature
```diff
@@ -0,0 +1,15 @@
+Feature: a user clicking contact is verified by a recaptcha
+
+  As prospective user of PETS
+  I want to make contact and send a message
+
+@selenium_browser
```
patmbolger: I had to switch this test to use Selenium, driving a Chrome browser in "visible" mode (both poltergeist, and headless Chrome, failed to work).

I also have to use special recaptcha keys - these were added to the `application.yml` file - will DM you on slack with that updated file.

See: https://developers.google.com/recaptcha/docs/faq (section on automated tests)

features/pages.feature
```diff
@@ -0,0 +1,15 @@
...
+Scenario: click the contact link
+  Given I am on the home page
+  And I click the first "Contact" link
+  And I fill in "Full Name" with "Charlie Bucket"
+  And I fill in "Email" with "cb@gmail.commetplus.org"
+  And I fill in "Message" with "Hi Metplus"
+  And I click the recaptcha
+  And I click the "Send Message" button
+  Then I should see "Your message was sent successfully!"
```
patmbolger: I suggest also adding a "sad" path - that is, the user fills out the info but fails to check the "not a robot" box.

spec/support/service_stub_helpers.rb
```diff
@@ -163,9 +163,9 @@ def stub_cruncher_job_create_retry_auth_fail
         .to_raise(RestClient::Unauthorized)
     end
 
-    def stub_cruncher_job_create_fail(result_code)
+    def stub_cruncher_job_create_fail(resultCode)
```
pcaston: changed to snake case

features/pages.feature
patmbolger: Also, I forgot to mention that one of the problems you were having with using poltergeist was that phantomjs support for ES2015 (aka ES6) is still a work-in-progress.  (https://github.com/ariya/phantomjs/issues/14506).

(Poltergeist uses phantomjs to support js code).

@arun1595 is using ES6 in some of his code (e.g. `promise`).  Poltergeist _silently_ fails when ES6 is encountered, so there is no indication of what caused the test to fail.  Going to Chrome as the Selenium browser fixed this issue (but, for these tests, only when I used Chrome in visible (non-headless) mode.

PR #601, opened 2017-07-20
Pagination conversion: job search view

patmbolger:
The documentation: https://github.com/AgileVentures/MetPlus_PETS/wiki/Add-JS-pagination-to-a-view-(preferred-approach)


app/helpers/application_helper.rb
```diff
@@ -51,8 +51,19 @@ def show_stars(rating)
     render_stars(rating)
   end
 
-  def paginate(collection, params = {})
-    will_paginate collection, params.merge(renderer:
-      RemoteLinkPaginationHelper::BootstrapLinkRenderer)
+  # def paginate(collection, params = {})
+  #   will_paginate collection, params.merge(renderer:
+  #     RemoteLinkPaginationHelper::BootstrapLinkRenderer)
+  # end
```
patmbolger: This method is associated with the module described in another comment.  It should be deleted before the final version of this PR.

app/helpers/remote_link_pagination_helper.rb
```diff
@@ -1,5 +1,5 @@
 module RemoteLinkPaginationHelper
-  class BootstrapLinkRenderer < WillPaginate::ActionView::BootstrapLinkRenderer
+  class BootstrapLinkRenderer # < WillPaginate::ActionView::BootstrapLinkRenderer
     def link(text, target, attributes = {})
       attributes['data-remote'] = true
       super
```
patmbolger: This module is not in use anymore (it was associated with an earlier approach to providing JS-driven pagination for the searched jobs view).

However, if I comment it out completely, I get an error when I start the app (it says it cannot find the ```RemoteLinkPaginateHelper``` module).

PR #607, opened 2017-08-02
Closes AgileVentures/MetPlus_tracker#628

hin101:
Add Years of experience field to jobs

### DIFF
--- a/app/views/jobs/_form.html.haml
+++ b/app/views/jobs/_form.html.haml
@@ -31,8 +31,8 @@
   .col-sm-3.control-label
     = f.label :years_of_experience, 'Years Of Experience'
   .col-sm-6
-    = f.collection_select :years_of_experience, Job::YEARS_OF_EXPERIENCE_OPTIONS, :to_i, :to_i,
-            {include_blank: true}, class: 'form-control'
+    = f.collection_select :years_of_experience, Job::YEARS_OF_EXPERIENCE_OPTIONS,
+      :to_i, :to_i, {include_blank: true}, class: 'form-control'
 
 .form-group
   .col-sm-3.control-label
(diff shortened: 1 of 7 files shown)


app/views/jobs/_job_info.html.haml
```diff
@@ -26,6 +26,12 @@
     %tr
       %td Job ID:
       %td=@job.company_job_id
+    - unless @job.years_of_experience.blank?
+      %tr
+        %td Years Of Experience:
+        %td
+          = @job.years_of_experience
+          #{@job.years_of_experience == 20 ? 'Or More Years' : ' Years'}
```
joaopapereira: It would be a nice touch if we could use pluralize the Years part of this, something like 

#{@job.years_of_experience == 20 ? 'Or More Years' : pluralize(@job.years_of_experience, 'Year')}

joaopapereira: An aside, we should not display the experience or say something like: 'Not Specified' when we have a job with nil on the experience

app/views/jobs/_form.html.haml
```diff
@@ -29,6 +29,13 @@
 
 .form-group
   .col-sm-3.control-label
+    = f.label :years_of_experience, 'Years Of Experience'
+  .col-sm-6
+    = f.collection_select :years_of_experience, Job::YEARS_OF_EXPERIENCE_OPTIONS, :to_i, :to_i,
```
joaopapereira: Can you split this line?

app/views/jobs/_job_info.html.haml
hin101: I have put it in the `unless @job.years_of_experience.blank?` do we want to display the experience when it is blank but with not specified instead?

joaopapereira:
not mergeable as one test is failing in cucumber

joaopapereira:
@hin101 can you corret the cucumber test and resubmit the PR?

PR #615, opened 2017-08-15
addes two new locales

gaare:
This is a suggestion for adding es locale in order to PETS be available on Spain and also some latin countries.


patmbolger:
There is no need for i18N translations at this point.

PR #616, opened 2017-08-16
Closes AgileVentures/MetPlus_tracker#623

hin101:
Functionality done but having issues with Cucumber Tests.


app/assets/javascripts/jobs.js
```diff
@@ -131,3 +131,7 @@ $(function () {
   });
 
 });
+
+$(function() {
+  $(".job-type-multiple").select2();
```
patmbolger: Don't need a separate DOM-ready function for this.  Could add this statement to the existing function - see similar statements starting at line 93.

app/models/job.rb
```diff
@@ -4,11 +4,11 @@ class Job < ActiveRecord::Base
   belongs_to :company_person
   belongs_to :address
   belongs_to :job_category
+  has_and_belongs_to_many :job_types
 
   has_many   :job_skills, inverse_of: :job, dependent: :destroy
   has_many   :skills, through: :job_skills
-  accepts_nested_attributes_for :job_skills, allow_destroy: true,
-                                             reject_if: :all_blank
+  accepts_nested_attributes_for :job_skills, allow_destroy: true
```
patmbolger: why drop the `reject_if:` argument?

db/migrate/20170814193715_add_default_job_types.rb
```diff
@@ -0,0 +1,11 @@
+class AddDefaultJobTypes < ActiveRecord::Migration
```
patmbolger: Unless we need to convert existing data in the DB, I would rather _not_ add data to a table with a migration (which should be limited to schema changes if at all possible).

I suggest adding this data create logic to `seeds.rb` (the part where we seed "production" data (create default Agency, create agency and company roles).

app/models/job.rb
hin101: Ah ill add it back in I think I accidently deleted this line

patmbolger:
Looks good, and thanks for making the requested changes.

PR #617, opened 2017-08-17
Closes AgileVentures/MetPlus_tracker#630

hin101:
I have added a csv with 5 generic skills. May need to get list from Client before merging.

** For future updates all we need to do is add to CSV and run rake task `rake job_skills:import_generic` **


patmbolger:
Looks good - merging.

PR #621, opened 2017-09-02
Closes AgileVentures/MetPlus_tracker#633

hin101:


app/views/jobs/_form.html.haml
```diff
@@ -73,12 +73,12 @@
...
-    = f.label :shift
-  .col-sm-1
-    = f.collection_select :shift, Job::SHIFT_OPTIONS, :to_s, :to_s,
-            {prompt: true}, class: 'form-control'
+                          
+.form-group.form-inline
+  .col-sm-3.control-label
+    = f.label :job_shift_ids, 'Shift'
+  .col-sm-5
+    = f.collection_select :job_shift_ids, JobShift.all, :id, :shift, {}, class: 'job-shift-multiple', multiple: true
```
patmbolger: If I change line 70-81 to this:
```
.form-group.form-inline
  .col-sm-3.control-label
    = f.label :job_type_ids, 'Job Type(s)'
  .col-sm-3
    = f.collection_select :job_type_ids, JobType.all, :id, :job_type, {},
                          class: 'job-type-multiple', multiple: true

  .col-sm-1.control-label
    = f.label :job_shift_ids, 'Shift'
  .col-sm-4
    = f.collection_select :job_shift_ids, JobShift.all, :id, :shift, {},
                          class: 'job-shift-multiple', multiple: true
```
then I get this layout:

screen shot 2017-09-08 at 9 59 17 am

Which I think looks a little better.

app/controllers/jobs_controller.rb
```diff
@@ -378,7 +378,7 @@ def job_params
     params.require(:job).permit(:description, :shift, :company_job_id,
```
patmbolger: Should remove `:shift`

db/migrate/20170902043316_create_jobs_jobs_shifts_join_table.rb
```diff
@@ -0,0 +1,8 @@
+class CreateJobsJobsShiftsJoinTable < ActiveRecord::Migration
+  def change
+    create_join_table :jobs, :job_shifts do |t|
+      t.index :job_id
+      t.index :job_shift_id
+    end
+  end
+end
```
patmbolger: In this or a separate migration file you'll have to:

1. Convert any existing `shift` values in the `jobs` table to this new structure
2. Drop the existing `shift` column in `jobs` when above task is complete

hin101: Won't be able to do as a migration because the initial data is populated in db.seeds, the only way I can think of doing this is we populate the data in the migration, convert then drop.

What are your thoughts?

patmbolger: Yes, I understand.  But for this, it's OK to just hard-wire the conversion values into the migrations file.  That is, existing value 'Day' maps to 'Day', etc., and 'Evening' to 'Afternoon'.

In the seeds file (in the production data section) we'll use `find_or_create_by` to create any remaining shift values (in the `shifts` table), and prevent duplicate records.  I updated `seeds.rb` yesterday to use that method with all of the production data.

app/views/jobs/_job_info.html.haml
```diff
@@ -47,6 +47,10 @@
       %tr
         %td Job Type:
         %td= @job.job_types.map(&:job_type).join(", ")
+    - unless @job.job_shifts.empty?
+      %tr
+        %td Job Shifts:
+        %td= @job.job_shifts.map(&:shift).join(", ")
```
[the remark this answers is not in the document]
patmbolger: I am getting an exception when I _show_ a job due to line 36 in this file:
```
      %td=@job.shift
```
Looks like a hold-over from prior version of this file.

When I comment out lines 34-36 it looks good.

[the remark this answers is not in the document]
patmbolger: @hinesh - I went ahead and removed lines 34-36.  I need to get this in so I can merge with another feature branch that I am working on (that also involves this file).

PR #635, opened 2017-10-06
Add JS pagination for company jobs in company person homepage

arun1595:
Closes AgileVentures/MetPlus_tracker#636

### DIFF
--- a/app/controllers/concerns/jobs_viewer.rb
+++ b/app/controllers/concerns/jobs_viewer.rb
@@ -6,15 +6,13 @@ module JobsViewer
     m.helper_method :job_fields
   end
 
-  def display_jobs(job_type, per_page = 10)
+  def display_jobs(job_type)
     case job_type
     when 'my-company-all'
-      collection = Job.order(:title).find_by_company(pets_user.company)
+      Job.order(:title).find_by_company(pets_user.company)
     when 'recent-jobs'
-      collection = Job.new_jobs(Time.now - 3.weeks).order(created_at: :desc)
+      Job.new_jobs(Time.now - 3.weeks).order(created_at: :desc)
     end
-    return collection if collection.nil?
-    collection.paginate(page: params[:jobs_page], per_page: per_page)
   end
 
   FIELDS_IN_JOB_TYPE = {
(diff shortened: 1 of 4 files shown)


app/controllers/concerns/jobs_viewer.rb
```diff
@@ -6,15 +6,15 @@ def self.included(m)
     m.helper_method :job_fields
   end
 
-  def display_jobs(job_type, per_page = 10)
```
patmbolger: Yes, this is a cleaner way to handle this helper method - drop the pagination stuff and apply that in the controller.  I will adapt my current PR in the same way.

app/views/jobs/_list_jobs.html.haml
```diff
@@ -1,63 +1,69 @@
...
+                  = job.number_applicants
+              - when :company
+                %td
+                  = job.company.name
+              - when :description
+                %td
+                  = truncate(job.description, length: 200)
+              - when :posted
+                %td
+                  #{time_ago_in_words(job.created_at)} ago
```
patmbolger: Please add sort capability to the remaining columns (besides `title`).  (my WIP PR just did it for `title` as an example - that will also have to provide sorting capability on all columns).

arun1595: @patmbolger Done. PTAL :smile:

app/views/jobs/_list_jobs.html.haml
```diff
@@ -1,63 +1,75 @@
...
+                %th.strong
+                  = sort_link(query, :status, {},
+                    class: 'jobs-pagination', remote: true)
+              - when :updated_at
+                %th.strong
+                  = sort_link(query, :updated_at, 'Last update', {},
+                    class: 'jobs-pagination', remote: true)
+              - when :poster
+                %th.strong
+                  = sort_link(query, :company_person_last_name, 'Poster', {},
```
patmbolger: Sorting by last_name is not working - because `last_name` is _not_ a column in the `company_people` table.

I tried this instead:
```
= sort_link(query, :company_person_user_last_name, ...
```
and it worked!  (I told you Ransack was amazing - I don't even think this second-level association traversal is documented from what I can find).

arun1595: haha, it is indeed awesome. My bad, I didn't test it

arun1595:
@patmbolger Done. 

https://waffle.io/AgileVentures/MetPlus_tracker/cards/59dd4db0eb3dfc009d8d11fe

PR #643, opened 2017-10-11
Add language proficiency field

hin101:
Closes AgileVentures/MetPlus_tracker#644

### DIFF
--- a/app/views/jobs/_form.html.haml
+++ b/app/views/jobs/_form.html.haml
@@ -129,6 +129,12 @@
 
 .form-group
   .col-sm-3.control-label
+    = f.label :language_proficiency
+  .col-sm-7
+    =f.text_area :language_proficiency, size: "40x5", class: 'form-control'
+
+.form-group
+  .col-sm-3.control-label
     %b Questions
   .col-sm-7
     #questions{ style: style }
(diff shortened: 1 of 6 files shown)


app/views/jobs/_form.html.haml
```diff
@@ -129,6 +129,12 @@
 
 .form-group
   .col-sm-3.control-label
+    = f.label :language_proficiency
+  .col-sm-7
+    =f.text_area :language_proficiency, size: "40x5", class: 'form-control'
```
patmbolger: Use `text_field` instead of `text_area` - we're looking for a sentence or two, so we don't need a large space for typing into.

app/views/jobs/_form.html.haml
```diff
@@ -129,6 +129,12 @@
 
 .form-group
   .col-sm-3.control-label
+    = f.label :language_proficiency
```
patmbolger: The label should have both words capitalized ("Language Proficiency").

app/views/jobs/_job_info.html.haml
```diff
@@ -65,6 +65,11 @@
             = q
             %br
 
+    - unless @job.language_proficiency.nil?
+      %tr
+        %td Language Proficiency
+        %td= @job.language_proficiency
```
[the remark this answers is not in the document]
patmbolger: Should be "Language Proficiency:"  (you need to add the colon).

PR #648, opened 2017-10-15
Override save! method in resume model

arun1595:
Fixes https://github.com/AgileVentures/MetPlus_tracker/issues/661

### DIFF
--- a/spec/models/resume_spec.rb
+++ b/spec/models/resume_spec.rb
@@ -81,6 +81,7 @@ RSpec.describe Resume, type: :model do
     end
 
     it 'fails with invalid model and invalid file type' do
+      file = fixture_file_upload('files/Test File.zzz')
       resume = Resume.new(file: file,
                           file_name: 'nil',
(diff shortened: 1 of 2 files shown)


spec/models/resume_spec.rb
```diff
@@ -81,7 +81,6 @@
     end
 
     it 'fails with invalid model and invalid file type' do
-      file = fixture_file_upload('files/Test File.zzz')
       resume = Resume.new(file: file,
```
patmbolger: Removing this line causes this test to fail.

arun1595: I am sorry, I don't understand.

Why do you remove that line?

patmbolger: @arun1595 - I _didn't_ remove that line.  That line was removed in your PR :)

arun1595: My bad. I misunderstood you were talking about line the comment is made on.

patmbolger:
Looks good.

PR #655, opened 2017-11-20
Complete assign jd to js

joaopapereira:
Closes AgileVentures/MetPlus_tracker#652


app/services/agency_people_service.rb
```diff
@@ -0,0 +1,41 @@
...
+  end
+
+  private
+
+  def assign_job_developer_to_job_seeker(job_seeker, job_developer)
+    obj = Struct.new(:job_seeker, :agency_person)
+
+    raise NotAJobDeveloper, '' unless job_developer.is_job_developer? job_developer.agency
+    job_seeker.assign_job_developer job_developer, job_developer.agency
+    Event.create(:JD_SELF_ASSIGN_JS, obj.new(job_seeker, job_developer))
```
patmbolger: There are two ways that an AP can be assigned to a JS (as JD or CM):

1) They can self-assign from the JS `show` view, or,

2) The AA can assign them by editing the AP (see `agency_person_controller#update`).

What is happening here is _not_ a AP-to-JS assignment, but rather the AA is assigning the "JS needs a JD/CM" _task_ to the AP.  It is then up to the AP to assign _themselves_ to the JS in the required role.

You can see this by assigning such a task to an AP (logged in as admin).  The result shows that the AP has a task assigned to themselves _which requires them to self-assign_:

screen shot 2017-11-26 at 8 20 44 am

Clicking on a JS name will show that the JS does not yet have an AP assigned:

screen shot 2017-11-26 at 8 22 55 am

`agency_person_controller#assign_job_seeker` is used when the AP self-assigns to a JS (in either role).  So, that action should not be used for this situation.

Perhaps we need another action, say, `assign_task_to_ap`, which performs the assignment.  Then, later (or part of this story) we can create an event that serves to notify the AP that a _task_ has been assigned to them.

app/controllers/agency_people_controller.rb
```diff
@@ -46,44 +46,35 @@ def update
...
-
-      flash[:notice] = 'Agency person was successfully updated.'
-      redirect_to agency_person_path(@agency_person)
+      assign_agency_person_to_job_seeker = AssignAgencyPersonToJobSeeker.new
+      begin
+        assign_agency_person_to_job_seeker.call(
+          JobSeeker.where(id: jd_job_seeker_ids),
+          :JD,
+          @agency_person,
+          false
```
patmbolger: This method call (and the similar one below) get passed all JS id's that come in via `params`.  The `params` will contain 1) all prior-existing JS's that had been assigned to the AP, 2) any _new_ JS's assigned to this AP, and 3) all assigned JS that are to be deleted from the association (unassigned from AP).

The prior logic created an assignment event only for the newly-assigned JS.  This logic results in that event being created for all new as well as existing (assigned) JS.  This means the notification for existing JS-to-AP assignments will go out to both the JS and the AP.

This screenshot shows a JD home page (Jane).  Jane already had JS "Frank Williams" assigned to her.

In another browser, as an admin, I assigned a second JS - "Rose Bradtke" - to Jane.  You can see that notification _also_ went out for the prior-existing association.

joaopapereira:
@patmbolger can you review the new change? Thanks

patmbolger:
The way this works is essentially the same as the way the current production code works.

That is, the admin assigns the "JS needs a JD" task is _assigned_ to a JD.  Then, that task does not appear in the admins list of "unassigned tasks"

The actual JS<>JD assignment is _not_ occurring in this PR.

I actually think the way the current prod code works is fine.  We have the mechanism described above (for self-assignment of the JD to JS) as well as the ability of the admin to directly assign a JD to a JS via the AP edit/update capability.

However, the way the story was written I got the impression that you wanted the admin to be able to assign the JS to the JD simply by the same action as assigning the _task_ to the JD. 

Is the way this story works consistent with your intent (and the story is now quite accurate)?

This PR, in any case, builds some infrastructure that will (presumably) be useful for the overall goal of auto-closing tasks with the association business action occurs.

Please advise.

joaopapereira:
This PR accomplishes 2 thinks:
 - Extract the code that lives in the controller into Interactors
   - Job Developer self assigns to a specific Job Seeker
   - Agency Person assigns a specific Job Developer to a Job Seeker
 - Close the task "JS needs a JD" when a Job Developer is assigned to a Job Seeker

The idea behind point 1 is to eventually remove as much code as possible from the controller. By doing this we can do tests the run faster because we do not need to run through the full Rails Stack.
I future step that I see here is to remove at all the dependencies from Rails so that our Rspec test run faster....... 10 minutes to run RSpec tests is too much

So a more detailed explanation of what I done:
 - Moved the code from AgencyPersonController#assign_job_seeker to JobSeekers::AssignAgencyPerson#call and created tests around the behavior expected, backfilled Controller tests with a double of this UseCase
- Moved the code from AgencyPersonController#update that assigned the JobSeekers into AgencyPeople::AssignNewJobSeekers#call and created tests around the behavior expected, backfilled Controller tests with a double of this UseCase

patmbolger:
OK, that makes sense.  Looks good.

PR #662, opened 2017-12-01
Add doku configuration

joaopapereira:


patmbolger:
I am wondering if we need to add something to the Azure config in order to start the DelayedJob process?

joaopapereira:
No, that need to be configured in the Host machine, already talked with Sam in order to do that change

PR #664, opened 2017-12-09
Update rails to 4.2.10

joaopapereira:


patmbolger:
This looks good. Should we go ahead and merge or did you want to address the remaining Hound violations?

Also, we should work on moving to rails 5.x.  Do you agree?

joaopapereira:
@patmbolger I am not sure how to split the regex into multiple lines.
This is an intermediary step to 5. Now that we are in the latest 4 we should have a smooth ride to 5

patmbolger:
Yes, thats a pain ... lets not worry about it.  We can merge this after we fix that production bug.

patmbolger:
All tests pass for me (locally).  Merging.

PR #665, opened 2017-12-13
End of Sprint 53

joaopapereira:


joaopapereira:
There is some flakyness on the CI, but in different builds of the same code I was able to see all tests pass.

So I am going to merge the PR

PR #668, opened 2017-12-21
WIP: Remove act as

joaopapereira:
This will be the first step to update rails to 5.1 since act_as gem is not supported in that version.
Removing act_as also allow us to think about the possibility of having tests that do not depend on the database, making them much faster then they currently are

related to AgileVentures/MetPlus_tracker/issues/724


tansaku:
@joaopapereira is saying that this is probably not an issue any more

PR #671, opened 2018-01-18
unit tests for dependent destroy associations

Erika-Barr:
Closes AgileVentures/MetPlus_tracker#671


tansaku:
thanks for this @Erika-Barr - will see it gets reviewed ASAP :-)

Erika-Barr:
Thank you, looking forward to feedback

spec/models/agency_person_spec.rb
```diff
@@ -19,6 +19,33 @@
...
+      aa_person2.save
+      aa_person2.active
+    end		
+    it 'status_changes association with dependent::destroy deletes record' do
+      first_status_change = aa_person2.status_changes.first.id	
+      aa_person2.destroy
+      expect{StatusChange.find(first_status_change)}.to raise_error(ActiveRecord::RecordNotFound)
+    end
+    it 'job_seeker :through association with dependent::destroy deletes record' do
+      expect(agency_relation).to be_valid
```
tansaku: some schools of thought suggest that there should only be one `expect` per `it` block in a unit test.  However since this test is using FactoryBot and involves creating many different models it's clearly an integration test, in which case perhaps we don't have to be so strict ...

tansaku:
So a high level comment on the task itself is that I wonder if there's a preference for integration tests over unit tests?  One possibility if we wanted to have unit tests rather than (or as well as integration tests) is that we could use thoughtbot shoulda matchers like so:

```rb
it { is_expected.to have_many(:status_changes).dependent(:destroy) }
```

https://github.com/thoughtbot/shoulda-matchers

spec/models/agency_person_spec.rb
```diff
@@ -19,6 +19,33 @@
     it { is_expected.to have_many(:status_changes) }
   end
 
+  describe 'When agency_person is destroyed' do
+    let(:agency)          { FactoryBot.create(:agency) }
+    let(:aa_person1)      { FactoryBot.build(:agency_person, agency: agency) }
```
tansaku: do we need to have two different persons here?  could we just have the job_seeker and status_change both be added to a single person and use the subject syntax, e.g.

```
subject(:agency_person) { ... }
```

to help indicate the focus of the test?

tansaku:
@Erika-Barr overall great job getting this done - I've added a few comments about how we might consider adjusting the style - hope they're helpful :-)

Erika-Barr:
Yes @tansaku, I think thoughtbot's matchers  are what I was supposed to use. As a result, is the describe block in "spec/models/branch_spec.rb:64" necessary? I initially used that as a template

spec/models/job_application_spec.rb
```diff
@@ -21,10 +21,11 @@
   describe 'Associations' do
     it { is_expected.to belong_to :job_seeker }
     it { is_expected.to belong_to :job }
-    it { is_expected.to have_many(:status_changes) }
+    it { is_expected.to have_many(:status_changes).dependent(:destroy) }
     it { is_expected.to have_many(:application_questions) }
-    it { is_expected.to have_many(:questions).dependent(:destroy) }
+    it { is_expected.to have_many(:questions).through(:application_questions).dependent(:destroy) }
```
patmbolger: Good catch on the need for `through` qualifier here.

Erika-Barr:
Ok, I added the changes, let me know if I need to change/add anything else,
thanks

patmbolger:
Looks good - merging.

PR #673, opened 2018-02-02
make GA id dynamic

tansaku:
fixes #698


tansaku:
@joaopapereira thanks - is this deployed to production now?

joaopapereira:
@tansaku yes it should be

tansaku:
@joaopapereira doesn't seem to be on master: https://github.com/AgileVentures/MetPlus_PETS/blob/master/app/assets/javascripts/analytics.js

PR #679, opened 2018-03-13
adding letsencrypt support

tansaku:
fixes #683


tansaku:
@joaopapereira I've pushed up some fixes - I'm not sure about how this would work with the cruncher - would have to look at how the two systems interact with each other

joaopapereira:
@tansaku It need to be a separate implementation where we create an endpoint for the cruncher or if we can share the certificate we do not need to do much.

tansaku:
@joaopapereira looks like we don't need this PR in order to get https working on dokku - see https://staging.petsworkforce.com - we can easily generate a separate cert for the cruncher

PR #682, opened 2018-04-03
Number of openings on a specific job

maryam21:
Fixes  AgileVentures/MetPlus_tracker#693

Enables company person to specify the number of available positions for a job opening, so that multiple people can be hired for the same position.
1. Added cucumber acceptance tests.
2. Added available positions and remaining positions fields in the job data model.
3. Added the available positions field in the job form and in job page.
4. Updated the create and update method in the jobs controller so that available positions field is assigned to remaining positions field.
5. Updated JobApplications module method call so that it calls the decrease_remaining_positions method that decreses the remaining positions field by 1.
6. Updated JobApplication model method accept so that the job status changes to filled and the other job applications are rejected only when there are no more remaining positions.
7. Added unit test for decrease_remaining_positions method  and update the tests for job application model.


app/models/job_application.rb
```diff
@@ -35,12 +35,12 @@ def processing?
   def accept
     accepted!
     StatusChange.update_status_history(self, :accepted)
-    reject_applications = job.job_applications.where.not(id: id)
+    reject_applications = job.job_applications.where.not(status: 1)
```
joaopapereira: Can we change this to
....not(status: :accepted)

Will this work?

app/models/job_application.rb
```diff
@@ -35,12 +35,12 @@ def processing?
...
     reject_applications.each do |application|
       application.not_accepted!
       StatusChange.update_status_history(application, :not_accepted)
-    end
-    job.filled
+    end if job.remaining_positions == 1
```
joaopapereira: Can we use a regular if statement here like

if job.remaing_positions == 1
  reject_applications = ......
else if job.remainig_positions == 0
 job.filled
end

spec/interactors/job_applications/hire_spec.rb
```diff
@@ -96,6 +96,14 @@ class TestTaskHelper
           expect(Event).not_to have_received(:create)
         end
       end
+
```
joaopapereira: Missing a test that reads:
`when remaining positions is 3 and a applications is accepted, it should not reject all the other applications`

maryam21: This case is already tested line 59 in features/number_of_openings.feature . For the method close_all_tasks(job_application) it only closes the tasks for the application just accepted.

maryam21: This case is already tested line 59 in features/number_of_openings.feature . For the method close_all_tasks(job_application) it only closes the tasks for the application just accepted.

joaopapereira: Ok

PR #695, opened 2018-07-13
Rake task to synchronize jobs from PETS to the cruncher

joaopapereira:
This patch creates a Rake task that will update the jobs on the cruncher and if they do not exist creates them

Closes AgileVentures/MetPlus_tracker#696
Closes AgileVentures/MetPlus_tracker#718

### DIFF
--- a/app/interactors/jobs/query.rb
+++ b/app/interactors/jobs/query.rb
@@ -1,0 +1,7 @@
+module Jobs
+  class Query
+    def all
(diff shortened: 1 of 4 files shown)


joaopapereira:
@tansaku if we did not store the information in the cruncher, every time that you would ask for a match you would have to send all the jobs or all the resumes for it to crunch them and then get you the result

PR #700, opened 2018-10-02
Display current agency in the title

NickSchimek:
Test added to check agency name appears in the title with the associated
step definition. Updated the page title to display the current agencies
name.

This is done to support multiple agencies.

Fixes AgileVentures/MetPlus_tracker#712

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -7,7 +7,7 @@ module ApplicationHelper
   end
 
   def full_title(page_title = '')
-    base_title = 'MetPlus'
+    base_title = current_agency.name
     page_title.empty? ? base_title : "#{page_title} | #{base_title}"
   end
 
--- a/features/multiple_agency.feature
+++ b/features/multiple_agency.feature
@@ -28,3 +28,8 @@ Feature: Support multiple agencies
     And I click "Update Agency" button
     Then I am on the home page
     And I should see "MetPlus"
+
+  Scenario: Agency name is displayed in the page title
+    When I am on the home page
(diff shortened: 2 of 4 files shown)


NickSchimek:
Not sure why current_agency is undefined when RSpec runs tests. 
```
undefined local variable or method `current_agency'
```
Does it have something to do with the configuration?

joaopapereira:
@NickSchimek you might need to include the helper AgencyHelper on ApplicationHelper. Because in rails the helpers are like special classes and are not automatically included in each other

NickSchimek:
hmmm the current_agency method is defined in the ApplicationController. The full_title helper method has no issues when a request or system test is routed through the controller... it's only during the RSPec unit test. I suppose it's probably possible to move the current_agency out of the controller and into the AgencyHelper and then require that. 

Line 14 of the ApplicationController defines it as a helper method too... so I was hoping it would already be visible in the ApplicationHelper, I guess that's not the case.
```
helper_method :pets_user, :current_agency, :determine_if_admin
```
I will see about moving the method unless you have a better a suggestion?

joaopapereira:
Sorry for the late reply. Looks like a good suggestion

NickSchimek:
no worries @joaopapereira. I've been busy with other things. I tried pulling the current_agency method out of the controller and into a helper, but once I did that then I would need to move the pets_user method too, and that caused more issues. I abandoned that approach and stubbed the method call instead. The cucumber test checks it end to end anyway.

It seems there is one semaphoreci failure, but that is passing locally for me. Can you re-run it? I don't seem to have the ability to do that.

PR #706, opened 2018-11-11
set max-width to 100% for all input fields

dkuku:
There is a bug, when the input field size parameter is set to too large value. This PR fixes it.
jobs


joaopapereira:
@dkuku we have this test
`cucumber features/sorting_job_applications.feature:46 # Scenario: Job seeker applies to job`

Always failing, does it also fail in your local?

joaopapereira:
@dkuku ^^^^

